PR #7, opened 2018-06-13
Update access permissions for billing

zaro0508:
Give Mike access to organizations and billing.


zaro0508:
Admins do not have access to billing.  You may want to add that in a separate PR.

PR #12, opened 2018-07-20
remove auditor policy

zaro0508:
giving auditors direct access to our account is a bad practice.
setup cross account access would be better.


zaro0508:
yes we should

PR #34, opened 2020-05-11
[IT-870] Setup jumpcloud IDP

zaro0508:
Setup AWS account with Jumpcloud IDP. The idea is to allow users/groups
in Jumpcloud to access AWS using an admin role. The user/group mapping
to this role is configured in jumpcloud.


config/prod/jumpcloud-idp.yaml
```diff
@@ -0,0 +1,10 @@
+template_path: "jumpcloud-idp.yaml"
+stack_name: "jumpcloud-idp"
+stack_tags:
+  Department: "Platform"
+  Project: "Infrastructure"
+  OwnerEmail: "it@sagebase.org"
+parameters:
+  OrganizationsAdminMetadata: !file_contents idp/jumpcloud/admincentral-admin.xml
```
ahayden, line 8: The filename of the XML source should be `organizations-admin.xml` no?

zaro0508: thanks, I fell asleep at the wheel.

PR #39, opened 2021-01-20
Setup org-formation to manage multiple accounts

zaro0508:
One repo to rule them all! Reesolves SC-286, IT-868, and IT-626


xschildw (approved):
Nice! Thx again for walk-through and detailed explanations.

ahayden (changes requested):
Please move to the sage-bionetworks-it org

zaro0508:
> Please move to the sage-bionetworks-it org

@ahayden How about we move after this PR merges?

PR #45, opened 2021-01-29
[IT-1060] migrate scicomp templates

zaro0508:


ahayden (approved):
This changes how sceptre runs, but doesn't change any of the inputs to sceptre. So, when it does run it will now rebuild any resources. Confirmed with Khai.

xschildw (approved):

PR #69, opened 2021-02-04
create KMS key

zaro0508:
Bootstrap every AWS account with a kms key for IT to use.
The idea is to have this key available in every account in case
we need to setup secrets for IT to use.

### DIFF
--- a/org-formation/300-account-defaults/kms-key.yaml
+++ b/org-formation/300-account-defaults/kms-key.yaml
@@ -1,0 +1,80 @@
+AWSTemplateFormatVersion: 2010-09-09
+Parameters:
+  AliasName:
+    MinLength: 1
+    MaxLength: 256
+    Type: String
+    AllowedPattern: ^[a-zA-Z0-9/_-]+$
+    Description: Friendly name for the KMS Key
+  EnableKeyRotation:
+    Type: String
+    AllowedValues: ['true', 'false']
+    Default: true
+    Description: Allow KMS to automatically rotate the key
+  PendingWindowInDays:
+    MinValue: 7
+    MaxValue: 30
+    Type: Number
+    Default: 7
+    Description: Number of days before KMS delets the key
+  AdminPrincipalArns:
+    Type: CommaDelimitedList
+    Description: List of user/role/account arns that can administer the key
+  UserPrincipalArns:
+    Type: CommaDelimitedList
+    Description: List of user/role/account arn that can use the key
+Resources:
+  Key:
+    Type: "AWS::KMS::Key"
+    Properties:
+      Description: !Sub '${AWS::StackName} created key'
+      EnableKeyRotation: !Ref EnableKeyRotation
+      PendingWindowInDays: !Ref PendingWindowInDays
+      KeyPolicy:
+        Version: "2012-10-17"
+        Statement:
+          -
+            Sid: "Allow administration of the key"
+            Effect: "Allow"
+            Principal:
+              AWS: !Ref AdminPrincipalArns
+            Action:
+              - "kms:Create*"
+              - "kms:Describe*"
+              - "kms:Enable*"
+              - "kms:List*"
+              - "kms:Put*"
+              - "kms:Update*"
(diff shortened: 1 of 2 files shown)


org-formation/300-account-defaults/organization-tasks.yaml
```diff
@@ -59,3 +59,20 @@ SceptreLambdaBucket:
...
+ItKmsKey:
+  Type: update-stacks
+  Template: ./kms-key.yaml
+  StackName: it-kms-key
+  MaxConcurrentStacks: 10
+  Parameters:
+    AliasName: alias/itkey
+    AdminPrincipalArns:
+      - !Sub 'arn:aws:iam::${AWS::AccountId}:root'
+      - !Sub 'arn:aws:iam::${AWS::AccountId}:role/OrganizationFormationBuildAccessRole'
```
ahayden, line 72: Include `arn:aws:iam::${AWS::AccountId}:role/AccountName-admin` in this list? Is the `AccountName` a stack output?

org-formation/300-account-defaults/organization-tasks.yaml
```diff
@@ -59,3 +59,20 @@ SceptreLambdaBucket:
...
+    UserPrincipalArns:
+      - !Sub 'arn:aws:iam::${AWS::AccountId}:role/OrganizationFormationBuildAccessRole'
```
ahayden, line 74: Include `arn:aws:iam::${AWS::AccountId}:role/AccountName-admin` in this list too?

org-formation/300-account-defaults/organization-tasks.yaml
```diff
@@ -59,3 +59,20 @@ SceptreLambdaBucket:
...
     Account: '*'
     Region: us-east-1
+
+ItKmsKey:
+  Type: update-stacks
+  Template: ./kms-key.yaml
+  StackName: it-kms-key
+  MaxConcurrentStacks: 10
+  Parameters:
+    AliasName: alias/itkey
```
ahayden, line 69: This name is a little too nonspecific... Maybe use `alias/it-buildtin-key` so that it can be differentiated from `alias/it-specialproj-key`?

org-formation/300-account-defaults/kms-key.yaml
```diff
@@ -0,0 +1,80 @@
...
+  UserPrincipalArns:
+    Type: CommaDelimitedList
+    Description: List of user/role/account arn that can use the key
+Resources:
+  Key:
+    Type: "AWS::KMS::Key"
+    Properties:
+      Description: !Sub '${AWS::StackName} created key'
+      EnableKeyRotation: !Ref EnableKeyRotation
+      PendingWindowInDays: !Ref PendingWindowInDays
```
ahayden, line 32: Nice- good idea

ahayden (approved):

org-formation/300-account-defaults/organization-tasks.yaml
xschildw: what about the idp admin role?

org-formation/300-account-defaults/organization-tasks.yaml
zaro0508: actually users in the JC `aws-admins` group can assume `OrganizationFormationBuildAccessRole` role,  so i think it’s good as is.

org-formation/300-account-defaults/organization-tasks.yaml
zaro0508: answered above @xschildw

org-formation/300-account-defaults/organization-tasks.yaml
zaro0508: well, i don't think we would call anything `it-specialproj-key`, each project might have a key of their own and it would probably just be named `specialproj-key`.  how name it `it-infra`?

zaro0508: alias name updated

xschildw (approved):

PR #71, opened 2021-02-10
Redeployed VPC so new PCX

xschildw:
I changed the synapsedev VPC and got a new peering connection. This PR just adds the routes in AdminCentral (and should also remove the existing routes which are now blackholes).


zaro0508 (approved):
looks good, although I'm not sure it will remove the existing pcx.  you might need to verify after merging.

PR #77, opened 2021-02-24
use shared files from aws-infra repo

zaro0508:
orf-formation allows deploying with remote template files. Use
shared remote template files instead of local files.


xschildw (approved):
I see what the changes is but not why... It's better to have the files in aws-infra so they can be used outside of organization context if need be?

zaro0508:
> I see what the changes is but not why... It's better to have the files in aws-infra so they can be used outside of organization context if need be?

@xschildw  It's better to push templates in aws-infra so they can be shared out to multiple org-formation/sceptre 
 projects.

PR #78, opened 2021-02-24
[SC-240] Allow S3 Launch role to create policies

zaro0508:
Setting up S3 same region access restriction requires creating
custom bucket policies. Allow the role that creates the bucket
permissions to create those policies.


ahayden (approved):
This doesn't look like a major escalation in rights given what the role already can do.

PR #79, opened 2021-02-24
grant SCS3LaunchRole Lambda Create/Delete permissions

zimingd:
Allow SCS3LaunchRole to create and Delete Lambda function, which is necessary for restricting S3 Bucket downloads to the same AWS region.


ahayden (approved):
Approved... needed addtion for this feature.

PR #80, opened 2021-02-25
Add a service user for NLP

zaro0508:
Add an IAM service account to AWS sandbox for @thomasyu888 so he
can use it to deploy his app https://github.com/nlpsandbox/aws-cloudformation

depends on https://github.com/Sage-Bionetworks/aws-infra/pull/272

### DIFF
--- a/org-formation/600-access/organization-tasks.yaml
+++ b/org-formation/600-access/organization-tasks.yaml
@@ -1,0 +1,39 @@
+CommunityIamSamlProviderRP:
+  Type: register-type
+  SchemaHandlerPackage: s3://community-resource-provider-catalog/community-iam-samlprovider-0.2.0.zip
+  ResourceType: 'Community::IAM::SamlProvider'
+  MaxConcurrentTasks: 10
+  OrganizationBinding:
+    IncludeMasterAccount: true
+    Account: '*'
+    Region: us-east-1
+
+AccessTestItOrgMgntAdmin:
+  Type: update-stacks
+  Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/master/templates/IAM/jumpcloud-idp.yaml
+  DependsOn: CommunityIamSamlProviderRP
+  StackName: access-testitorgmgnt-admin
+  MaxConcurrentStacks: 10
+  Parameters:
+    MetadataDocument: !ReadFile ./MetadataDocuments/testitorgmgnt-admin.xml
+    SamlProviderName: testitorgmgnt-admin
+    SamlProviderPolicyArns:
+      - arn:aws:iam::aws:policy/AdministratorAccess
+  DefaultOrganizationBinding:
+    IncludeMasterAccount: true
+    Account: !Ref MasterAccount
+    Region: us-east-1
+
+# A service account for https://github.com/nlpsandbox/aws-cloudformation
+NlpSandboxServiceAccount:
+  Type: update-stacks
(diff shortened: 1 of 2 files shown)


org-formation/600-access/organization-tasks.yaml
```diff
@@ -0,0 +1,14 @@
+
+# A service account for https://github.com/nlpsandbox/aws-cloudformation
+NlpSandboxServiceAccount:
+  Type: update-stacks
+  Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/master/templates/IAM/service-account.yaml
+  StackName: NlpSandboxServiceAccount
+  MaxConcurrentStacks: 10
+  Parameters:
+    ServiceUserPolicies:
+      - arn:aws:iam::aws:policy/AdministratorAccess
```
ahayden, line 10: Does this account really need this policy? If its in sandbox, it doesn't matter much, but one day it will migrate to a prod account, so we need to get the permissions right in the first place.

ahayden (changes requested):
Please reduce perms

org-formation/600-access/organization-tasks.yaml
zaro0508: We don't need to deploy this same policy for prod account.  We can create a new `NlpProdServiceAccount` with a different set of policies and deploy that to prod.

xschildw: +1

ahayden: That sounds right... except that we won't know what the correct perms should be unless the sandbox prototype proves it

zaro0508: @thomasyu888 will figure that out in sandbox then we can tighten it up before deploying to scicomp

xschildw: I agree, sandbox should be where we figure out the problems, including what permissions are needed to run...
Is this the right place to define the policy (i.e. should we not just have to setup a service account that assumes
a role, the role itself being defined by the app dev)?

zaro0508: I think this PR is doing exactly what you are suggesting @xschildw?   I'm not sure what you feel needs to change about it.

xschildw: My bad, you're correct. My only issue is with not specifying the actual permissions in sandbox.

xschildw (approved):
Let's just get on with it, we can debate out of the way...

thomasyu888 (approved):

org-formation/600-access/organization-tasks.yaml
thomasyu888: It seems like what i need is `arn:aws:iam::aws:policy/AmazonEC2FullAccess`.  This user needs to: 

- create, update, delete ec2
- create, update, delete ELB2 resources (listeners, load balancer, Target group)
- create, update, delete security groups

ahayden: That seems to right to me... you can either try it with the permission set you think it needs and then confirm it works, or give it the admin policy and guess later... the second method seems wrong, but since this is sandbox it doesn't matter much

thomasyu888, line 10:
```suggestion
      - arn:aws:iam::aws:policy/AmazonEC2FullAccess
```

PR #92, opened 2021-03-10
[IT-1074] Remove IAM user accounts from sandbox

zaro0508:
We federate all user access from the jumpcloud IDP which means users should no
longer have direct IAM access to the AWS sandbox account.


ahayden (approved):
Do you think we need to notify @sandbox?

xschildw (approved):

zaro0508:
i have no idea what you mean @ahayden.  who's @sandbox?

PR #94, opened 2021-03-10
update NlpSandboxServiceAccount with template and policies

zaro0508:


thomasyu888 (approved):

thomasyu888:
Ah, i see.  Was it because it was using `v0.2.1` that it didn't create the policies?  Also do we see an issue with adding this service user to scicomp as well?

PR #98, opened 2021-03-12
refactor project structure

zaro0508:
Refactor to more closely align with the ofn reference project
structure[1].

* create separate folders for ofn resource providers
* move scp out of organizations.yaml and into it's own folder

[1] https://github.com/org-formation/org-formation-reference


org-formation/organization.yaml
```diff
@@ -444,60 +444,3 @@ Organization:
                 'aws:RequestedRegion':
                   - us-east-1
                   - us-east-2
```
ahayden, line 446: Do you mean to include `us-east-2`? I see that its in the sid name on L436 too

ahayden (approved):
Nice

org-formation/005-types/_tasks.yaml
```diff
@@ -0,0 +1,114 @@
...
+
+PasswordPolicyRp:
+  Type: register-type
+  ResourceType: 'Community::IAM::PasswordPolicy'
+  SchemaHandlerPackage: !Sub 's3://${catalogBucket}/community-iam-passwordpolicy-0.2.0.zip'
+  MaxConcurrentTasks: 10
+  OrganizationBinding:
+    IncludeMasterAccount: true
+    Account: '*'
+    Region: !Ref primaryRegion
```
ahayden, line 75: Why is this only on primary region?

org-formation/005-types/_tasks.yaml
```diff
@@ -0,0 +1,114 @@
...
+    Region: !Ref allRegions
+
+SsoAssignmentGroupRp:
+  Type: register-type
+  ResourceType: 'Community::SSO::AssignmentGroup'
+  SchemaHandlerPackage: !Sub 's3://${catalogBucket}/community-sso-assignmentgroup-0.3.1.zip'
+  MaxConcurrentTasks: 10
+  OrganizationBinding:
+    IncludeMasterAccount: true
+    Region: !Ref primaryRegion
```
ahayden, line 114: Same as above... why only primary?

org-formation/organization.yaml
zaro0508: actually those were read by ofn and included in this file.  I'm going to move all of those SCPs into the `010-scps` folder like i did with RestrictUnusedRegionsSCP.  will do it in follow on PR.

org-formation/005-types/_tasks.yaml
zaro0508: i just copied these settings from the ofn reference project.  I currently have primaryRegion and all AllRegions both set to us-east-1 right now.  so it's the same anyways

PR #99, opened 2021-03-12
refactor ofn scp

zaro0508:
* remove unused SCPs
* apply PreventDisableCloudtrail SCP to all accounts
* apply PreventDisableCloudwatchConfigs SCP to all accounts


ahayden (approved):
The SCP I care most about are still in here... looks good

PR #101, opened 2021-03-16
update SCPs

zaro0508:
Restrict allowed regions on OUs and accounts.

### DIFF
--- a/org-formation/010-scps/_tasks.yaml
+++ b/org-formation/010-scps/_tasks.yaml
@@ -43,7 +43,7 @@ DenyAllRegionsOutsideUsEast1:
   DefaultOrganizationBinding:
     IncludeMasterAccount: true
   Parameters:
-    targetIds: [ !Ref TestOU ]
+    targetIds: [ !Ref ITOU, !Ref ScipoolDevAccount, !Ref ScipoolProdAccount, !Ref ScicompAccount]
 
 DenyAllRegionsOutsideUs:
   Type: update-stacks


xschildw (approved):

org-formation/010-scps/_tasks.yaml
```diff
@@ -43,7 +43,7 @@ DenyAllRegionsOutsideUsEast1:
   DefaultOrganizationBinding:
     IncludeMasterAccount: true
   Parameters:
-    targetIds: [ !Ref TestOU ]
+    targetIds: [ !Ref ITOU, !Ref ScipoolDevAccount, !Ref ScipoolProdAccount, !Ref ScicompAccount]
```
ahayden, line 46: What not use the `science` OU instead of referencing these three accounts which are members of the `science` OU? Also... Can we not apply this to all OUs?

ahayden (changes requested):
This should be applied across all OUs IMO

org-formation/010-scps/_tasks.yaml
zaro0508: ScienceOU has scipool, scicomp and sandbox.  People (like @tschaffter) have already provisioned stuff in non us-east-1 regions so we need to allow non us-east-1 for sandbox.   of course unless you can convince him to either abandone or migrate his resources in non us-east-1 region.

tschaffter: @zaro0508 I can move resources if you give me a list. But that's only part of the solution. The solution should be to disable people to create resources in region other than us-east-1 is that's what Sage wants.

ahayden: Check the EC2 console of `us-west-2` @tschaffter ... there's a few parked things... Also @jaeddy has a couple things. This PR does your suggestion @tschaffter 

Alternatively, we can remove sandbox from that OU, which might actually be better long term, huh @zaro0508 ?

zaro0508: > Alternatively, we can remove sandbox from that OU, which might actually be better long term, huh @zaro0508 ?

We can @ahayden, how about you do that as a follow on PR?

ahayden: OK, I think the way you have it is correct for today... when @tschaffter and @jaeddy remove the us-west resources, we can add the regional restriction to all accounts... and I think the correct OUs after that point would be two: `science-prod` and `science-dev`

ahayden (approved):

org-formation/010-scps/_tasks.yaml
zaro0508: thanks.  i'm looking forward to reviewing your PR for that.

PR #105, opened 2021-03-19
[IT-1124] Create account for iAtlas project

zaro0508:


xschildw (approved):

ahayden (approved):
I think this is the way to go for projects like iAtlas... What determines the dependencies that get build into it? Like logging and GuardDuty, etc.

zaro0508:
> I think this is the way to go for projects like iAtlas... What determines the dependencies that get build into it? Like logging and GuardDuty, etc.

nothing at this moment @ahayden but you are welcome to make suggestions, in issue IT-1101, on what you think should be setup with every new account.

PR #107, opened 2021-03-22
Disable EBS encryption for imagecentral account

zaro0508:
EBS encryption  needs to be disabled in order to build public image because
packer can no longer start unencrypted boot images
https://github.com/hashicorp/packer/issues/9529


ahayden (approved):
Looks good

xschildw (approved):

PR #108, opened 2021-03-22
[SC-314] AWS account for ECR

zaro0508:
Setup ECR in an account to allow Sage to share docker containers.

### DIFF
--- a/org-formation/organization.yaml
+++ b/org-formation/organization.yaml
@@ -50,6 +50,7 @@ Organization:
         - !Ref SecurityCentralAccount
         - !Ref ImageCentralAccount
         - !Ref ITSandboxAccount
+        - !Ref ContainerCentralAccount
 
   TestOU:
     Type: OC::ORG::OrganizationalUnit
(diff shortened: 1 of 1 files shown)


xschildw (approved):

brucehoff (approved):

org-formation/organization.yaml
```diff
@@ -50,6 +50,7 @@ Organization:
         - !Ref SecurityCentralAccount
         - !Ref ImageCentralAccount
         - !Ref ITSandboxAccount
+        - !Ref ContainerCentralAccount
```
ahayden, line 53: Is the plan that scipool users will register containers? I wonder if the `ContainerCentral` account should go in a ScienceOU rather than an ITOU... since non-IT users will be registering things in it

ahayden (approved):
Looks good, but not sure if the account is for endusers or for IT, or some combo

org-formation/organization.yaml
zaro0508: let's figure out OU and account placement in PR #123

zaro0508:
At this time there is no use case for creating a shared ECR

PR #110, opened 2021-03-23
[IT-1137] Use the official AWS SAML provider

zaro0508:
* Switch to using the officially supported AWS SAML provider
* Use org-formation to setup JC SAML connector instead of sceptre
* Remove remove existing connectors.


ahayden (approved):
Seems fine... is there a risk of service denial with with? Sessions should continue until expiration, so I don't think the risk is significant.

zaro0508:
I don't know what you mean @ahayden.  i guess thee's always some type of risk but there's nothing on my radar at this time.

PR #112, opened 2021-03-24
update template references

zaro0508:
Our workflow has been to add CFN template files to aws-infra repo let the
CI deploy to S3 bucket then reference the file from the S3 bucket.  This
changes it so that we reference the file directly from GH instead of the S3
bucket.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -122,7 +122,7 @@ lambdas/*.zip
 .DS_Store
 
 # temp files
-/temp/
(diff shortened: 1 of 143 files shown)


zaro0508:
Decided against this change because it would introduce two ways to reference files, GH and Buckets.  Since we can only  reference built SAM lambda templates from buckets it's just easier to standardize on one way to reference files.

PR #81, opened 2021-02-25
create AWS account for Tess

zaro0508:

PR #113, opened 2021-03-24
Access to MobileHealthDataEngineering AWS account

zaro0508:
Setup jumpcloud access to the aws mobilehealth-dataengineering account.

depends on PR #81


org-formation/600-access/_tasks.yaml
```diff
@@ -1,4 +1,19 @@
 
+JumpcloudMobileHealthDataEngineeringAdmin:
+  Type: update-stacks
+  Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/master/templates/IAM/jumpcloud-idp.yaml
+  StackName: jc-access-mobilehealth-dataengineering-admin
+  MaxConcurrentStacks: 10
+  Parameters:
+    MetadataDocument: !ReadFile ./MetadataDocuments/MobileHealthDataEngineering-admin.xml
```
zaro0508, line 8: the metadata file is created in JC it can only be created after creating account. will need to update this file. this is only a placeholder.

zaro0508: updated with valid file

tthyer (approved):

PR #115, opened 2021-03-25
[IT-1101] Setup budget notifications

zaro0508:
Setup a AWS account budget notification. This will allow us to configure
a budget on an account with tags.  Notifications are sent when an account
exceeds the budget.

depends on https://github.com/Sage-Bionetworks/aws-infra/pull/286

PR #116, opened 2021-03-25
[IT-1101] budget notification for mobile health data engeering account

zaro0508:
Setup a budget for mobile health data engeering account.  Get notified
if the account cost exceed the expected threshold.

depends on #115

### DIFF
--- a/org-formation/_parameters.yaml
+++ b/org-formation/_parameters.yaml
@@ -16,3 +16,7 @@ allRegions:
   Type: List<String>
   Default:
     - us-east-1
+
+budgetOwnerEmail:
+  Type: String
+  Default: bruce.hoff@sagebase.org
--- a/org-formation/_parameters.yaml
+++ b/org-formation/_parameters.yaml
@@ -16,7 +16,3 @@ allRegions:
   Type: List<String>
   Default:
     - us-east-1
-
-budgetOwnerEmail:
-  Type: String
-  Default: bruce.hoff@sagebase.org
--- a/org-formation/organization.yaml
+++ b/org-formation/organization.yaml
@@ -1,9 +1,6 @@
 AWSTemplateFormatVersion: '2010-09-09-OC'
 Description: Manage AWS organizations with master account
 
-Parameters:
-  <<: !Include './_parameters.yaml'
-
 Organization:
   MasterAccount:
     Type: OC::ORG::MasterAccount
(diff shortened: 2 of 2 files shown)


ahayden (approved):
Threshold is in dollars I assume

xschildw (approved):

org-formation/_parameters.yaml
```diff
@@ -16,3 +16,7 @@ allRegions:
   Type: List<String>
   Default:
     - us-east-1
+
+budgetOwnerEmail:
+  Type: String
+  Default: bruce.hoff@sagebase.org
```
brucehoff, line 22: Seems like this should be an email alias (a Google group in the sagebase.org domain) that individuals could subscribe to / unsubscribe from.

brucehoff (changes requested):

org-formation/_parameters.yaml
zaro0508: OK.  who should be in the alias @brucehoff ?

brucehoff: I need more context to answer your question, @zaro0508 .  Who asked for this?  Who is the one responsible for controlling costs incurred in the "mobile health data engeering account"?  It seems like that person would want to subscribe to the alias.

zaro0508: set to send @tthyer budget notifications

brucehoff (approved):

tthyer (changes requested):
Why is it necessary to set up a budget notification on this account? I anticipated it would be tracked like any other internal account. 

This is not an account for use by external users.

ahayden:
This is a new kind of account.... a dev account for a specific project. I think it's a good idea to put budget alerts on all accounts like this, just so the project/system owner is aware of the costs.

tthyer:
@ahayden Fair enough -- however, 200 is very low. Maybe multiply that by 100?
IIRC when I wrote the budget notification lambda, it is operating on annual rather than monthly costs. I would prefer the latter for internal use. I would also prefer to receive budget reports than simple notifications.

@zaro0508 @ahayden Can we use aws-mobilehealth-dataengineering@sagebase.org rather than my email?

zaro0508:
updated per your request @tthyer.  you have access to the account so you can create your own budget reports after getting notifications.

tthyer (approved):

ahayden (approved):
Looks fine to me

PR #120, opened 2021-03-26
[IT-1101] Setup GuarDuty

zaro0508:
Enable guard duty on all accounts and have findings sent to our
securitycentral account.  This is the master member guard duty
configuration[1]

[1] https://docs.aws.amazon.com/guardduty/latest/ug/guardduty_organizations.html

depends on https://github.com/Sage-Bionetworks/aws-infra/pull/288


org-formation/070-guard-duty/_tasks.yaml
```diff
@@ -0,0 +1,75 @@
...
+  StackDescription: GuardDuty - Trusted IPs bucket
+  DefaultOrganizationBindingRegion: !Ref primaryRegion
+  DefaultOrganizationBinding:
+    Account: !Ref accountId
+  TerminationProtection: true
+  Parameters:
+    resourcePrefix: !Ref resourcePrefix
+    bucketName: !Sub '${resourcePrefix}-${appName}-trusted-ips'
+
+GuardDutyTrustedIps:
```
ahayden, line 51: I think we should leave this out as its unnecessary

org-formation/070-guard-duty/trusted-ips.txt
```diff
@@ -0,0 +1 @@
+127.0.0.1
```
ahayden, line 1: And the `trusted-ips.txt` file can be removed

org-formation/070-guard-duty/_tasks.yaml
```diff
@@ -0,0 +1,75 @@
...
+  RemotePath: !Sub 's3://${resourcePrefix}-${appName}-trusted-ips/trusted_ips.txt'
+  OrganizationBinding:
+    Account: !Ref accountId
+    Region: !Ref primaryRegion
+
+GuardDutyExtension:
+  Type: update-stacks
+  DependsOn:
+    - GuardDuty
+    - GuardDutyTrustedIps
```
ahayden, line 64: This would get excised too

org-formation/070-guard-duty/_tasks.yaml
```diff
@@ -0,0 +1,75 @@
...
+  StackName: !Sub '${resourcePrefix}-${appName}-extension'
+  StackDescription: GuardDuty - Extension (including IP Set)
+  DefaultOrganizationBindingRegion: !Ref primaryRegion
+  DefaultOrganizationBinding:
+    Account: !Ref accountId
+  TerminationProtection: true
+  Parameters:
+    resourcePrefix: !Ref resourcePrefix
+    detectorId: !CopyValue [ !Sub '${resourcePrefix}-${appName}-detector-id' ]
+    bucketName: !CopyValue [ !Sub '${resourcePrefix}-${appName}-trusted-ips-bucket-name' ]
```
ahayden, line 75: Not needed

org-formation/070-guard-duty/_tasks.yaml
```diff
@@ -0,0 +1,75 @@
...
+    Region: !Ref primaryRegion
+
+GuardDutyExtension:
+  Type: update-stacks
+  DependsOn:
+    - GuardDuty
+    - GuardDutyTrustedIps
+  Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/master/templates/GuardDuty/guard-duty-extension.yaml
+  StackName: !Sub '${resourcePrefix}-${appName}-extension'
+  StackDescription: GuardDuty - Extension (including IP Set)
```
ahayden, line 67: Then this would be `GuardDuty Extension`

org-formation/070-guard-duty/_tasks.yaml
```diff
@@ -0,0 +1,75 @@
...
+      ExcludeAccount:
+        - !Ref accountId
+    AllBinding:
+      Account: '*'
+      IncludeMasterAccount: true
+  Parameters:
+    resourcePrefix: !Ref resourcePrefix
+    accountId: !Ref accountId
+
+# bucket that contains the trusted IP addresses that are excluded from GuardDuty findings
```
ahayden, line 37: I suggest not adding any trusted ips list

ahayden (approved):
I think it's best to not include a trusted ips list, but it's up to you if you want to try it. `127.0.0.1` might be harmless today, though that might not always be the case as AWS is adding new analysis that might include loopback on EC2 or container runners, etc. at some point.

org-formation/070-guard-duty/_tasks.yaml
zaro0508: i think it's nice to have it as an option.  I mean if you don't put anything on the list then it's basically the same as not having it there at all.

org-formation/070-guard-duty/_tasks.yaml
zaro0508: it's optional.  it doesn't do anything if you don't put any values into trusted-ips.txt file.

PR #122, opened 2021-03-26
[IT-1101] Setup AWS Config

zaro0508:
Enable aws config on all accounts and have findings sent to our
logcentral account.  This is the master member config setup[1]

[1] https://docs.aws.amazon.com/organizations/latest/userguide/services-that-can-integrate-config.html

depends on https://github.com/Sage-Bionetworks/aws-infra/pull/289


org-formation/080-aws-config-inventory/_tasks.yml
```diff
@@ -0,0 +1,24 @@
...
+  DefaultOrganizationBindingRegion: !Ref primaryRegion
+  DefaultOrganizationBinding:
+    IncludeMasterAccount: true
+    Account: '*'
+  OrganizationBindings:
+    LogArchiveBinding:
+      Account: !Ref LogCentralAccount
+  Parameters:
+    resourcePrefix: !Ref resourcePrefix
+    bucketName: !Sub '${resourcePrefix}-${appName}-${CurrentAccount.AccountId}'
```
ahayden, line 24: What will be the the actual bucket names? Looks like logs for each account will be written to different buckets within logcentral. Is that how we want it for Sumo?

zaro0508: config logs works like cloudtrail.  there will be a bucket in each account with the above name, the distinction is the AccountId.  The bucket in logcentral will contain an aggregate of all accounts and that's the one that we will setup with Sumo.

ahayden (approved):
Looks good

ahayden (commented):
Yes, please

PR #123, opened 2021-03-26
[IT-1101] re-org OUs and SCPs

zaro0508:
Re organize accounts and org units to separate dev and prod accounts.
Update SCPs on OUs and accounts.  This brings in more inline with AWS best
practice for organizations[1]

[1] https://docs.aws.amazon.com/whitepapers/latest/organizing-your-aws-environment/recommended-ous.html

### DIFF
--- a/org-formation/organization.yaml
+++ b/org-formation/organization.yaml
@@ -18,16 +18,23 @@ Organization:
     Properties:
       DefaultOrganizationAccessRoleName: OrganizationAccountAccessRole
 
-  ScienceOU:
+  ScienceDevOU:
     Type: OC::ORG::OrganizationalUnit
     Properties:
-      OrganizationalUnitName: science
+      OrganizationalUnitName: ScienceDev
       Accounts:
         - !Ref ScipoolDevAccount
-        - !Ref ScipoolProdAccount
-        - !Ref ScicompAccount
         - !Ref SandboxAccount
         - !Ref MobileHealthDataEngineeringAccount
+        - !Ref SageITAccount
+
+  ScienceProdOU:
+    Type: OC::ORG::OrganizationalUnit
+    Properties:
+      OrganizationalUnitName: ScienceProd
+      Accounts:
+        - !Ref ScipoolProdAccount
+        - !Ref ScicompAccount
 
   PlatformOU:
     Type: OC::ORG::OrganizationalUnit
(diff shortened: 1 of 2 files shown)


org-formation/organization.yaml
```diff
@@ -18,16 +18,23 @@ Organization:
...
     Properties:
-      OrganizationalUnitName: science
+      OrganizationalUnitName: ScienceDev
       Accounts:
         - !Ref ScipoolDevAccount
-        - !Ref ScipoolProdAccount
-        - !Ref ScicompAccount
         - !Ref SandboxAccount
         - !Ref MobileHealthDataEngineeringAccount
+        - !Ref SageITAccount
```
ahayden, line 29: Should SageIT really be in this OU?

org-formation/organization.yaml
```diff
@@ -37,24 +44,29 @@ Organization:
         - !Ref SynapseOU
         - !Ref BridgeOU
 
-  ITOU:
+  ItDevOU:
+    Type: OC::ORG::OrganizationalUnit
+    Properties:
```
ahayden, line 49: This definition doesn't need an `OrganizationalUnits` property?

org-formation/organization.yaml
```diff
@@ -18,16 +18,23 @@ Organization:
...
         - !Ref SandboxAccount
         - !Ref MobileHealthDataEngineeringAccount
+        - !Ref SageITAccount
+
+  ScienceProdOU:
+    Type: OC::ORG::OrganizationalUnit
+    Properties:
+      OrganizationalUnitName: ScienceProd
+      Accounts:
+        - !Ref ScipoolProdAccount
```
ahayden, line 36: \+ `stridespool`?

ahayden (changes requested):
I think the design makes sense, but is missing a few entries?

org-formation/organization.yaml
zaro0508: SageITAccount currently has resources in us-west region and so that's why it's in this group.   Please propose something if you would like to see something different.

org-formation/organization.yaml
zaro0508: Nope, it can be a standalone OU with nothing in it.

org-formation/organization.yaml
zaro0508: stridespool  is not in our org we have no control

ahayden (approved):
The organization makes sense to me.

PR #124, opened 2021-03-26
[IT-1101] remove sceptre cloudtrail and config setup

zaro0508:
We are replacing the sceptre deployment of cloudtrail and config services with
org-formation deployment.  The advantage of the ofn deployment is trails will get
automatically created for every account we add to the org.


xschildw (approved):

ahayden (commented):
Given that `stridespool` will be not included in an OU, will org-f still deploy templates for services like CloudTrail? We need to make sure that service still runs in the strides accounts if sceptre isn't deploying any more

zaro0508:
hmm, it looks like cloudtrail is setup by the shared `essentials.yaml` file. I guess one way we can manage this is to make strides use a custom `essentials.yaml` instead of using the shared one.

zaro0508:
@ahayden we deploy the shared essentials.yaml without cloudtrail & config.  Then PR #128 will deploy cloudtrail & config into strides independently of essentials.yaml

ahayden (approved):
Seems fine. Sceptre will still deploy to strides accounts. I'm not sure why the bridge config was remnant in this change.

zaro0508:
@ahayden bridge was setup way before all other accounts were setup so it's a bit of a snowflake.  we need to keep it as-is due to legacy dependencies.  

I've also added the removal of AWS config to this PR.

ahayden (approved):
Added the removal of Config

PR #129, opened 2021-04-05
[SC-306] Create a kms key for lambda docker runner

zaro0508:
Create a KMS key that will be used to encrypt lambda ENV VARS
so that it does not get exposed to any user.  We only allow access
to the env vars by setting policies on this key[1]

[1] https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-encryption

### DIFF
--- a/sceptre/scipool/config/develop/docker-runner-kms-key.yaml
+++ b/sceptre/scipool/config/develop/docker-runner-kms-key.yaml
@@ -6,9 +6,10 @@ stack_tags:
   OwnerEmail: "it@sagebase.org"
 dependencies:
   - "develop/jumpcloud-idp.yaml"
+  - "develop/docker-runner-efs.yaml"
 parameters:
   KeyAdminArns:
-    - arn:aws:iam::804034162148:root
-    - !stack_output_external jumpcloud-idp::ItsandboxAdminSamlProviderRoleArn
+    - arn:aws:iam::465877038949:root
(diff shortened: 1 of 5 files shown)


sceptre/scipool/config/develop/docker-runner-kms-key.yaml
```diff
@@ -0,0 +1,14 @@
...
+stack_tags:
+  Department: "Platform"
+  Project: "Infrastructure"
+  OwnerEmail: "it@sagebase.org"
+dependencies:
+  - "develop/jumpcloud-idp.yaml"
+parameters:
+  KeyAdminArns:
+    - arn:aws:iam::804034162148:root
+    - !stack_output_external jumpcloud-idp::ItsandboxAdminSamlProviderRoleArn
```
ahayden, line 12: Is this the correct SAML entity? Should it not be something like `...ScipooldevAdmin...`

ahayden (approved):
Cool, looks good. I just left one question

sceptre/scipool/config/develop/docker-runner-kms-key.yaml
zaro0508: good catch.  it's fixed. thanks.

zimingd (approved):

xschildw (approved):

zaro0508:
we are going to use AWS batch instead of lambda.

PR #140, opened 2021-04-14
Allow more permissive regions on IT accounts and put SageIT in ItProdOU

ahayden:
- [x] This allows IT prod and dev accounts to have resources in multiple US regions, which is needed to allow resources that have already been created in us-west and could be useful in the future as long as IT is careful.


zaro0508 (approved):
this change allows more permissive access which LGTM.  If your ok with it then i'm ok.

ahayden:
@xschildw please review this... I believe we actually don't need to make an exception for allowing `us-west-1` any more.

xschildw (approved):
I don't see anything running in us-westx, looks good.

ahayden:
I believe we can just close this PR if there are no IT resources in us-westx

PR #142, opened 2021-04-14
Proposed structure

ahayden:
- [x] This structure would use OUs rather than tags to map AWS accounts to resource classifications
- [x] passes pre-commit


zaro0508 (commented):
Did you want to apply any SCP changes for this new configuration?

org-formation/organization.yaml
```diff
@@ -39,19 +38,22 @@ Organization:
         - !Ref ScicompAccount
         - !Ref iAtlasProdAccount
         - !Ref WorkflowsNextflowProdAccount
+#              StridesPoolAccount must be managed outside of orgf but should have parity with this OU
 
   PlatformOU:
```
zaro0508, line 43: do we need PlatformOU now that there's nothing in it?

zaro0508 (changes requested):

org-formation/organization.yaml
ahayden: I left the PlatformOU in to contain the SynapseProdOU and BridgeProdOU ... that makes sense, right?

zaro0508: oops, sorry.  i missed that.

org-formation/organization.yaml
```diff
@@ -72,23 +75,35 @@ Organization:
...
     Properties:
-      OrganizationalUnitName: synapse
+      OrganizationalUnitName: SynapseProd
       Accounts:
-        - !Ref SynapseDevAccount
         - !Ref SynapseDWAccount
         - !Ref SynapseProdAccount
 
-  BridgeOU:
+  SynapseDevOU:
```
zaro0508, line 86: These new OUs only inherit SCPs from the ORG root.  Did you want to apply any additional SCPs to these?

zaro0508 (changes requested):

org-formation/organization.yaml
```diff
@@ -39,19 +38,22 @@ Organization:
...
-        - !Ref BridgeOU
+        - !Ref SynapseProdOU
+        - !Ref BridgeProdOU
 
   ItDevOU:
     Type: OC::ORG::OrganizationalUnit
     Properties:
       OrganizationalUnitName: ItDev
+      OrganizationalUnits:
+        - !Ref PolicyStaging
```
xschildw, line 57: Should be PolicyStagingOU?

org-formation/organization.yaml
ahayden: I don't know of different SCPs we would want to deploy today, but this would let us do that in the future.

org-formation/organization.yaml
ahayden: Yes, you're right

ahayden: I made this change

org-formation/organization.yaml
```diff
@@ -72,23 +75,35 @@ Organization:
       Accounts:
         - !Ref ITSandboxAccount
 
-  SynapseOU:
+  SynapseProdOU:
```
zaro0508, line 78: changing SynapseOU and BridgeOU requires a corresponding change to SCP policy references, https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/cfc4327ae461a5877076e97714d3297fd5b07c54/org-formation/010-scps/_tasks.yaml#L19

we might just be able to remove them from the SCP file since they are under the `PlatformOU` which means that they will inherit SCP from the `PlatformOU`

org-formation/organization.yaml
```diff
@@ -64,6 +66,7 @@ Organization:
         - !Ref LogCentralAccount
         - !Ref SecurityCentralAccount
         - !Ref ImageCentralAccount
+        - !Ref SageITAccount
```
zaro0508, line 69: moving this from ScienceDev might be problematic because the SCP on `ItProdOU` is more restrictive than `ScienceDevOU`. https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/cfc4327ae461a5877076e97714d3297fd5b07c54/org-formation/010-scps/_tasks.yaml#L68 

@xschildw might know better but I think we placed it in ScienceDev because there's some resources in that account that are in US regions other than us-east-1

zaro0508 (changes requested):

org-formation/organization.yaml
ahayden: I believe the only resource outside of `us-east-1` in that account is offline... and could be terminated per https://sagebionetworks.jira.com/browse/IT-447

org-formation/organization.yaml
ahayden: Good catch... I agree that removing `SynapseOU` and `BridgeOU` should be removed and the SCPs applied via `PlatformOU` is best

org-formation/organization.yaml
```diff
@@ -39,19 +38,22 @@ Organization:
...
 
   PlatformOU:
     Type: OC::ORG::OrganizationalUnit
     Properties:
       OrganizationalUnitName: Platform
       OrganizationalUnits:
-        - !Ref SynapseOU
-        - !Ref BridgeOU
+        - !Ref SynapseProdOU
+        - !Ref BridgeProdOU
```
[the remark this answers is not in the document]
ahayden: I see what you mean... maybe a `PlatformDevOU` makes sense

org-formation/organization.yaml
zaro0508: @ahayden do you plan on making the changes to the SCPs to match the changes in this organizations file?

ahayden:
Closing as choosing not to do

PR #143, opened 2021-04-15
Allow Admin role to archive GD findings in securitycentral

ahayden:
@zaro0508 will this change override the orgf SCP that denies archive on GD findings?

- [x] This fix is to allow IAM entities with the admin role to archive findings from the GD dashboard, so we can know if they are handled or not
- [x] passes pre-commit


zaro0508:
SCPs override IAM policies.  What you should probably do is exclude the SCP from the securitycentral account.

ahayden:
Hmm not sure I like that solution though...

PR #147, opened 2021-04-17
Suggest users to prune and add deny unless MFA

ahayden:
- [x] This PR suggests which IAM users to prune. We should wait for the individuals to confirm that they don't need their user entity before merging. 
- [x] Passes pre-commit


xschildw (changes requested):
I think the disable MFA is required for the builds. The IAM users are essentially service accounts for the builds. Technically we could remove console access and access through IDP (the delete KMS is handled by a Jenkins job, but we're missing permissions to do IAM operations with devs being PowerUsers).

zaro0508 (approved):
isn't the purpose of this PR to switch to accessing AWS with JC?  if so, then why not just remove all human IAM users?

ahayden:
I've deleted resources attached to the IAM users we want to delete, so this can merge now if @xschildw is OK with re-enforcing the MFA requirement.

ahayden:
> isn't the purpose of this PR to switch to accessing AWS with JC? if so, then why not just remove all human IAM users?

I think the SynapseDev workflow isn't ready to switch to JC login just yet... the purpose of this PR is to force MFA again and prune IAM users that don't need access

ahayden:
OK, I've re-commented out the MFA deny policy

PR #152, opened 2021-04-27
decouple root OU from all OUs.

zaro0508:
Applying a SCP to the org root OU means applying it to all OUs
and all accounts in the org.  Sometimes you may want to apply
to all OUs but not all accounts.  This PR separates applying
of SCP from root OU and all OUs.  This will make it more
flexible for us to shift SCPs around.

Also one of the AWS organization best practice is to apply SCP
guardrails to individual OUs.


ahayden (approved):
`AllOUs` is a little misleading if it doesn't have the root OU in it...

PR #154, opened 2021-04-27
pin shared template versions

zaro0508:
Pinning versions makes it safer to update shared templates.  This only
pins org-formation usage of the shared templates.  We should make Sceptre
use pinned versions as well but will save that for a separate PR.

### DIFF
--- a/org-formation/010-scps/_tasks.yaml
+++ b/org-formation/010-scps/_tasks.yaml
@@ -3,7 +3,7 @@ Parameters:
 
 PreventDisableGuardDuty:
   Type: update-stacks
-  Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/master/templates/SCP/prevent-disable-guardduty.yaml
+  Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.2.4/templates/SCP/prevent-disable-guardduty.yaml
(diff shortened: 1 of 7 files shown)


ahayden (commented):
I could see this as becoming a maintenance problem. Should the version string, like `v0.2.4` be the same across all these templates, always? If so, maybe it makes more sense to do a rotating deployment like red-green-blue, rather than a version string

zaro0508:
no, it does not need to be the same across all templates.  you need to pick and choose the version you want to use.  I set them all to `v2.4.0` because I just created `v2.4.0` off master.  

this is pretty standard workflow for any app that has dependencies.  Yes, it's takes more work to do the updating when you make changes to templates however it's not more work if templates don't change very often.

ahayden:
What I mean is, unless there's a reason to use v0.2.4 of a release in `aws-infra` as a base template for one section and v0.2.5 for a different section, it would be simpler to just rotate through release tags. I'm OK with any method, and long as we can keep current versions clear.

zaro0508:
it's designed for both workflows (rotate release tags or pick and choose) @ahayden, it's really up to you.  if there are no other issues can we get an approval to move forward?

ahayden (approved):
Yes, let's use vNNNN tags

PR #155, opened 2021-04-27
[IT-1101] Setup GuarDuty

zaro0508:
Enable guard duty on specific accounts and have findings sent to our
securitycentral account.  This is the master member guard duty
configuration[1]

This is a redo of PR #120, 1st attempt failed due to SCP preventing
guardduty changes.

[1] https://docs.aws.amazon.com/guardduty/latest/ug/guardduty_organizations.html

depends on https://github.com/Sage-Bionetworks/aws-infra/pull/288 #159

PR #159, opened 2021-04-29
[IT-1101] Remove Guardduty from sceptre management

zaro0508:
We are planning to move management of Guardduty from sceptre to
org-formation.  We need to remove the sceptre deployment before
setting up a new Guardduty with ofn which is done in PR #155


sceptre/scipool/config/strides/guardduty-member.yaml
```diff
@@ -1,10 +0,0 @@
-template_path: remote/GuardDutyMember.yaml
-stack_name: guardduty-member
-dependencies:
-  - strides/essentials.yaml
```
ahayden, line 4: I believe we'll still need sceptre to deploy DG to strides accounts... True?

ahayden (changes requested):
Looks ok, but one question on STRIDES.

sceptre/scipool/config/strides/guardduty-member.yaml
zaro0508: i guess if you want keep guardduty on strides then yes.  I have restored the file.

xschildw (approved):

sceptre/scipool/config/strides/guardduty-member.yaml
zaro0508: it turns out that this file sets up guardduty member with the guardduty master that was setup with sceptre.  since we are going to setup a new master this will no longer work.  We need to remove then re-add and associate with a new master.

PR #160, opened 2021-04-29
Setup CI service user/role for iAtlas account

zaro0508:
This is to setup a AWS service user/role to deploy CFN templates to
the iAtlas AWS account.


ahayden (approved):
Yes, this is good

PR #164, opened 2021-05-04
Remove unused VPCs.

zaro0508:
VPCs cost money, if we don't need them we should remove them.


ahayden (approved):
Seems fine if you've confirmed they are unused

xschildw (approved):

PR #165, opened 2021-05-04
[IT-1101] Allow access to cloudtrail buckets

zaro0508:
Some of our accounts, like org-sagebase-strides is not part of the Sage
organziations account therefore we need to manage them as external
entities.  We need to setup the master cloudtrail bucket in logcentral
to allow external accounts to send cloudtrail logs to the master
bucket.

depends on https://github.com/Sage-Bionetworks/aws-infra/pull/298


sceptre/logcentral/config/prod/ConfigLogAggregation.yaml
```diff
@@ -0,0 +1,7 @@
+template_path: ConfigLogAggregation.j2
+stack_name: ConfigLogAggregation
+sceptre_user_data:
+  stackname: !stack_name
+  config_bucket: !stack_output_external sagebase-cloudtrail::sagebaseDashcloudtrailDashCloudTrailBucket
```
ahayden, line 5: Is this bucket name correct? Is there a different `sagebase-clonfig::sagebaseDashconfigtrailDashConfigTrailBucket` output?

ahayden (changes requested):
Need to double check the bucket names

sceptre/logcentral/config/prod/ConfigLogAggregation.yaml
zaro0508: good catch.  I've updated this to reference the master config bucket.  I also changed this PR to deploy with ofn instead of sceptre.

PR #168, opened 2021-05-07
[IT-1178] AWS SSO

zaro0508:
Setup AWS SSO and integrate it with Jumpcloud IDP.  This allows
us to manage AWS account access from JC and more.

Benefits:
* Simplifies management of JC SSO applications.  Only need one JC
  app per AWS organization.
* Jumpcloud users and groups are synced to AWS SSO with SCIM which
  should make onboarding and offboarding of users much easier.
* Allows users to easily request temporary AWS access tokens from the
  SSO portal.
* Works with the AWS sso CLI[1] which allows easy programmatic
  access to accounts.  This may also allow ssh/rdp access to private
  EC2s using the AWS CLI.

[1] https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-sso.html

### DIFF
--- a/org-formation/700-aws-sso/README.md
+++ b/org-formation/700-aws-sso/README.md
@@ -1,0 +1,19 @@
+### Purpose of these templates
+The templates in this folder enable [AWS SSO](https://docs.aws.amazon.com/singlesignon/latest/userguide/what-is.html) in the Management account. SSO is used to enable human access to accounts within the organization, both to the console as well as through the cli.
+
+This is one of the few folders that does not come with '**batteries included**' as it requires you to manually create Groups and Users. By default it comes with four profiles:
+1. Auditor: tied to the default policy `arn:aws:iam::aws:policy/SecurityAudit`
+2. Developer: tied to the default policy `arn:aws:iam::aws:policy/PowerUserAccess`
+3. ScienceSupporter: tied to the default policies `arn:aws:iam::aws:policy/PowerUserAccess` and `arn:aws:iam::aws:policy/job-function/ViewOnlyAccess`
+4. Administrator: tied to the default policy `arn:aws:iam::aws:policy/AdministratorAccess`
+
+You will notice that there are more than 4 tasks, that is because for production access we want to limit the session time to 1 hour, while for non-production it can be 12 hours. This can only be done by deploying different permission sets.
+
+Some notable resources in this folder:
+
+| Resource | Description |
+| - | - |
+| Managed policies | An example stack to design your own IAM policies to assign to Groups |
+| SSO | Deploys a set of permissions (roles basically) and connects that to groups that you create earlier |
+
+It is recommended to re-use AWS provided policies where possible because of the maintenance benefit. The AWS environment constantly changes and AWS ensures that the managed policies they provide give access to new AWS Services and new actions within existing services.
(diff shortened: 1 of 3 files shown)


org-formation/700-aws-sso/_tasks.yaml
```diff
@@ -0,0 +1,171 @@
+Parameters:
```
brucehoff, line 1: Is this an org-formation config' file?  What would I read up on to have the knowledge to review this?

org-formation/700-aws-sso/README.md
```diff
@@ -0,0 +1,25 @@
...
+   [JC instructions](https://support.jumpcloud.com/support/s/article/Single-Sign-On-SSO-With-AWS-SSO)
+5. Copy the "AWS SSO Sign-in URL" from AWS SAML 2.0 authentication and paste it into the JC SSO setting "Login URL"
+6. In JC SSO setting "Enable management of User Groups and Group Membership in this application" to allow
+   JC to manage AWS SSO user and groups with SCIM.
+7. In the AWS SSO identity source setting select "Enable automatic provisioning" then
+   copy the "SCIM endpoint" and "Access Token" from AWS and paste it into the JC SCIM settings "SP base URL" and
+   "SP API Token" fields.
+8. Setup JC users/groups and give access to JC SSO app.
+9. Check that JC users/groups are automatically synced to AWS SSO
+10. Verify that JC app from the JC user login portal has access to the AWS account(s) and can sign into the
```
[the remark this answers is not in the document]
zaro0508: yes, info on JC portal looks like this  https://pasteboard.co/K0Q8PVEF.png while the AWS SSO portal looks something like this https://pasteboard.co/K0Q795y.png

org-formation/700-aws-sso/_tasks.yaml
zaro0508: yes, it's an org-formation template.  docs for org-formation are here https://github.com/org-formation/org-formation-cli

the example templates to setup SSO used for this PR are here  https://github.com/org-formation/org-formation-reference/tree/master/src/templates/100-aws-sso

brucehoff (approved):

org-formation/700-aws-sso/_tasks.yaml
```diff
@@ -0,0 +1,171 @@
...
+  DefaultOrganizationBinding:
+    IncludeMasterAccount: true
+  OrganizationBindings:
+    TargetBinding:
+      OrganizationalUnit:
+        - !Ref ScienceDevOU
+        - !Ref ScienceProdOU
+  Parameters:
+    instanceArn: !Ref instanceArn
+    principalId: !Ref scienceSupporterGroup
```
ahayden, line 69: I don't think this group is granular enough for the two OUs... what if we want to give admin to some accounts within ScienceDevOU, but not all, or all admin to ScienceDevOU but not ScienceProdOU? I think we might need all permutations of a role like this for every account

org-formation/700-aws-sso/_tasks.yaml
```diff
@@ -0,0 +1,171 @@
...
+  TerminationProtection: false
+  DefaultOrganizationBindingRegion: !Ref primaryRegion
+  DefaultOrganizationBinding:
+    IncludeMasterAccount: true
+  OrganizationBindings:
+    TargetBinding:
+      OrganizationalUnit:
+        - !Ref ScienceDevOU
+        - !Ref ScienceProdOU
+      Account: '*'
```
ahayden, line 107: Doesn't `Account: '*'` take precedent over the OU binding?

PR #173, opened 2021-05-11
[IT-1178] Remove Jupcloud AWS SAML integrations

zaro0508:
We are replacing the JC AWS SAML integrations with AWS SSO.  This removes
the SAML integrations on AWS that have already been removed from JC.
The aws-sagebase-organizations JC app provides equivalent access to AWS
accounts.

One caveat, we need to keep jumpcloud SAML for strides because strides account
is not in our AWS ORG therefore we can't manage it using ofn.


brucehoff (approved):

ahayden (approved):
yesss

PR #178, opened 2021-05-12
[IT-1178] Setup SSO access to AWS imagecentral account

zaro0508:
Setup to allow users to deploy AMIs to imagecentral account (our shared
AWS AMI repository).

The purpose of the ami librarian group is to allow developers to execute
packer which deploys and removes AMIs from imagecentral account. This
allows developers to test their packer templates.


org-formation/700-aws-sso/_tasks.yaml
```diff
@@ -30,6 +30,28 @@ Parameters:
     Type: String
     Default: '906769aa66-5d23a723-54f3-4c08-a67b-311e555f4e85'
 
+  imageCentralAmiLibrarianGroup:
+    Type: String
+    Default: '906769aa66-8bfbe918-2d30-467e-9e98-92232e9f8410'
```
brucehoff, line 35: where does this string ('906769aa66-8bfbe918-2d30-467e-9e98-92232e9f8410') come from?

zaro0508: it's the AWS SSO group ID which is generated by AWS.  You can find it on the AWS SSO configuration and it looks like this.. https://pasteboard.co/K1BlqLw.png

brucehoff: I see:  Elsewhere when we have one CF script to create a resource we then pass the generated ID (for example, an ARN) automatically to the CF script(s) that use it.  Why don't we do that here rather than hard coding the ID into the script?

zaro0508: Because we don't generate those group IDs in our templates so we don't have a reference to them.  That's why we need to hard code them into the template.  We use the [jumpcloud SCIM feature](https://jumpcloud.com/blog/scim-provisioning-defined) and that is the thing that triggers AWS to generate the IDs.

ahayden (approved):
Looks good. I think EC2FullAccess is appropriate

PR #181, opened 2021-05-13
[IT-1178] Setup SSO access to AWS bridge accounts

zaro0508:
Setup to allow users in the JC bridge admin, developer and iOS developer
groups access to AWS bridge dev and bridge prod accounts.


brucehoff (approved):

org-formation/700-aws-sso/_tasks.yaml
```diff
@@ -30,6 +30,153 @@ Parameters:
...
+            "Resource": "arn:aws:s3:::ios-apps.sagebridge.org/*"
+          }
+        ]
+      }
+  DefaultOrganizationBindingRegion: !Ref primaryRegion
+  DefaultOrganizationBinding:
+    IncludeMasterAccount: true
+  OrganizationBindings:
+    TargetBinding:
+      Account: !Ref MasterAccount
```
ahayden, line 159: This target should be a bridge acct

ahayden (changes requested):
I think one account target is wrong

org-formation/700-aws-sso/_tasks.yaml
zaro0508: fixed by replacing this block with `inlinePolicy` in SsoBridgeProdIosDeveloper

ahayden (approved):
Looks good

PR #183, opened 2021-05-14
[IT-1205] deploy VPC request authorizer role

zaro0508:
Create a VPC peering request role to allow account to automatically accept
peering requests . We did this same thing in sceptre with essentials.yaml[1]

[1] https://github.com/Sage-Bionetworks/aws-infra/blob/master/templates/essentials.yaml#L92

depends on https://github.com/Sage-Bionetworks/aws-infra/pull/304

PR #184, opened 2021-05-14
PLFM-1221/PLFM-1223: setup pcx to nextflow* and nlpsandbox

xschildw:
This PR creates peering connections from the VPN vpc to the VPCs for nextflow and nlpsandbox.
It uses the authorizer roles defined in https://github.com/Sage-Bionetworks-IT/organizations-infra/pull/183.


sceptre/admincentral/config/prod/peering-nextflow-dev.yaml
```diff
@@ -0,0 +1,7 @@
+template_path: VPCPeer.yaml
+stack_name: peering-nextflow-dev
+parameters:
+  PeerVPC: vpc-04c1457103adc1fdf
+  PeerVPCOwner: 035458030717
```
zaro0508, line 5: i think you need to put quotes around numbers.  also can you please add comments like the ones in peering-iatlasvpc.yaml?  https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/sceptre/admincentral/config/prod/peering-iatlasvpc.yaml

zaro0508 (changes requested):

ahayden (approved):
This looks good

PR #185, opened 2021-05-17
IT-1221: Add support for /20 CIDR

xschildw:
This PR should fix the validation error in https://github.com/Sage-Bionetworks-IT/organizations-infra/runs/2603419390?check_suite_focus=true.

### DIFF
--- a/sceptre/admincentral/templates/VPCPeer.yaml
+++ b/sceptre/admincentral/templates/VPCPeer.yaml
@@ -10,11 +10,11 @@ Parameters:
     Description: VPC ID in the peer account (i.e. vpc-1234abcd)
     Type: String
   PeerVPCCIDR:
-    Description: CIDR of the VPC in the peer account (i.e. 10.17.0.0/16 or 10.255.17.0/20 or 10.255.17.0/24)
+    Description: CIDR of the VPC in the peer account (i.e. 10.17.0.0/16)
     Type: String
     MinLength: '9'
     MaxLength: '18'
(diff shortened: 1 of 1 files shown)


sceptre/admincentral/templates/VPCPeer.yaml
```diff
@@ -10,11 +10,11 @@ Parameters:
     Description: VPC ID in the peer account (i.e. vpc-1234abcd)
     Type: String
   PeerVPCCIDR:
-    Description: CIDR of the VPC in the peer account (i.e. 10.17.0.0/16 or 10.255.17.0/24)
+    Description: CIDR of the VPC in the peer account (i.e. 10.17.0.0/16 or 10.255.17.0/20 or 10.255.17.0/24)
     Type: String
     MinLength: '9'
     MaxLength: '18'
-    AllowedPattern: '((\d{1,3})\.(\d{1,3})\.0\.0/16)|((\d{1,3})\.(\d{1,3})\.(\d{1,3})\.0/24)'
+    AllowedPattern: '((\d{1,3})\.(\d{1,3})\.0\.0/16)|((\d{1,3})\.(\d{1,3})\.(\d{1,3})\.0/20)|((\d{1,3})\.(\d{1,3})\.(\d{1,3})\.0/24)'
```
zaro0508, line 17: making it too specific might be difficult to maintain.  how about doing a general CIDR match?  example here, https://regexr.com/5t4b7

zaro0508 (changes requested):

sceptre/admincentral/templates/VPCPeer.yaml
xschildw: Sure, my reasoning was the opposite: I thought that restricting to large (/16), medium (/20) and small (/24) would make it easier to track the CIDRs, but it can be done somewhere else.

zaro0508 (approved):

PR #186, opened 2021-05-17
Guardduty for strides account

zaro0508:
Setup guardduty member in AWS strides account sending it's finding
to the master securitycentral account.  We setup with sceptre because
strides is no in our ORG therefore it cannot be managed by
org-formation.

### DIFF
--- a/sceptre/strides/config/prod/guardduty.yaml
+++ b/sceptre/strides/config/prod/guardduty.yaml
@@ -2,5 +2,4 @@
 template_path: "guardduty.yaml"
 stack_name: "guardduty"
 parameters:
-  resourcePrefix: "sagebase"
   accountId: "140124849929"       # org-sagebase-securitycentral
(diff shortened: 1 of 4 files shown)


sceptre/strides/config/prod/guardduty.yaml
```diff
@@ -0,0 +1,6 @@
+# setup AWS guard duty member
+template_path: "guardduty.yaml"
+stack_name: "guardduty"
+parameters:
+  resourcePrefix: "sagebase"
```
ahayden, line 5: Where is this used?

ahayden (approved):
This is good as long as we don't need the `resourcePrefix` to categorize things

sceptre/strides/config/prod/guardduty.yaml
zaro0508: it's been removed

xschildw (approved):

ahayden (approved):
Looks good

ahayden (approved):

xschildw (approved):

PR #188, opened 2021-05-18
Sumologic access to config and cloudtrail buckets

zaro0508:
Setup roles to allow Sumologic to ingest cloudtrail and config data so
we can use Sumo to easily query the data.

depends on https://github.com/Sage-Bionetworks/aws-infra/pull/305

PR #192, opened 2021-05-19
Revert "Sumologic access to config and cloudtrail buckets (#188)"

zaro0508:
This reverts commit 77711558c385214bbca6ad58796387c4786a0dc3.

The sumo template adds the strides account number to the bucket policy
which breaks it from allowing other accounts to write logs
to the bucket.


zaro0508:
evaluation on this is incorrect.

PR #195, opened 2021-05-21
PLFM-6702: Add Athena permissions for read-only users

xschildw:
Need to check with Marco...


zaro0508:
what is this about?  could you please provide more info in commit PR description?

xschildw:
Meant to discuss with Marco before submitting (hence Draft). This issue is about giving read-only users the right to interact with Athena. It can't destroy the underlying S3 data but it's still still a lot (too much?) permissions on the Athena schemas etc. .

PR #196, opened 2021-05-21
[SC-320] Product version per environment

zaro0508:
This is a refactor to de-couple SC product versions from the environment
that it gets deployed to.  The goal is to allow developers to add
a product to one environment at a time.  For example when I'm developing
a new SC product I may want to merge a change that only deploys to the
development environment for testing.

To pass in ProvisioningArtifactParameters from sceptre template we had to
change to using jinja templates.


sceptre/scipool/config/develop/sc-product-ec2-linux-jumpcloud-notebook.yaml
```diff
@@ -9,6 +9,63 @@ dependencies:
...
+      Name: 'v1.1.12'
+    - Description: 'Fix policies'
+      Info:
+        LoadTemplateFromURL: 'https://{{stack_group_config.admincentral_cf_bucket}}.s3.amazonaws.com/{{stack_group_config.service_catalog_library}}/v1.1.13/ec2/sc-ec2-linux-jumpcloud-notebook.yaml'
+      Name: 'v1.1.13'
+    - Description: 'Restore set_env_var.'
+      Info:
+        LoadTemplateFromURL: 'https://{{stack_group_config.admincentral_cf_bucket}}.s3.amazonaws.com/{{stack_group_config.service_catalog_library}}/v1.1.14/ec2/sc-ec2-linux-jumpcloud-notebook.yaml'
+      Name: 'v1.1.14'
+    - Description: 'Synapse Credential passthrough. {{ range(1, 10000) | random }}'
```
ahayden, line 68: I don't think this needs to be a random number, and setting it to a random value could be troublesome. Why not add a "tag" to the description? Just append `[LATEST]` to the description and remove the "tag" later when we release a new version? Changing the description string should force an update.

sceptre/scipool/templates/sc-product-ec2-linux-jumpcloud-notebook.j2
```diff
@@ -0,0 +1,67 @@
...
+      Name: !Ref ProductName
+      Description: This product builds one Linux EC2 instance using an Rstudio
+        AMI based on Ubuntu and maintained by Sage Bionetworks.
+      ProvisioningArtifactParameters:
+        {{ sceptre_user_data.ProvisioningArtifactParameters|indent(8) }}
+      'Fn::Transform':
+        Name: 'AWS::Include'
+        Parameters:
+          # source: https://github.com/Sage-Bionetworks/admincentral-infra/blob/master/templates/cfn-snippets-bucket.yaml
+          Location: "s3://cfn-snippets-bucket-cloudformationsnippetsbucket-elu83sv8ocdz/scipool/products.yaml"
```
ahayden, line 31: Couldn't this be parameterized?

sceptre/scipool/templates/sc-product-s3-private-enc.j2
```diff
@@ -0,0 +1,47 @@
...
+        Parameters:
+          Location: "s3://cfn-snippets-bucket-cloudformationsnippetsbucket-elu83sv8ocdz/scipool/products.yaml"
+      ReplaceProvisioningArtifacts: true
+  Associates3:
+    Type: AWS::ServiceCatalog::PortfolioProductAssociation
+    Properties:
+      PortfolioId: !ImportValue
+        'Fn::Sub': '${AWS::Region}-sc-portfolio-s3-basic-SCS3portfolioId'
+      ProductId: !Ref scs3product
+  constraintec2linux:
```
ahayden, line 34: It looks like the launch role and portfolio id are correct... this constraint should be named differently though, for clarity.

ahayden (approved):
Approving, though commented a few times on parts that I think make it better

sceptre/scipool/templates/sc-product-ec2-linux-jumpcloud-notebook.j2
zaro0508: yes, i'm not opposed to that but it wasn't before this refactor.  Let's keep as is for this refactor then parameterize on a follow on change?

sceptre/scipool/templates/sc-product-s3-private-enc.j2
zaro0508: this is the same as before this refactor so how about leave this for a follow on PR?

brucehoff (approved):

PR #197, opened 2021-05-21
[IT-1148] remove account names

zaro0508:
We are resetting AWS account names. If we change the name in the
organization.yaml file ofn will try to rename the account and fail.
To get in sync we need to stop ofn from managing the account name.

Once accounts names have been manually updated we can go back and
add the corresponding names into organization.yaml again.


xschildw (approved):
Once this is merged, the names in the Your

PR #200, opened 2021-05-24
[IT-1178] Update Bridge prod SSO policy

zaro0508:
The existing Bridge prod SAML policy restricts develops from being able to
delete RDS and Dynamo DB tables[1].  This change adds that restriction.

[1] https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/sceptre/bridge/templates/iam-policies.yaml#L30-L56


ahayden (approved):
Looks good. This is a model for other PU access

PR #201, opened 2021-05-24
[IT-1178] Update Scicomp SSO viewer policy

zaro0508:
The current Scicomp SAML integration setup has a custom policy[1] for
the viewers, it's a viewer job function policy[2] and view other things
role.  We setup Scicomp SSO to match.

[1] https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/sceptre/scicomp/templates/jumpcloud-idp.yaml#L74-L85
[2] https://docs.aws.amazon.com/IAM/latest/UserGuide/access_policies_job-functions.html#security-iam-awsmanpol-jobfunction-updates


ahayden (approved):
Looks good. support user can open support tickets.

PR #202, opened 2021-06-02
Update jumpcloud SAML role to SSO role

zaro0508:
We switch to using AWS SSO and removed SAML integration therefore
the SAML roles are gone.  We replace with the AWS generated SSO role.

We need to hard code the value because the SSO role is created by
the organizations account where AWS SSO runs.


zaro0508:
merging to fix the build.

PR #203, opened 2021-06-02
Allow JC users access to initiate SSM sessions

zaro0508:
Setup a SSO role to allow JC users to initiate SSM sessions so that they
can access EC2 instances using the AWS SSM sessions.  It will also allow
users to setup port forwarding for their apps running in private EC2
instances.

### DIFF
--- a/org-formation/700-aws-sso/_tasks.yaml
+++ b/org-formation/700-aws-sso/_tasks.yaml
@@ -126,6 +126,10 @@ Parameters:
     Type: String
     Default: '906769aa66-8f870ac5-9280-4c19-b477-461592fe92a6'
 
+  AwsSsmSessionInitiatorGroup:    #JC aws-ssm-session-initiators
+    Type: String
+    Default: '906769aa66-33f17c18-7b59-4aa2-9334-91dfce046f63'
+
 SsoAdministrator:
   Type: update-stacks
   Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.2.6/templates/SSO/aws-sso.yaml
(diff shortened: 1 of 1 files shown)


brucehoff (approved):

org-formation/700-aws-sso/_tasks.yaml
```diff
@@ -266,6 +270,67 @@ SsoViewerSupporter:
...
+      Account:
+        - !Ref ScipoolDevAccount
+        - !Ref ScipoolProdAccount
+        - !Ref ScicompAccount
+  Parameters:
+    instanceArn: !Ref instanceArn
+    principalId: !Ref AwsSsmSessionInitiatorGroup
+    permissionSetName: 'ssm-session-initiator'
+    sessionDuration: 'PT12H'
+    inlinePolicy: >-
```
zaro0508, line 293: @ahayden This is for JC users only and SC users can only see the instance IDs of their own provisioned instances so how about we give this role ssm session access to all resources?

xschildw (approved):

org-formation/700-aws-sso/_tasks.yaml
ahayden: I don't think we want to do that... the instanceid is not a secret and shouldn't be used for authorization. If you remove the condition to allow start-session only on resources tagged for the resource owner, we are essentially giving root to all instances for all service catalog users.

ahayden (changes requested):
The AccessApprovedCaller tag name will only match when a user logs in via Synapse IDP (will be different for JC SAML login)

zaro0508:
bad idea so closing.

PR #207, opened 2021-06-05
revert-aws-marketplace-changes

brucehoff:
Remove the DynamoDB database used by the experimental AWS marketplace integration.

Depends on Sage-Bionetworks/synapse-login-aws-infra#61
Depends on Sage-Bionetworks-IT/cfn-cr-synapse-tagger#28
Depends on #208


xschildw (approved):

zaro0508 (changes requested):
if we remove this dynamo db we will also need to remove the following items..

The [sc cost lambda](https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/sceptre/scipool/config/develop/lambda-sc-cost-meter.yaml) which assumes the existence of this db 

from https://github.com/Sage-Bionetworks-IT/lambda-sc-cost-meter/blob/master/README.md  
```
.. This app depends on an existing dynamo db with valid customer data.
```

We need to remove the parts of the [synapse tagger lambda](https://github.com/Sage-Bionetworks-IT/cfn-cr-synapse-tagger) that also depends on this dynamo db..
```
.. Marketplace SC product tags - Get Marketplace SC product code and customer ID from subscriber Dynamo DB and apply them as tags.
```

brucehoff:
> if we remove this dynamo db we will also need to remove the following items..
> 
> The [sc cost lambda](https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/sceptre/scipool/config/develop/lambda-sc-cost-meter.yaml) which assumes the existence of this db
> 
> from https://github.com/Sage-Bionetworks-IT/lambda-sc-cost-meter/blob/master/README.md
> 
> ```
> .. This app depends on an existing dynamo db with valid customer data.
> ```
> 
> We need to remove the parts of the [synapse tagger lambda](https://github.com/Sage-Bionetworks-IT/cfn-cr-synapse-tagger) that also depends on this dynamo db.
> 
> from https://github.com/Sage-Bionetworks-IT/cfn-cr-synapse-tagger/blob/master/README.md
> 
> ```
> .. Marketplace SC product tags - Get Marketplace SC product code and customer ID from subscriber Dynamo DB and apply them as tags.
> ```

Thanks, @zaro0508 .  I created this PR to update the 'tagger':
https://github.com/Sage-Bionetworks-IT/cfn-cr-synapse-tagger/pull/28
... and made this PR dependent on that one.

How shall we remove the lambda-sc-cost-meter?  Do we have to delete the entire repoo'?  I'd rather not if possible, since it would be nice to preserve this work for the future.

zaro0508:
> How shall we remove the lambda-sc-cost-meter? Do we have to delete the entire repoo'? I'd rather not if possible, since it would be nice to preserve this work for the future.

we remove it very carefully.  no need to delete entire repo.

PR #209, opened 2021-06-09
Update AMI Librarian access

zaro0508:
Packer will create EC2 instances so it to make AMIs.  The AMI librarian
needs to be able to start SSM sessions to EC2 instances on the
ImageCentral account to allow debugging the instances.


xschildw (approved):
Sounds good.

brucehoff (approved):

PR #215, opened 2021-06-16
[SC-251] allow SSM AWS-StartInteractiveCommand & AWS-StartSSHSession docs

zaro0508:
Allow SC users to use the SSM start-session and AWS-StartSSHSession
command to start a session with a specific command. This, for example,
will let the user start a session with the `bash` shell instead of the default `sh`
shell.  The command would be..

```
aws ssm start-session --target <INSTANCE_ID> \
                      --document-name AWS-StartInteractiveCommand
                      --parameters command="bash -l"
```
more info here: https://aws.amazon.com/blogs/mt/limit-ssm-interactive-session-commands-by-users/

The AWS-StartSSHSession allows users to start an SSH session with
which will allow users to secure copy from local machine and their instance.

more info here: https://docs.aws.amazon.com/systems-manager/latest/userguide/session-manager-getting-started-enable-ssh-connections.html

documentation update here: https://sagebionetworks.jira.com/wiki/spaces/IT/pages/938836322/Service+Catalog+Provisioning#SSM-access-with-custom-commands

### DIFF
--- a/sceptre/scipool/templates/sc-enduser-iam.yaml
+++ b/sceptre/scipool/templates/sc-enduser-iam.yaml
@@ -66,7 +66,9 @@ Resources:
           - Effect: Allow
             Action:
               - ssm:StartSession
-            Resource: !Sub "arn:aws:ec2:${AWS::Region}:${AWS::AccountId}:instance/*"
+            Resource:
+              - !Sub "arn:aws:ssm:*:${AWS::AccountId}:document/AWS-StartInteractiveCommand"
+              - !Sub "arn:aws:ec2:${AWS::Region}:${AWS::AccountId}:instance/*"
             Condition:
               StringLike:
                 "ssm:ResourceTag/Protected/AccessApprovedCaller": "${aws:userid}"
--- a/sceptre/scipool/templates/sc-enduser-iam.yaml
+++ b/sceptre/scipool/templates/sc-enduser-iam.yaml
@@ -67,6 +67,7 @@ Resources:
             Action:
               - ssm:StartSession
             Resource:
+              - !Sub "arn:aws:ssm:*:${AWS::AccountId}:document/AWS-StartSSHSession"
               - !Sub "arn:aws:ssm:*:${AWS::AccountId}:document/AWS-StartInteractiveCommand"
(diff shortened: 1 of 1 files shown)


tthyer (approved):
We also need `arn:aws:ssm:*:*:document/AWS-StartSSHSession`

zaro0508:
> We also need `arn:aws:ssm:*:*:document/AWS-StartSSHSession`
> See https://docs.aws.amazon.com/systems-manager/latest/userguide/session-manager-getting-started-enable-ssh-connections.html

hmm, that thing seems to require a PEM file which will not work because we don't have ssh key pair access setup on SC instances. From the docs.. ```Create or verify that you have a Privacy Enhanced Mail certificate (a PEM file), or at minimum a public key, to use when establishing connections to managed instances. This must be a key that is already associated with the instance. ```

sceptre/scipool/templates/sc-enduser-iam.yaml
```diff
@@ -66,7 +66,10 @@ Resources:
           - Effect: Allow
             Action:
               - ssm:StartSession
-            Resource: !Sub "arn:aws:ec2:${AWS::Region}:${AWS::AccountId}:instance/*"
+            Resource:
+              - !Sub "arn:aws:ssm:*:${AWS::AccountId}:document/AWS-StartSSHSession"
+              - !Sub "arn:aws:ssm:*:${AWS::AccountId}:document/AWS-StartInteractiveCommand"
+              - !Sub "arn:aws:ec2:${AWS::Region}:${AWS::AccountId}:instance/*"
```
brucehoff, line 72: The "*" at the end of the ARN pattern makes me wonder what keeps one user from accessing another user's machine.

zaro0508: the restriction is here: https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/sceptre/scipool/templates/sc-enduser-iam.yaml

that works because the ssm session is tagged with the synapse user and resolves to the aws userid, which is the same as a tag on the ec2

ahayden: There's a conditional statement below that limits the allowing start-session to 'callers' match a tag on the instance:

PR #223, opened 2021-06-28
Fix shared templates references

zaro0508:
wget when used without the -O option will not download
a file if it already exists. Since we have multiple
sceptre templates referencing different versions of a CFN
template it can download the 1st version then not download the
2nd version.

For example:

assume the following:

config/prod/a.yaml references v1.0 of x.yaml CFN template
config/prod/b.yaml references v2.0 of x.yaml CFN template
workflow:

sceptre runs config/prod/a.yaml and downloads v1.0 of x.yaml
sceptre runs config/prod/b.yaml but does not download v2.0 of x.yaml
becuase v1.0 of x.yaml already exists. sceptre will deploy with the old v1.0 of x.yaml
To force wget to download and overwrite the file we need to use
the -O option.

### DIFF
--- a/.github/workflows/aws-deploy.yaml
+++ b/.github/workflows/aws-deploy.yaml
@@ -81,6 +81,7 @@ jobs:
       - name: Deploy with sceptre
         run: |
           cd sceptre/organizations
+          mkdir -p templates/remote
           sceptre launch prod --yes
   sceptre-transit:
     if: github.event_name == 'push'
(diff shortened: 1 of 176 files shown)


zaro0508:
would you mind taking a look at this one @BrunoGrandePhD ?

BrunoGrandePhD (approved):
I skimmed through the changes and everything looks good. I did a quick search for any leftover usage of `-N -P`, but nothing came up. 

🚀

PR #226, opened 2021-07-03
IT-1294: remove-ahayden IAM user

xschildw:
This PR removes Aaron's IAM user from synapseprod and synapsedw accounts.


ahayden (approved):
Looks good

PR #227, opened 2021-07-03
[SC-304] deploy cfn-cr-sc-product-provider

zaro0508:
Deploy the  cfn-cr-sc-product-provider[1] to allow us to
automate updating of product versions.  The main use case
for this custom resource is so that we can automate the
deprecation of products.

[1] https://github.com/Sage-Bionetworks-IT/cfn-cr-sc-product-provider


brucehoff (commented):
Nvmd.  I looked at PR 228 which answered my question.

brucehoff (approved):

PR #230, opened 2021-07-12
[SC-304] force update version on every deploy

zaro0508:
The lambda for UpdateProductVersions will not run if there aren't
any changes for the resource.  That's how cloudformation
idempotency works.

The problem is that when a new version of a product is added the
UpdateProductVersions lambda does not execute which means that
the previous version of the product will not get deprecated.
In this case we need to force a change to the resource so that
CFN will execute the lambda on every deployment.  That will make
sure that all previous versions of a product is deprecated.

### DIFF
--- a/sceptre/scipool/templates/sc-product-ec2-linux-jumpcloud-workflows.j2
+++ b/sceptre/scipool/templates/sc-product-ec2-linux-jumpcloud-workflows.j2
@@ -93,6 +93,7 @@ Resources:
   UpdateProductVersions:
     Type: Custom::ProductProvider
     Properties:
+      ReDeploy: '{{ range(1, 10000) | random }}'  # force CFN to redeploy this resource
       ServiceToken: !ImportValue
         'Fn::Sub': '${AWS::Region}-cfn-cr-sc-product-provider-UpdateProvisioningArtifactFunctionArn'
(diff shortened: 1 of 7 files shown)


sceptre/scipool/templates/sc-product-ec2-linux-jumpcloud-workflows.j2
```diff
@@ -93,6 +93,7 @@ Resources:
   UpdateProductVersions:
     Type: Custom::ProductProvider
     Properties:
+      ReDeploy: '{{ range(1, 10000) | random }}'  # force CFN to redeploy this resource
```
brucehoff, line 96: Do you think 10000 is big enough?  Maybe use 2^31?

brucehoff (commented):
Do you think 10000 is big enough? Maybe use 2^31?

sceptre/scipool/templates/sc-product-ec2-linux-jumpcloud-workflows.j2
zaro0508: yeah i think but since you're asking i'll bumped it to 1 million.

brucehoff:
When I click "Add Your Review" the screen gets mangled.  There is some sort of Github bug to sort out.  Sorry for the delay.

brucehoff (approved):

PR #232, opened 2021-07-13
SC-275: Remove expensive and outdated instance types

brucehoff:
I removed C4 and R4 instance types since those lines are outdated.
I verified that all other instance types used are considered 'current generation' by AWS.
For each retained instance type I removed the very large, expensive variations, using $1/hour as the threshold.
The total number of instance types is reduced from 145 to 77.
Of the instances removed, only two are in use in scipool-prod: (1) an r4.4xlarge (this is considered outdated) and (2) a c5d.24xlarge.


zaro0508 (changes requested):
I don't see any evidence that we are referencing this file from the service catalog.  I believe the correct place to make your change is in each ec2 file in the https://github.com/Sage-Bionetworks/service-catalog-library repo.

for example: https://github.com/Sage-Bionetworks/service-catalog-library/blob/master/ec2/sc-ec2-linux-docker.yaml#L35

PR #233, opened 2021-07-14
Leverage new 5-TB limit in Docker and Workflows EC2 SC products

BrunoGrandePhD:
Do we add the new versions to the `develop` stack group first and if that deploys fine, perform the same change for `prod` and `strides` in a separate PR?


zaro0508 (approved):

zaro0508:
@BrunoGrandePhD yes, that's the typical workflow but with this simple change you can probably just update all environments  with a single PR.

zaro0508 (approved):

PR #246, opened 2021-07-27
[SC-307] Create account for BMGF

zaro0508:
Create an AWS account to run a Service Catalog for the Bill and
Melinda gates foundation.

PR #254, opened 2021-07-29
[SC-307] setup AWS SSO for BMGF KI account

zaro0508:
Enable SSO and create the Admin and community manager roles for the
service catalog users.

depends on #246


tthyer (approved):
LGTM

PR #269, opened 2021-08-18
[SC-307] Add bmgfki vpc to transit gateway

zaro0508:
Setup service catalog  VPC in bmgfki account with the transit gateway


brucehoff (approved):

zaro0508:
this PR is no longer valid, we are using org-formation to deploy spokes now.

PR #271, opened 2021-08-24
[SC-359] Fix S3 RestrictSameRegionDownload

zaro0508:
Something changed in AWS that now requires the SCS3LaunchRole to have
additional permissions to setup RestrictSameRegionDownload.  This
change adds the additional permission.


BrunoGrandePhD (approved):
Thanks for investigating this issue! 🚀

PR #278, opened 2021-08-26
Create `TowerUser` SSO role for Viewer plus AssumeRole

BrunoGrandePhD:
Given that the AWS `sandbox` account is being deprecated, we want to provide a Viewer role to Nextflow users so they can access a secret in Secrets Manager (for configuring Nextflow Tower) and an S3 bucket (for storing workflow output files). This role also provides the ability to assume other roles (`sts:AssumeRole`) such that it can assume the roles defined [here](https://github.com/Sage-Bionetworks-Workflows/aws-workflows-nextflow-infra/pull/50/files#diff-3c3e4cb346cd6a36455d477015c96c5544148affef185b7218742ce0a08ad593R50-R80), which grant access to specific resources.


BrunoGrandePhD:
We're sticking with Viewer for now.

PR #291, opened 2021-09-16
Add k8 develop group

thomasyu888:
I'm not actually sure what policies are required for a specific user to use kubernetes... I did see this: https://aws.amazon.com/premiumsupport/knowledge-center/amazon-eks-cluster-access/, but I'm also not really certain...


org-formation/700-aws-sso/_tasks.yaml
```diff
@@ -664,6 +664,48 @@ SsoImageCentralAmiLibrarian:
...
+                  ],
+                  "Resource": "*"
+              },
+              {
+                  "Effect": "Allow",
+                  "Action": "iam:PassRole",
+                  "Resource": "*",
+                  "Condition": {
+                      "StringEquals": {
+                          "iam:PassedToService": "arn:aws:iam::563295687221:role/accounts-EksServiceRole-MU4BSZOHRHCG"
```
thomasyu888, line 701: This is a service role that got created automatically when i manually created the EKS cluster.

org-formation/700-aws-sso/_tasks.yaml
```diff
@@ -664,6 +664,48 @@ SsoImageCentralAmiLibrarian:
...
+  DefaultOrganizationBindingRegion: !Ref primaryRegion
+  DefaultOrganizationBinding:
+    IncludeMasterAccount: true
+  OrganizationBindings:
+    TargetBinding:
+      Account:
+        - !Ref SandboxAccount
+  Parameters:
+    instanceArn: !Ref instanceArn
+    principalId: !Ref TBD
```
zaro0508, line 681: please follow the existing pattern, this should reference a parameter and the `TBD` should be on the `Default:` of the parameter.  something like this.. `principalId: !Ref kubernetesDeveloperGroup`

org-formation/700-aws-sso/_tasks.yaml
```diff
@@ -664,6 +664,48 @@ SsoImageCentralAmiLibrarian:
...
+    permissionSetName: 'Kubernetes-Developer'
+    sessionDuration: 'PT12H'
+    inlinePolicy: >-
```
zaro0508, line 684: I don't think this policy, by itself, is enough for users to use or manage EKS.  It looks like there are a lot of different features of EKS and policy you apply depends on what EKS features you setup.  I would suggest you combine the `arn:aws:iam::aws:policy/PowerUserAccess` policy with a few of the EKS inline or managed policies.  Again it depends on which EKS options you setup.  I would take a look here for an explanation of inline policies that are required, https://docs.aws.amazon.com/eks/latest/userguide/security_iam_id-based-policy-examples.html   and take a look here for the managed policies, https://docs.aws.amazon.com/eks/latest/userguide/security-iam-awsmanpol.html

You could probably test those out manually in sandbox before deciding which set to go with?

zaro0508 (changes requested):

org-formation/700-aws-sso/_tasks.yaml
thomasyu888: Thanks Khai.  I have actually looked at those and will have to test some of those policies out.  I think the most confusing part is that all of these policies seem to allow people to create, list and manage the cluster, but theres not really something that is obvious to me that allows for people to just "use the cluster".   

For instance, I don't know whether a user would need permission to "ec2:createInstance" since sometimes when you deploy a kubernetes deployment, it may spin up a new instance.

PR #306, opened 2021-10-28
refactor: replace hooks with template handler

zaro0508:
Sceptre has template handlers[1] now so we can abandon the use of
hooks in favor of the http template handler to get templates
from our shared template repository. This is only to update
scipool accounts.

[1] https://sceptre.cloudreach.com/dev/docs/template_handlers.html#template-handlers

### DIFF
--- a/sceptre/scipool/config/bmgfki/AccountAlertTopics.yaml
+++ b/sceptre/scipool/config/bmgfki/AccountAlertTopics.yaml
@@ -1,4 +1,6 @@
-template_path: "remote/AccountAlertTopics.yaml"
+template:
+  type: http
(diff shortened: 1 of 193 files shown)


brucehoff (approved):

zaro0508:
> Look great!! I noticed that they do have an S3 type where you can set the bucket/key and it would use the same credentials that are used to run sceptre: do you think that it would be worth switching? I'm not sure if there is added value as all of our templates are accessible through a web server? (are they public?)

no, not worth switching because yes, all our templates are indeed public.   We would need to use the S3 type if we wanted to make our templates private.

PR #310, opened 2021-11-02
[IT-1478] deploy cost anomaly detector services

zaro0508:
Setup AWS cost anomaly detection on every account to get notifications
of unforeseen cost changes.  Notifications are sent to account owner email
lists.  This will hopefully help prevent situations where accounts are racking
up big costs without anybody knowing.

depends on https://github.com/Sage-Bionetworks/aws-infra/pull/325

### DIFF
--- a/org-formation/050-costs/_tasks.yaml
+++ b/org-formation/050-costs/_tasks.yaml
@@ -1,0 +1,18 @@
+Parameters:
+  <<: !Include '../_parameters.yaml'
+
+AnomalyDetectorServices:
+  Type: update-stacks
+  Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.3.3/templates/Cost/anomaly-detector-services.yaml
+  StackName: !Sub '${resourcePrefix}-anomaly-detector-services'
+  DefaultOrganizationBinding:
+    IncludeMasterAccount: true
+    Account: '*'
+    Region: !Ref primaryRegion
+  Parameters:
+    Subscribers: [
+      {
+        "Type": "EMAIL",
+        "Address": !GetAtt CurrentAccount.RootEmail
+      }
+    ]
--- a/org-formation/050-costs/_tasks.yaml
+++ b/org-formation/050-costs/_tasks.yaml
@@ -1,18 +1,13 @@
 Parameters:
   <<: !Include '../_parameters.yaml'
 
-AnomalyDetectorServices:
+AnomalyDetectorService:
   Type: update-stacks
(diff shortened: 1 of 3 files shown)


brucehoff (approved):

org-formation/050-costs/_tasks.yaml
```diff
@@ -0,0 +1,13 @@
...
+
+AnomalyDetectorService:
+  Type: update-stacks
+  Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.3.3/templates/Cost/anomaly-detector-service.yaml
+  StackName: !Sub '${resourcePrefix}-anomaly-detector-service'
+  DefaultOrganizationBinding:
+    IncludeMasterAccount: true
+    Account: '*'
+    Region: !Ref primaryRegion
+  Parameters:
```
BrunoGrandePhD, line 12: Do you know if the daily reports (the [current default value](https://github.com/Sage-Bionetworks/aws-infra/blob/master/templates/Cost/anomaly-detector-service.yaml#L9-L16) for the `Frequency` parameter) are sent regardless if there was a triggering event? I just don't want the root emails being spammed with empty reports.

BrunoGrandePhD, line 12: Is there a way of setting the `Threshold` parameter on a per-account basis? Like what's currently being done with the budget alerts.

zaro0508: I  believe notifications will only trigger when the `threshold` is exceeded.. 
 https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ce-anomalysubscription.html#cfn-ce-anomalysubscription-threshold

zaro0508: Do you mean per AWS account or per IAM account?  My understanding is this will monitor per AWS service in each AWS account.

BrunoGrandePhD: Sorry, I was referring to each account deployed in organizations-infra. Or is this monitor set once on the parent account (_i.e._ the account at the head of the AWS Organization) and cannot be tailored for member accounts (_i.e._ individual accounts in the AWS Organization, like nextflow-prod and nextflow-dev)?

BrunoGrandePhD: Got it!

zaro0508 (commented):
darn you GH!

org-formation/050-costs/_tasks.yaml
BrunoGrandePhD: That's good to know. I'll have to see how sensitive the Cost Anomaly Detection is before adjusting the threshold for the Nextflow AWS accounts.

BrunoGrandePhD (approved):

PR #316, opened 2021-11-12
Setup admin access to scicomp

zaro0508:
We setup AWS SSO to allow JC users to access the AWS Scicomp account with
admin privileges.  This is to allow the IT team to manage Scicomp.


zaro0508:
This is not needed. admin for Scicomp is already setup

PR #317, opened 2021-11-14
[SC-306] Scheduled jobs VPC

zaro0508:
Create a VPC for the SC scheduled jobs product.  The idea is to allow
docker images to run on instances in an isolated network.

We use the Mini VPC which will allow batch to run up to 28 jobs in parallel.


brucehoff (commented):
You say

> We use the Mini VPC which will allow batch to run up to 28 jobs in parallel.

How is the VPC shared?  If each provisioned product has its own VPC then allowing up to 28 jobs seems reasonable.  If the entire Service Catalog shares one VPC then it seems too restrictive.

zaro0508:
> How is the VPC shared?

All script runner instances share a VPC in the SC.  

>  If each provisioned product has its own VPC then allowing up to 28 jobs seems reasonable. If the entire Service Catalog shares one VPC then it seems too restrictive.

It's currently the latter approach.  What makes you think that it's too restrictive?  If there are two many cron jobs to run at the same time batch will put the jobs in a queue and run the cron job as the queue winds down.

brucehoff:
> > How is the VPC shared?
> 
> All script runner instances share a VPC in the SC.
> 
> > If each provisioned product has its own VPC then allowing up to 28 jobs seems reasonable. If the entire Service Catalog shares one VPC then it seems too restrictive.
> 
> It's currently the latter approach. What makes you think that it's too restrictive? If there are two many cron jobs to run at the same time batch will put the jobs in a queue and run the cron job as the queue winds down.

How does the script runner product use Batch?  If 30 users run 30 SC Batch products, is there some central/shared queue that tells some of the script runners to wait?

brucehoff (approved):

PR #321, opened 2021-11-22
Update all accounts with S3 read access

zaro0508:
This will include changes in PR https://github.com/Sage-Bionetworks/aws-infra/pull/326
which will allow S3 read access to all regions.  This will be applied
to all AWS accounts in our ORG.

depends on https://github.com/Sage-Bionetworks/aws-infra/pull/326


brucehoff (approved):
Bruno should also review before merging.

BrunoGrandePhD (approved):
Thanks!

brucehoff (approved):

xschildw (approved):

PR #324, opened 2021-12-03
[IT-1549] Fix AWS config logging to logcentral account

zaro0508:
AWS Config from strides account wasn't logging to logcentral
account due to missing permissions.  We also need to give
strides account permissions to write to the central s3 config
bucket in logcentral account.

We are following the pattern established by org-formation which
does work as expected. org-formation patter is defined in
[ConfigurationRecorderRole](https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/master/templates/Config/config.yaml)

### DIFF
--- a/sceptre/strides/templates/aws-config.yaml
+++ b/sceptre/strides/templates/aws-config.yaml
@@ -73,10 +73,14 @@ Resources:
           Statement:
           - Effect: Allow
             Action: s3:GetBucketAcl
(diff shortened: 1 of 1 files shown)


brucehoff (approved):

ahayden (approved):
Should work as long as ${LogCentralBucket} is the Cloudtrail target

PR #327, opened 2021-12-09
Deploy CI service account to sandbox

BrunoGrandePhD:
I would like to create a CI service account in sandbox to test the deployment of workflows into an AWS account that isn't one of our two Nextflow accounts. These credentials will be added to our [nextflow-infra repository](https://github.com/Sage-Bionetworks-Workflows/nextflow-infra) as secrets and used for Sceptre deployment. 

Alternatively, if the AMP-AD STRIDES account can be bootstrapped ([IT-1536](https://sagebionetworks.jira.com/browse/IT-1536)), I can test the deployment there instead.


zaro0508:
working on setting up account in PR #329

PR #329, opened 2021-12-10
[IT-1536] Setup org-sagebase-strides-ampad-workflows AWS Account

zaro0508:
We need to setup some snowflake infra for the
org-sagebase-strides-ampad-workflows account because it's not
under our org account.  This sets up master member configuration
for guard duty, AWS config, and cloudtrail.


zaro0508:
replacing this with PR #330

PR #330, opened 2021-12-10
[IT-1536] Setup org-sagebase-strides-ampad-workflows AWS Account

zaro0508:
We need to setup some snowflake infra for the
org-sagebase-strides-ampad-workflows account because it's not
under our org account.  This sets up master member configuration
for guard duty, AWS config, and cloudtrail.


BrunoGrandePhD (approved):
I can't comment on the stacks being deployed (like if they're configured properly), but I acknowledge this PR. 🚀 

Given the way this account is being bootstrapped, will I be able to update #327 to refer to this account instead of sandbox? Or will that have to be done via Sceptre?

PR #331, opened 2021-12-10
[IT-1536] Setup services for org-sagebase-strides-ampad-workflows

zaro0508:
Setup member portions of guardduty, cloudtrail, and aws config for
org-sagebase-strides-ampad-workflows account.  Also setup anomaly
detector.

depends on PR #330


BrunoGrandePhD (approved):
I can't comment on the stacks being deployed (like if one is missing), but I acknowledge this PR. 🚀

PR #337, opened 2021-12-11
[IT-1536] setup new account

zaro0508:
This is the snowflake infra that's needed when we create accounts that
are not in our Sage Organizations. 

* setup CI/CD for org-sagebase-strides-ampad-workflows account
* setup base infrastructure
* setup the account as guardduty, cloudtrail and aws config member
  accounts sending logs to the master account.
* setup jumpcloud integration
* setup slack integration


BrunoGrandePhD (approved):
I can't comment on the stacks being deployed (like if one is missing), but I acknowledge this PR. 🚀

PR #344, opened 2021-12-16
Remove `templates/` from S3 URL

BrunoGrandePhD:
I copy-pasted the GitHub URL and changed it to the S3 bucket, but `templates/` isn't part of the key in S3.


zaro0508 (approved):
i didn't mirror the GH dir structure when i deployed aws-infra templates to S3 bucket which is coming to bite us.  unfortunately it is something that would be difficult to make right now.

BrunoGrandePhD:
No worries!

PR #357, opened 2022-01-22
[IT-1549] Fix config and cloudtrail log forwarding

zaro0508:
AWS Config and cloudtrail log forwarding to logcentral is not working for
all our accounts.  It currently only forwards logs from the strides accounts.

The problem is that we are using two separate methods (org-formation & sceptre)
to configure the S3 bucket policies for these services.  The sceptre one
is trampling over the org-formation one thus breaking log forwarding for
all of our accounts in our AWS organizations.

The cause of the problem is that we execute the org-formation templates before
then sceptre templates.  The org-formation template correctly configures the S3
cloudtrail and config bucket policies for the AWS organization.  After that we
execute the sceptre template for strides which overwrites the S3 bucket policy
allowing only the strides account access to the config and cloudtrail buckets.

This will disable the execution of the sceptre template for strides
which will restore log forwarding from our organizations accounts.


ahayden:
Seems like a necessary temp fix

ahayden (approved):

zaro0508:
Further investigation indicates that this isn't the root cause of the problem.

PR #377, opened 2022-02-11
PLFM-7169: Remove  IAM users from Synapse Prod account

xschildw:
Developers should have required access through AWS SSO / JC. Per comment in template, **do not merge** until we have cleaned up the users (MFA etc.). Leaving groups etc in place in case we need to revert.


zaro0508 (changes requested):
why not delete everything?  I don't think we need the `AWSIAMSynapseEmergencyUser` anymore which means that all the supporting roles and policies in this file are not used anymore.  I suggest just remove it all.

xschildw:
My intention was to keep it around in case we have a problem with JC/SSO and need to get in. I guess we could revert to root account in that case?

PR #378, opened 2022-02-14
Add Nick Grosenbacher IAM user

nickgros:
Getting access to synapsedw org so I can experiment with Redash, our visualization tool for the data warehouse


xschildw:
@nickgros , if you just need Redash access you need to be put in the correct group on the VPN (same for database access).

zaro0508:
closing this because we are going to remove all users in PR https://github.com/Sage-Bionetworks-IT/organizations-infra/pull/377

PR #322, opened 2021-11-22
Move SageIT to prod OU after IT-447 resolved

ahayden:
The follow up to IT-447 is to move the SageIT account into ITProd

PR #380, opened 2022-02-17
[IT-1621] Enable aws guard duty on all accounts

zaro0508:
Run the AWS guardduty service on all accounts and have them
report back to the guardduty hub (logcentral).


ahayden (approved):
Looks good. This adds the entire ItProdOU and org account to the binding and removes the atomic accounts.

brucehoff (approved):

zaro0508:
> Should #322 be resolved first?

good observation @DiepThach however we can do either one first.

PR #382, opened 2022-02-28
IT-2309: Remove IAM accounts

xschildw:
This PR removes IAM accounts from the SynapseDW account. Wait for IT-1393 to be resolved to merge.


zaro0508 (approved):
remember you'll need to manually delete the `accounts` CFN stack after merging.

PR #384, opened 2022-03-02
Fix Guardduty

zaro0508:
Guardduty got into a broke state due to PR #380.

The problem was that we aded ItProdOU to the list of accounts
That OU contained securitycentral which should not be added as
a guardduty member because it's already the guardduty master.

This change uses `*` to adds all accounts and excludes securitycentral
which will essentially enable guardduty for all accounts going
forward.

Note: in order to get guardduty out of the bad state it had to be
completely reset.

### DIFF
--- a/org-formation/070-guard-duty/_tasks.yaml
+++ b/org-formation/070-guard-duty/_tasks.yaml
@@ -12,7 +12,7 @@ Parameters:
 
 GuardDuty:
   Type: update-stacks
-  Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.2.4/templates/GuardDuty/guard-duty.yaml
+  Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.3.8/templates/GuardDuty/guard-duty.yaml
   StackName: !Sub '${resourcePrefix}-${appName}'
   StackDescription: GuardDuty - Base
   DefaultOrganizationBindingRegion: !Ref primaryRegion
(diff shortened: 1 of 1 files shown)


org-formation/070-guard-duty/_tasks.yaml
```diff
@@ -12,7 +12,7 @@ Parameters:
 
 GuardDuty:
   Type: update-stacks
-  Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.2.4/templates/GuardDuty/guard-duty.yaml
+  Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.3.8/templates/GuardDuty/guard-duty.yaml
```
zaro0508, line 15: these updates are essentially no-op changes.  it's just nice to update to a newer version occasionally.

org-formation/070-guard-duty/_tasks.yaml
```diff
@@ -22,32 +22,12 @@ GuardDuty:
...
-      OrganizationalUnit:
-        - !Ref BridgeOU
-        - !Ref ServiceCatalogOU
-        - !Ref ScienceDevOU
-        - !Ref ScienceProdOU
-        - !Ref SynapseOU
+      Account: '*'
       IncludeMasterAccount: true
+      ExcludeAccount:
+        - !Ref accountId
```
ahayden, line 28: This is an explicit exclude of the securitycentral account.  Can you confirm the `IncludeMasterAccount: true` binding will enforce GD on securitycentral and that the organizations master account is included in the `*` member binding?

ahayden (approved):

org-formation/070-guard-duty/_tasks.yaml
zaro0508: to confirm accounts that have guardduty enabled just goto securitycentral -> guardduty -> accounts

ahayden: confirmed, zaro manually ran this config first, and the state is as expected except for `strides-ampad-workflows`

PR #388, opened 2022-03-09
IT-1710: GH deployer user and role

xschildw:
This PR creates a user and a role to be used by GH workflow for portal deployment (sageit account)


sceptre/sageit/templates/gh_deployer.yaml
```diff
@@ -0,0 +1,62 @@
+Description: Setup IAM service user and policies for Github deployment of S3 website
```
zaro0508, line 1: A better place to create service accounts is at https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/org-formation/600-access/_tasks.yaml#L4

zaro0508 (changes requested):

xschildw:
This is superseded by the GH OIDC in infra.

PR #389, opened 2022-03-09
IT-1694: add towerViewer permissions to template

xschildw:
DO NOT MERGE: Need to add .xml in idp/jumpcloud.


sceptre/strides-ampad-workflows/templates/jumpcloud-idp.yaml
```diff
@@ -60,6 +63,94 @@ Resources:
...
+          - Effect: Allow
+            Principal:
+              Federated: !Ref TowerViewerSamlProvider
+            Action: sts:AssumeRoleWithSAML
+            Condition:
+              StringEquals:
+                "SAML:aud": "https://signin.aws.amazon.com/saml"
+      Path: /
+      Policies:
+        - PolicyName: TowerViewerPolicy
```
BrunoGrandePhD, line 91: This policy was streamlined since this draft PR was opened in early March. 

https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/org-formation/700-aws-sso/_tasks.yaml#L609-L621

BrunoGrandePhD (approved):
🚀

BrunoGrandePhD:
@zaro0508: Any chance you can review this today?

sceptre/strides-ampad-workflows/templates/jumpcloud-idp.yaml
```diff
@@ -60,6 +63,39 @@ Resources:
             Condition:
               StringEquals:
                 "SAML:aud": "https://signin.aws.amazon.com/saml"
+  ViewerSamlProvider:
+    Type: Custom::SAMLProvider
+    Properties:
+      ServiceToken: !ImportValue
+        'Fn::Sub': '${AWS::Region}-cfn-cr-saml-provider-FunctionArn'
+      Name: "strides-ampad-workflows-viewer"
```
zaro0508, line 71: The resource name and this name implies viewer access to the entire aws account however that doesn't match with the actual permission that's given for this SAML provider.  Would it make sense to provide a name that is more appropriate to the access that's provided?

zaro0508 (changes requested):

sceptre/strides-ampad-workflows/templates/jumpcloud-idp.yaml
xschildw: I thought that but the permissions were Batch:ListJobs on * (which I saw as not limited to Tower) so I named it more generic to match what we do elsewhere. I'll create matching assets in JC and rename, we can keep the existing JC stuff in case we want to create a generic Viewer group.

BrunoGrandePhD: This role will be granted additional access on resource policies (like S3 and KMS). It also used to have more permissions prior to this [pull request](https://github.com/Sage-Bionetworks-IT/organizations-infra/pull/426/files), but we removed them with the plan of adding them back incrementally once we establish their usefulness. 

That said, if you want to change the name, I'm happy to switch to `TowerUser`, but you'll need to update the name [here](https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/org-formation/700-aws-sso/_tasks.yaml#L606) too.

BrunoGrandePhD: You would need to change the group names in JumpCloud too

BrunoGrandePhD (approved):

PR #394, opened 2022-03-10
Authorization rule for discussion

ahayden:


org-formation/720-client-vpn/vpn.njk
```diff
@@ -120,6 +120,18 @@ Resources:
 {%   endfor %}
 {% endfor %}
 
+  EndpointAuthorizationtoPublic:
+    Type:AWS::EC2::ClientVpnAuthorizationRule
+    DependsOn:
```
zaro0508, line 125: duplicate `DependsOn`

org-formation/720-client-vpn/vpn.njk
```diff
@@ -133,7 +145,7 @@ Resources:
     Properties:
       Description: {{ spoke }}-Subnet{{ subnet_id | last }}
       ClientVpnEndpointId: !Ref Endpoint
-      DestinationCidrBlock: {{ spoke_data.CIDR }}
+      DestinationCidrBlock: 0.0.0.0/0
```
zaro0508, line 148: hmm, so we are changing this from being able to access specific spokes to being able to access anything?  seems too permissive to me.

ahayden, line 148: Probably shouldn't do this, actually... but instead leave the individual cidr ranges to spokes in the templates. If so, there needs to an explicit route to `0.0.0.0/0` to catch any traffic destined outside of the private ranges

PR #395, opened 2022-03-10
[IT-1234] Enable VPN full tunnel

zaro0508:
To setup VPN to access internet traffic we need to provide a
route between the VPN and the internet as described in the
AWS documentation[1].  This will allow a full VPN tunnel to
access the internet.

[1] https://docs.aws.amazon.com/vpn/latest/clientvpn-admin/scenario-internet.html

### DIFF
--- a/org-formation/720-client-vpn/vpn.njk
+++ b/org-formation/720-client-vpn/vpn.njk
@@ -101,20 +101,20 @@ Resources:
   EndpointAuthorizationPublic:
     Type: AWS::EC2::ClientVpnAuthorizationRule
     Properties:
-      Description: "authorization between VPN and public internet"
-      AccessGroupId: {{ access_group }}
+      Description: "public"
+      AuthorizeAllGroups: true
       ClientVpnEndpointId: !Ref Endpoint
       TargetNetworkCidr: "0.0.0.0/0"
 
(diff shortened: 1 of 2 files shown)


ahayden (approved):

org-formation/720-client-vpn/vpn.njk
```diff
@@ -102,6 +102,26 @@ Resources:
       SubnetId: {{ subnet_id }}
 {% endfor %}
 
+{# Authorization and route between VPN and public internet #}
+  EndpointAuthorizationPublic:
+    Type: AWS::EC2::ClientVpnAuthorizationRule
+    Properties:
+      Description: "authorization between VPN and public internet"
+      AccessGroupId: {{ access_group }}
```
ahayden, line 110: Will `{{ access_group }}` resolve outside of the loop on ln 127?

zaro0508: ah yes, nice catch.  This will not resolve here.  We are gonna need a loop to give all access groups access to 0.0.0.0/0

zaro0508: fixed

zaro0508:
@ahayden tested and says it works as expected

PR #396, opened 2022-03-11
Setup VP for full tunnel

zaro0508:
The aws documentation[1] suggests that we need to associate the AWS
VPN to public subnets instead of private subnets to allow
internet traffic.  This re-enables the TGW and VPN setup and update
the VPN to use public subnets.

[1] https://docs.aws.amazon.com/vpn/latest/clientvpn-admin/scenario-internet.html

### DIFF
--- a/org-formation/710-tgw/vpc.yaml
+++ b/org-formation/710-tgw/vpc.yaml
@@ -15,8 +15,12 @@ Mappings:
       CIDR: "2.0/24"
     SubnetC:  # private
       CIDR: "3.0/24"
-#    SubnetD:  # public
-#      CIDR: "100.0/24"
+    SubnetD:  # public
+      CIDR: "100.0/24"
+    SubnetE:  # public
+      CIDR: "101.0/24"
+    SubnetF:  # public
+      CIDR: "102.0/24"
 Resources:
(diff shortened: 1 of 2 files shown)


org-formation/710-tgw/vpc.yaml
```diff
@@ -73,16 +77,36 @@ Resources:
...
-#      CidrBlock: !Join
-#        - '.'
-#        - - !Ref VpcSubnetPrefix
-#          - !FindInMap [SubnetConfig, SubnetD, CIDR]
-#      VpcId: !Ref VPC
-#      MapPublicIpOnLaunch: true
+  SubnetD:
+    Type: "AWS::EC2::Subnet"
+    Properties:
+      AvailabilityZone: !Select [0, !GetAZs '']
```
ahayden, line 83: SubnetA is already select [0] from !GetAZs... this will not conflict?

ahayden (approved):

org-formation/710-tgw/vpc.yaml
zaro0508: i don't think so because the VPN is no longer associated with private subnets (A,B, and C).

PR #403, opened 2022-03-16
IT-1710: Delete gha user/role/policy

xschildw:
The resources created by this stack have drifted. The user has been deleted and the policy has been modified, better to delete everything and start over.


zaro0508 (approved):
it looks like `gha-svc-account` is still there, please remember to manually delete it.

PR #417, opened 2022-03-23
[IT-1598] Setup AWS inspector

zaro0508:
Enable AWS inspector service in all accounts and delegate
management of the service to org-sagebase-securitycentral
account.

### DIFF
--- a/org-formation/075-inspector/inspector.yaml
+++ b/org-formation/075-inspector/inspector.yaml
@@ -1,0 +1,76 @@
+AWSTemplateFormatVersion: '2010-09-09-OC'
+OrganizationBindings:
+  InspectorBinding:
+Mappings:
+  RulesPackagesAmazonInspectorArns: # ARNs derived from https://docs.aws.amazon.com/inspector/latest/userguide/inspector_rules-arns.html
+    us-east-1:
+      CommonVulnerabilitiesExposures: arn:aws:inspector:us-east-1:316112463485:rulespackage/0-gEjTy7T7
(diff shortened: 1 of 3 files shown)


org-formation/075-inspector/inspector.yaml
```diff
@@ -0,0 +1,91 @@
...
+      SecurityBestPractices: arn:aws:inspector:us-west-2:758058086616:rulespackage/0-JJOtZiqQ
+Resources:
+  InspectorServiceLinkedRole:
+    Type: AWS::IAM::ServiceLinkedRole
+    OrganizationBinding: !Ref InspectorBinding
+    Properties:
+      AWSServiceName: inspector.amazonaws.com
+      Description: Service-Linked-Role for Amazon Inspector
+  AssessmentTarget:
+    Type: AWS::Inspector::AssessmentTarget
```
ahayden, line 49: I think we would want to use InspectorV2, as it runs through the SSM agent and can examine ECR containers, doesn't run on a cron schedule, and doesn't accept targets or rule packages likes this (which is V1): https://docs.aws.amazon.com/inspector/latest/user/getting_started_tutorial.html

ahayden (changes requested):

zaro0508:
nice catch @ahayden.  There doesn't seem to be full Inspector V2 support in cloudformation yet.  It looks like the only thing that is supported is the ability to [create inspector filters](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-inspectorv2-filter.html).  I have manually enabled inspector V2 so I will close this PR.  We can create PRs to add CFN templates defining custom filters if you wish.

PR #418, opened 2022-03-24
PLFM-1695: Add service account for GHA portals deployment

xschildw:
This is the replacement for https://github.com/Sage-Bionetworks-IT/organizations-infra/pull/388.
The intent is to create a role and user to use in Github workflow to deploy files to S3 bucket (and
invalidate a Cloudfront distribution).

### DIFF
--- a/org-formation/600-access/_tasks.yaml
+++ b/org-formation/600-access/_tasks.yaml
@@ -115,6 +115,51 @@ CnbCIServiceAccount:
       - !Ref CnbAccount
     Region: !Ref primaryRegion
 
+# Service account for GHA portal deployer in SageIT account
+PortalDeployGhaServiceAccount:
+  Type: update-stacks
+  Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.3.8/templates/IAM/service-account.yaml
+  StackName: portal-deployer-gha-service-account
+  Parameters:
+    PolicyDocument: |
+      {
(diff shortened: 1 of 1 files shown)


org-formation/600-access/_tasks.yaml
```diff
@@ -115,6 +115,51 @@ CnbCIServiceAccount:
...
+              "s3:PutBucketWebsite",
+              "s3:PutObject",
+              "s3:PutObjectAcl",
+              "s3:GetObject",
+              "s3:ListBucket",
+              "s3:DeleteObject",
+              "s3:GetBucketLocation"
+            ],
+            "Effect": "Allow",
+            "Resource": "*"
```
zaro0508, line 152: Should this be further scoped to only allow access to specific buckets?

zaro0508 (changes requested):

org-formation/600-access/_tasks.yaml
xschildw: Sure, but I'd say 70% of buckets (maybe more if we clean up) and 80% of distributions are in the list...
Should I try to deny on stuff we know is critical (e.g. essentials*)?

zaro0508: It probably depends on the use case.  i took a look at the referenced issue PLFM-1695 and didn't see anything that explains what this role is for.

xschildw:
Superseded by GH OIDC.

PR #423, opened 2022-03-30
Prevent parallel AWS deployments

zaro0508:
We use github actions to deploy cloudformation to AWS.  This deployment
is currently setup to run in parallel when multiple PRs are merged at
the same time.  Parallel infrastructure deployments to AWS do not work therefore
we want to turns off github action parallel deployments to AWS.

This separates the GH action to two workflows.  The deployment
workflow will only run if there is a merge to the master branch and
the validate workflow passes.

### DIFF
--- a/.github/workflows/aws-deploy.yaml
+++ b/.github/workflows/aws-deploy.yaml
@@ -1,37 +1,17 @@
 name: aws-deploy
 
 on:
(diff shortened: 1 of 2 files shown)


xschildw (approved):
Did not know about this...

PR #427, opened 2022-04-01
[IT-1234] Setup VPN access to synapsedw resources

zaro0508:
* Setup routes from synapsedw VPC to TGW hub VPC
* Allow JC users to access synapsedw VPC thru the VPN

### DIFF
--- a/org-formation/720-client-vpn/_tasks.yaml
+++ b/org-formation/720-client-vpn/_tasks.yaml
@@ -97,6 +97,12 @@ Vpn:
         AccessGroups:
           - "aws-admins"
           - "bridge-developers"
+      # org-sagebase-synapsedw
+      synapsedw-vpc:
+        CIDR: "10.12.0.0/16"
+        AccessGroups:
+          - "aws-admins"
+          - "aws-synapse-admins"
 
   Parameters:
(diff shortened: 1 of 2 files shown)


org-formation/720-client-vpn/_tasks.yaml
```diff
@@ -97,6 +97,12 @@ Vpn:
         AccessGroups:
           - "aws-admins"
           - "bridge-developers"
+      # org-sagebase-synapsedw
+      synapsedw-vpc:
+        CIDR: "10.12.0.0/16"
+        AccessGroups:
+          - "aws-admins"
+          - "aws-synapse-admins"
```
xschildw, line 105: We'll need more than admins, users of the datawarehouse snapshots will need access (they run a job to provision an RDS instance and then connect to the instance to run queries).

xschildw (approved):
We may need to create a new group to give access (I think now it's a group on the Sophos appliance).

org-formation/720-client-vpn/_tasks.yaml
zaro0508: you can add a new group in a follow on PR

PR #428, opened 2022-04-04
[IT-1674] Widen scope of cross-account deny

ahayden:
[x] Deny all actions on all resources
[x] Passes pre-commit


BrunoGrandePhD (approved):
As far as I can tell, this will have the intended effect of blocking all requests granted by `AmazonS3FullAccess` except for those relating to the three CHOP AWS accounts.

zaro0508 (approved):

PR #435, opened 2022-04-05
[IT-1757] give our security guy access to security stuff

zaro0508:
Provide our security auditors more access to cloud resources so
they can review more things in our AWS accounts.


xschildw (approved):
All RO so no risk...

ahayden (approved):

PR #436, opened 2022-04-05
Limit ALB logging bucket cross account to single region

ahayden:
[ x ] The org should need only a cross account S3 logger access for the AWS ALB main account from a single region
[ x ] Passes pre-commit


xschildw (approved):
We can't deploy anything outside us-east-1 anyway, correct?

zaro0508 (approved):
seems fine to me, although i don't see why we need to remove it?  I do think that some comments here would help us understand why this is here and what it's used for next time we need to look at it again.

ahayden:
`797873946194` is the wrong account to include in the map, and I believe `027434742980` is not needed. Let's limit the templates only to the account ids we expect to find in them.

PR #438, opened 2022-04-08
[IT-607] Move security hub service

zaro0508:
We are moving the security hub service to the organizations account
and delegating administrating to securitycentral account.

Note: delegation happens after setting this up and it can only be
done manually thru the AWS console or CLI

### DIFF
--- a/org-formation/075-security-hub/_tasks.yaml
+++ b/org-formation/075-security-hub/_tasks.yaml
@@ -7,7 +7,7 @@ Parameters:
 
   accountId:
     Type: String
-    Description: The identifier from the account used to manage GuardDuty
+    Description: The identifier from the account used to manage this service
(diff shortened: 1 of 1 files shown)


org-formation/075-security-hub/_tasks.yaml
```diff
@@ -8,13 +8,13 @@ Parameters:
   accountId:
     Type: String
     Description: The identifier from the account used to manage GuardDuty
```
ahayden, line 10: s/GuardDuty/SecurityHub

ahayden (approved):

org-formation/075-security-hub/_tasks.yaml
zaro0508: fixed

PR #442, opened 2022-04-12
[IT-1234] Setup additional routes for synapse

zaro0508:
AWS requires setting up a route to the TGW VPC in each route table
to allow VPN access to each subnet group (RED, GREEN, BLUE, and
ORANGE)


xschildw (approved):
Let's try...

PR #444, opened 2022-04-12
[IT-1234] Setup TGW and VPN for synapseprod

zaro0508:
* Setup routes from synapseprod VPC to TGW hub VPC
* Allow JC users to access synapseprod VPC resources thru the VPN

Note that the synapse VPC is very unique so we need to handle it with
a custom nunjucks template.

### DIFF
--- a/org-formation/710-tgw/_tasks.yaml
+++ b/org-formation/710-tgw/_tasks.yaml
@@ -188,7 +188,7 @@ TransitGatewaySpokesSynapseDev:
   DependsOn: [ TransitGatewayRoutes ]
   Type: update-stacks
   Template: tgw-spoke-synapse.njk
-  StackName: !Sub '${resourcePrefix}-${appName}-spoke-s'
+  StackName: !Sub '${resourcePrefix}-${appName}-spoke-synapse-dev'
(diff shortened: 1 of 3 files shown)


org-formation/710-tgw/_tasks.yaml
```diff
@@ -123,9 +126,12 @@ Mappings:
     VpcName:
       '055273631518': 'snowflakevpc'   # org-sagebase-scicomp
       '420786776710': 'bridge-aux'   # org-sagebase-bridgedev
-  TgwSpokesS:
+  TgwSpokesSynapseDev:
     VpcName:
       '449435941126': 'synapse-dev-vpc-2'      # org-sagebase-synapsedev
+  TgwSpokesSynapseProd:
+    VpcName:
+      '325565585839': 'synapse-prod-vpc-2'      # org-sagebase-synapseprod
```
xschildw, line 134: I see 'synapse-prod-vpc' (note: need to delete old VPC b4 merging if this is the case). Need to double-check in dev.

xschildw: Scratch, internal ref.

xschildw (approved):

PR #450, opened 2022-04-13
[IT-1742] Updated cron duration to 6 hours

zaro0508:
Update cron duration in shared maintenance template

depends on https://github.com/Sage-Bionetworks/aws-infra/pull/339


zaro0508:
@BrunoGrandePhD i think you have one of these as.. https://github.com/Sage-Bionetworks-Workflows/nextflow-infra/blob/main/config/common/maintenance.yaml#L3

ahayden (approved):
@BrunoGrandePhD may want to release the maintenace template from the v0.4.0 tag

PR #451, opened 2022-04-13
[IT-1234] Setup VPN access

zaro0508:
Allow users to access resources in the VPCs from the AWS VPN

depends on https://github.com/Sage-Bionetworks/aws-infra/pull/338


xschildw (approved):
That's  redeploying the VPCs with additional secGroups for TGW.

PR #452, opened 2022-04-14
Run git-secrets in pre-commit

zaro0508:
Add the git-secrets linter[1] to pre-commit and have the CI
run it on every PR.

[1] https://github.com/awslabs/git-secrets


.pre-commit-config.yaml
```diff
@@ -1,4 +1,12 @@
 repos:
+  - repo: https://github.com/awslabs/git-secrets
+    rev: b9e96b3212fa06aea65964ff0d5cda84ce935f38
+    hooks:
+    - id: git-secrets
+      entry: git-secrets --register-aws --global
+    - id: git-secrets
+      entry: git-secrets --scan
+      args: [--untracked]
```
jesusaurus, line 9: It looks like `--untracked` and `--recursive` are mutually exclusive, maybe we should run the scan twice, once with each option?

[Options for --scan](https://github.com/awslabs/git-secrets#options-for---scan)

zaro0508: I think we need to allow untracked configuration files that may contain secrets.

jesusaurus (approved):

PR #456, opened 2022-04-19
Add SecurityWriter with managed roles into central

ahayden:
[ X ] Adds managed roles for r/w on security central findings. TODO: needs JC group step and GUID
[ X ] Passes pre-commit


zaro0508:
@ahayden please fix conflicts.  also the policy seems very permissive, not sure why we shouldn't just give you the entire enchilada?

PR #463, opened 2022-04-28
PLFM-7096: Add lambda permissions for stack builder

marcomarasca:


xschildw (approved):
We'll tighten all this later....

PR #466, opened 2022-04-30
IT-1234: Add synapse-ops-vpc-v2

xschildw:
Needed to give access to build-system-ops. This VPC was built with infra so put in spokeB.


brucehoff (approved):

zaro0508 (approved):
To give VPN access you will also need to add a section to  [720-client-vpn/_tasks.yaml](https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/org-formation/720-client-vpn/_tasks.yaml#L153L-59) as well.  You can do that in this PR or a follow on PR

PR #468, opened 2022-05-02
PLFM-7262: Add iam perms, need to back down later

xschildw:
Need more perms


xschildw:
This does not get deployed by the infra, it needs to be manually deployed in dev stack account.

marcomarasca (approved):

PR #469, opened 2022-05-02
IT-1850: Fix access to DW

xschildw:
Attempt to access DW infra on VPN

### DIFF
--- a/org-formation/710-tgw/_tasks.yaml
+++ b/org-formation/710-tgw/_tasks.yaml
@@ -111,6 +111,9 @@ TransitGatewayRoutes:
       # org-sagebase-scipoolprod
       internalpoolvpc:
         CIDR: "10.41.0.0/16"
+      # org-sagebase-synapsedw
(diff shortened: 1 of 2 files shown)


org-formation/710-tgw/_tasks.yaml
```diff
@@ -111,6 +111,9 @@ TransitGatewayRoutes:
       # org-sagebase-scipoolprod
       internalpoolvpc:
         CIDR: "10.41.0.0/16"
+      # org-sagebase-synapsedw
+      vpc:
```
zaro0508, line 115: the TgwSpokes needs to match in VPN config file as well, could you please update here as well?  https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/org-formation/720-client-vpn/_tasks.yaml#L141

zaro0508 (changes requested):

zaro0508 (approved):

PR #470, opened 2022-05-04
IT-1234: Add sageit, access to prod and dw

xschildw:
This PR adds the sageIT VPC to TGW/VPN and gives access to infra in synapseProd and synapseDev

### DIFF
--- a/org-formation/710-tgw/_tasks.yaml
+++ b/org-formation/710-tgw/_tasks.yaml
@@ -42,6 +42,7 @@ TransitGateway:
       - !Ref iAtlasProdAccount
       - !Ref ScipoolDevAccount
       - !Ref ScipoolProdAccount
+      - !Ref SageITAccount
 
 UnionStationVpc:
   Type: update-stacks
@@ -114,6 +115,9 @@ TransitGatewayRoutes:
       # org-sagebase-synapsedw
       vpc:
         CIDR: "10.12.0.0/16"
+      # org-sagebase-sageit
+      sagedefaultvpc:
+        CIDR: "10.13.0.0/16"
 
   Parameters:
     VpcSubnetIDs:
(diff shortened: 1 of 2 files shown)


org-formation/710-tgw/_tasks.yaml
```diff
@@ -142,6 +146,7 @@ Mappings:
       '055273631518': 'snowflakevpc'   # org-sagebase-scicomp
       '420786776710': 'bridge-aux'   # org-sagebase-bridgedev
       '325565585839': 'synapse-ops-vpc-v2' # org-sagebase-synapseprod
+      '797640923903': 'sagedefaultvpc' # org-sagebase-sageit
```
zaro0508, line 149: TgwSpokeB is for secondary VPCs in the same account.  Since there are no VPCs in defined TgwSpokeA for sageit yet this should be moved to TgwSpokeA.

zaro0508 (changes requested):

org-formation/720-client-vpn/_tasks.yaml
```diff
@@ -143,6 +143,7 @@ VpnAuthRoutes:
         AccessGroups:
           - "aws-admins"
           - "aws-synapse-admins"
+          - "aws-synapseprod-viewers"
```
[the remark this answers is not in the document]
xschildw: Maybe you're right, they overlap but are not the same...

org-formation/710-tgw/_tasks.yaml
xschildw: I did not make the connection, will do.

xschildw:
Also created group on JC with access the transitvpn application.

org-formation/720-client-vpn/_tasks.yaml
```diff
@@ -143,6 +143,7 @@ VpnAuthRoutes:
         AccessGroups:
           - "aws-admins"
           - "aws-synapse-admins"
+          - "aws-synapsedw-users"
```
zaro0508, line 146: i think we need to make a corresponding sso group, like the synapseAdminGroup here https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/org-formation/700-aws-sso/_tasks.yaml#L84-L86 

an alternative is to set this to `aws-synapseprod-developers` and put kevin in it. if that makes sense to you.

zaro0508 (changes requested):

xschildw:
I somehow pushed my branch to upstream instead of origin. Should close this one.

PR #475, opened 2022-05-05
IT-1234: Fixes

xschildw:
Add sageiIT VPC to TGW/VPN, give synapseprod devs access to VPCs, give synapsedw-users access to dw VPC, give sageit-users access to sageIT VPC.


org-formation/710-tgw/_tasks.yaml
```diff
@@ -137,6 +141,7 @@ Mappings:
       '607346494281': 'agoravpc'      # org-sagebase-agora-dev
       '465877038949': 'cesspoolvpc'      # org-sagebase-scipooldev
       '237179673806': 'internalpoolvpc'      # org-sagebase-scipooldev
+      '797640923903': 'sagedefaultvpc' # org-sagebase-sageit
```
zaro0508, line 144: should be `defaultvpc`

org-formation/720-client-vpn/_tasks.yaml
```diff
@@ -180,6 +183,14 @@ VpnAuthRoutes:
         AccessGroups:
           - "aws-admins"
           - "scientists"
+      # org-sagebase-sageit
+      sagedefaultvpc:
```
zaro0508, line 187: should be `defaultvpc`

org-formation/710-tgw/_tasks.yaml
```diff
@@ -114,6 +115,9 @@ TransitGatewayRoutes:
       # org-sagebase-synapsedw
       vpc:
         CIDR: "10.12.0.0/16"
+      # org-sagebase-sageit
+      sagedefaultvpc:
```
zaro0508, line 119: i don't see a CFN stack in sageit called sagedefaultvpc, I do see a stack called defaultvpc. This should match the CFN stack name of the VPC, which in this case should probably be defaultvpc

zaro0508 (changes requested):

org-formation/710-tgw/_tasks.yaml
xschildw: Good catch. Thx!

zaro0508 (approved):

PR #476, opened 2022-05-06
IT-1874: Remove Sophos UTM

xschildw:
This PR removes the Sophos UTM.

depends on #477 #478 #479 #486


zaro0508 (approved):
This template is not managed by CI but was deployed manually into cloudformation.  We need to manually go into cloudformation and delete it.  I think the peering stuff needs to be removed before this base stack is removed.

PR #479, opened 2022-05-06
IT-18874: Remove routes from spokes to admincentral

xschildw:
All spokes except synapsedev/synapseprod (need to check these first as they could be different).


zaro0508 (approved):
Don't we need to remove in a certain order?  like remove peering from all member accounts before removing peering from master account?

xschildw:
Correct, this removes the routes to central in all spokes.

PR #480, opened 2022-05-11
Migrate bridge-ops to template-managed host

DwayneJengSage:
https://sagebionetworks.jira.com/browse/BRIDGE-3185 - Migrate bridge-ops to template-managed host

PR Checklist:
[X] Describe and explain your intentions for this change
[X] Setup pre-commit and run the validators (info in README.md)


sceptre/bridge/config/prod/bridge-ops-djeng.yaml
```diff
@@ -0,0 +1,16 @@
...
+stack_name: jc-ec2-linux
+dependencies:
+  - prod/bootstrap.yaml
+parameters:
+  JcUserId: dwayne.jeng@sagebase.org
+  KeyName: bridge-ops
+  InstanceType: t3a.micro
+  JcConnectKey: !ssm /infra/JcConnectKey-djeng
+  JcServiceApiKey:
+  JcSystemsGroupId:
```
DwayneJengSage, line 13: Not sure what I'm supposed to put for these attributes.

sceptre/bridge/config/prod/service-linked-roles.yaml
```diff
@@ -0,0 +1,6 @@
+template:
+  type: http
+  url: https://{{stack_group_config.admincentral_cf_bucket}}.s3.amazonaws.com/aws-infra/v0.3.0/service-linked-roles.yaml
```
zaro0508, line 3: is there a reason why you are deploying this into the account? It's not security related and I don't think you actually need it.

sceptre/bridge/config/prod/bridge-ops-djeng.yaml
zaro0508: These are to allow access via jumpcloud & VPN.  You guys already have full  access to bridge accounts so I'm wondering if these things are needed at all?

The `jc-ec2-linux.j2` template was specifically created for scicomp provisioning and to allow scientist access to the instance without giving them access to the AWS account.  I think this template probably does not fit your use case?

sceptre/bridge/config/prod/service-linked-roles.yaml
DwayneJengSage: I was following the instructions provided to me at https://sagebionetworks.jira.com/browse/BRIDGE-3185?focusedCommentId=145723 I had assumed that I needed this. Perhaps I misunderstood the instructions?

sceptre/bridge/config/prod/bridge-ops-djeng.yaml
DwayneJengSage: The use case is that we need to be able to SSH into this EC2 machine and run scripts from the command-line. We also want to ensure the EC2 instance is accessible only from VPN.

sceptre/bridge/config/prod/service-linked-roles.yaml
zaro0508: I don't see anything in the dock that references service-linked-roles.  I would suggested removing this template.

sceptre/bridge/config/prod/bridge-ops-djeng.yaml
```diff
@@ -0,0 +1,16 @@
+template:
+  type: http
+  url: https://{{stack_group_config.admincentral_cf_bucket}}.s3.amazonaws.com/aws-infra/v0.4.0/EC2/jc-ec2-linux.j2
```
zaro0508, line 3: I would suggest changing this to reference `https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/master/templates/EC2/basic-ec2.j2`

sceptre/bridge/config/prod/bridge-ops-djeng.yaml
zaro0508: the `basic-ec2.j2` templates does not require setting `JcUserId`, `KeyName`, `JcConnectKey`, `JcServiceApiKey`, `JcSystemsGroupId`,  or `AMIId` parameters.

Once provisioned you can access the instance using the [AWS SSM session manager](https://sagebionetworks.jira.com/wiki/spaces/IT/pages/2632286259/AWS+SSM+Session+Manager#Access-to-a-private-EC2-instance%3A) instead of the VPN.

zaro0508 (changes requested):

sceptre/bridge/config/prod/bridge-ops.yaml
```diff
@@ -0,0 +1,10 @@
+template:
+  type: http
+  url: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/master/templates/EC2/basic-ec2.j2
+stack_name: bridge-ops
+dependencies:
+  - prod/bootstrap.yaml
+parameters:
+  InstanceType: t3a.micro
```
zaro0508, line 8: `t3a.micro` probably won't work because that's for ARM.  The template is used is [targeting intel (x86) AMIs](https://github.com/Sage-Bionetworks/aws-infra/blob/master/templates/EC2/basic-ec2.j2#L26).  You'll probably want to change to `t3.micro`

zaro0508 (approved):

sceptre/bridge/config/prod/bridge-ops.yaml
DwayneJengSage: Fixed, thanks!

zaro0508 (commented):
@DwayneJengSage sorry, i forgot to ask initially.  This repo is supposed to be for deploying base infrastructure for bridge account therefore deploying `maintenance.yaml` makes sense here however `bridge-ops.yaml` is just an EC2.  Why deploy an EC2 from this repo and not from one of the Bridge infra specific repos like `BridgeServer2-infra`, `Bridge-Exporter-infra`, and `BridgeWorkerPlatform-infra`?

DwayneJengSage:
> @DwayneJengSage sorry, i forgot to ask initially. This repo is supposed to be for deploying base infrastructure for bridge account therefore deploying `maintenance.yaml` makes sense here however `bridge-ops.yaml` is just an EC2. Why deploy an EC2 from this repo and not from one of the Bridge infra specific repos like `BridgeServer2-infra`, `Bridge-Exporter-infra`, and `BridgeWorkerPlatform-infra`?

The purpose of this host is to run one-off jobs in production. It exists separate from Server2, Exporter, or Worker, so it doesn't make sense to bundle this with any of those projects.

sceptre/bridge/config/prod/bridge-ops.yaml
```diff
@@ -0,0 +1,10 @@
+template:
+  type: http
+  url: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/master/templates/EC2/basic-ec2.j2
+stack_name: bridge-ops
+dependencies:
+  - prod/bootstrap.yaml
```
zaro0508, line 6: sorry, i missed this.  The VpcId and SubnetId that's passed into this template are created by [BridgeServer2-vpc.yaml](https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/sceptre/bridge/config/develop/BridgeServer2-vpc.yaml) so probably should replace this with `prod/BridgeServer2-vpc.yaml`

zaro0508 (changes requested):
ok, sounds fine.  just one minor thing and we should be good to merge

zaro0508 (approved):

PR #482, opened 2022-05-11
[IT-1818] Add data and tooling dev AWS account

thomasyu888:
PR Checklist:
- [x] Describe and explain your intentions for this change

See [JIRA ticket](https://sagebionetworks.jira.com/browse/IT-1818).

- [x] Setup pre-commit and run the validators (info in README.md)

PR #483, opened 2022-05-11
[IT-1905] Add include prod account

thomasyu888:
PR Checklist:
See JIRA for more details


org-formation/organization.yaml
```diff
@@ -287,6 +288,20 @@ Organization:
         budget-alarm-threshold: 20000
         budget-alarm-threshold-email-recipient: aws-nlpsandbox@sagebase.org
 
+  IncludeAccount:
+    Type: OC::ORG::Account
+    Properties:
+      AccountName: IncludeAccount
```
zaro0508, line 294: This is a very generic name.  Could we come up with something a little more specific that includes some indication whether it's gonna be a dev or prod account?

org-formation/organization.yaml
```diff
@@ -287,6 +288,20 @@ Organization:
...
+      AccountId: '841415736403'
```
zaro0508, line 295: new account correct?  so how do you know the account id alread?

org-formation/organization.yaml
```diff
@@ -287,6 +288,20 @@ Organization:
...
+      RootEmail: ibc-ops@sagebase.org
+      Alias: org-sagebase-include
```
zaro0508, line 297: more specific alias

org-formation/organization.yaml
```diff
@@ -287,6 +288,20 @@ Organization:
         budget-alarm-threshold: 20000
         budget-alarm-threshold-email-recipient: aws-nlpsandbox@sagebase.org
 
+  IncludeAccount:
+    Type: OC::ORG::Account
+    Properties:
+      AccountName: IncludeAccount
+      AccountId: '841415736403'
+      RootEmail: ibc-ops@sagebase.org
```
zaro0508, line 296: data tooling account in PR #482 is already using this email.  AWS requires a unique email for every AWS account.

zaro0508 (changes requested):

org-formation/organization.yaml
thomasyu888: Since INCLUDE is the grant name, do you recommend something like `IncludeProdAccount`?

org-formation/organization.yaml
thomasyu888: Ah, I don't know the AccountId, I would need this information

org-formation/organization.yaml
thomasyu888: `org-sagebase-include-prod`?

org-formation/organization.yaml
thomasyu888: understood, I will create a new google group for include.

thomasyu888:
Thanks for your review @zaro0508. I made comments and will make changes.

org-formation/organization.yaml
```diff
@@ -287,6 +288,19 @@ Organization:
         budget-alarm-threshold: 20000
         budget-alarm-threshold-email-recipient: aws-nlpsandbox@sagebase.org
 
+  IncludeAccount:
+    Type: OC::ORG::Account
+    Properties:
+      AccountName: IncludeAccount
+      RootEmail: aws-include-prod@sagebase.org
+      Alias: org-sagebase-include
+      Tags:
```
thomasyu888, line 297: @zaro0508, is there a tag we can add to enable all the resource that are created in this account to be tagged with a specific ProgramCode?

org-formation/organization.yaml
zaro0508: If we can't be more creative then I guess that's fine.  Please also make AccountName the same as Alias.  I guess we like the `org-sagebase-*` convention.

org-formation/organization.yaml
zaro0508: ok

org-formation/organization.yaml
zaro0508: Not that i'm aware of.  This tag is for the account and I believe we will be separating cost by account.  Which essentially means that we will be applying one cost center to the entire account.

Now that I think about it we should add a `CostCenter` tag to every account we create.  Please do that.

zaro0508 (changes requested):

thomasyu888:
Thanks again for your review.  I changed the account name to be `IncludeChopProdAccount` to be more descriptive and added the cost center.

PR #487, opened 2022-05-16
IT-1874: remove peering hub2spoke

xschildw:
This PR removes the hub-to-spoke peerings + associated resources. Need to merge #486 first.


zaro0508 (approved):
I think you might need to delete peering from member account before master account, or maybe the other way around?  I'm not sure but i think you might have to do one before the other.

PR #488, opened 2022-05-16
[IT-1921] Update costcenter tags for some D&T AWS accounts

thomasyu888:
AWS accounts can be tagged with `CostCenter` so that IT can report all AWS expenses under the account to a particular code if a code exists per Bruce's comment [on slack](https://sagebionetworks.slack.com/archives/CEFQD0KU1/p1652735680132299).  This is an effort to add the CostCenter tags and have a discussion about the added tags.

PR Checklist:


org-formation/organization.yaml
```diff
@@ -314,6 +315,7 @@ Organization:
         <<: !Include ./_default_org_tags.yaml
         Department: CNB
         Project: Challenges
+        CostCenter: NO PROGRAM / 000000
```
thomasyu888, line 318: This account is assigned with NO PROGRAM code because it is meant to be a "team account" where we host different challenges.  The idea is that the resources internal to this account should be tagged accordingly.

org-formation/organization.yaml
```diff
@@ -328,6 +330,7 @@ Organization:
         <<: !Include ./_default_org_tags.yaml
         Department: IBC
         Project: mobile-toolbox
+        CostCenter: Mobile Toolbox Project Core / 112501
```
thomasyu888, line 333: Needs to be confirmed with Larsson

org-formation/organization.yaml
```diff
@@ -342,6 +345,7 @@ Organization:
         <<: !Include ./_default_org_tags.yaml
         Department: IBC
         Project: Infrastructure
+        CostCenter: Mobile Toolbox Project Core / 112501
```
thomasyu888, line 348: Needs to be confirmed with Larsson

org-formation/organization.yaml
```diff
@@ -356,6 +360,7 @@ Organization:
         <<: !Include ./_default_org_tags.yaml
         Department: CompOnc
         Project: iAtlas
+        CostCenter: iAtlas 3 / 314900
```
thomasyu888, line 363: Needs confirmation from James

org-formation/organization.yaml
```diff
@@ -371,6 +376,7 @@ Organization:
         <<: !Include ./_default_org_tags.yaml
         Department: CompOnc
         Project: htan
+        CostCenter: HTAN-DFCI / 120100  # HTAN Supp DFCI / 121700
```
thomasyu888, line 379: Needs confirmation from HTAN team

org-formation/organization.yaml
```diff
@@ -385,6 +391,7 @@ Organization:
         <<: !Include ./_default_org_tags.yaml
         Department: IBC
         Project: Infrastructure
+        CostCenter: NO PROGRAM / 000000
```
thomasyu888, line 394: Since nextflow-tower supports multiple projects, see the granular budget report comment.  Do we leave out the CostCenter tag on the account level if there is already more granular budget reporting?

org-formation/organization.yaml
```diff
@@ -399,6 +406,7 @@ Organization:
...
```
thomasyu888, line 409: Since nextflow-tower supports multiple projects, see the granular budget report comment. Do we leave out the CostCenter tag on the account level if there is already more granular budget reporting?

org-formation/organization.yaml
```diff
@@ -286,6 +286,7 @@ Organization:
         Department: CompOnc
         Project: NLP-Sandbox
         budget-alarm-threshold: 20000
+        CostCenter: CSBC - NCI / 101300  # need to be confirmed
```
thomasyu888, line 289: Needs confirmation from Abby

org-formation/organization.yaml
```diff
@@ -286,6 +286,7 @@ Organization:
         Department: CompOnc
         Project: NLP-Sandbox
         budget-alarm-threshold: 20000
+        CostCenter: CD2H Total / 10133
```
thomasyu888, line 289: Seems to be missing from program code list : But confirmed from Abby

org-formation/organization.yaml
brucehoff: We can consider resource-level cost center tags to override the default, account level tag.  From this perspective I think it's fine to add `NO PROGRAM / 000000` as an account-level tag for the Nextflow account.

org-formation/organization.yaml
```diff
@@ -314,6 +315,7 @@ Organization:
         <<: !Include ./_default_org_tags.yaml
         Department: CNB
         Project: Challenges
+        CostCenter: NO PROGRAM / 000000  # resource-level cost center tags to override this default
```
thomasyu888, line 318:
```suggestion
        CostCenter: NO PROGRAM / 000000  # resource-level cost center tags will override this default
```

org-formation/organization.yaml
```diff
@@ -385,6 +391,7 @@ Organization:
         <<: !Include ./_default_org_tags.yaml
         Department: IBC
         Project: Infrastructure
+        CostCenter: NO PROGRAM / 000000  # resource-level cost center tags to override this default
```
thomasyu888, line 394:
```suggestion
        CostCenter: NO PROGRAM / 000000  # resource-level cost center tags will override this default
```

org-formation/organization.yaml
```diff
@@ -399,6 +406,7 @@ Organization:
...
```
thomasyu888, line 409:
```suggestion
        CostCenter: NO PROGRAM / 000000  # resource-level cost center tags will override this default
```

org-formation/organization.yaml
thomasyu888, line 379:
```suggestion
        CostCenter: HTAN-DFCI / 120100
```

org-formation/organization.yaml
```diff
@@ -314,6 +315,7 @@ Organization:
         <<: !Include ./_default_org_tags.yaml
         Department: CNB
         Project: Challenges
+        CostCenter: NO PROGRAM / 000000  # resource-level cost center tags will override this default
```
zaro0508, line 318: What would happen if, for some reason, a resource level tag was not applied or was missing?

zaro0508 (changes requested):
* I think we'll want an `AccountOwner` tag on every account as well.
* would you be able to slap a CostCenter and AccountOwner tag on agora as well?

org-formation/organization.yaml
brucehoff: > What would happen if, for some reason, a resource level tag was not applied or was missing?

 Then we would categorize the resource as not being for a directly funded project.  At the end of the month we would let the account owner know how much activity in their account lacks a funding source.  If the amount is small then it's OK. The owner won't care.  If the amount is large, the account owner can question their team and perhaps add missing CostCenter tags.

zaro0508: I'm not opposed your proposed approach but I'll share my concerns..

Ok, however I think you're making an assumption that we have some reporting mechanism in place to break down costs and intelligently give the account owner info about what's being tracked and what isn't.  The question is do we wan to accept `NO PROGRAM` until we have something like that in place?  

Also from my experience using AWS I will guarantee that there will be things that won't have tags on them even if you use cloudformation to provision resources.  Whether that's due to AWS bugs or the idiosyncratic ways that tags work in AWS.

org-formation/organization.yaml
thomasyu888: Confirmed with Sonia.

thomasyu888:
Thanks @zaro0508 and @brucehoff !  I'm not opposed to adding `AccountOwner`, but...

* sometimes multiple people "own" accounts, should I specify a list?  The full list of AWS accounts and system owners are found here: https://sagebionetworks.jira.com/wiki/spaces/IT/pages/352976898/Infrastructure+Inventory
* Were you thinking of just adding their email?

org-formation/organization.yaml
brucehoff: > I think you're making an assumption that we have some reporting mechanism in place to break down costs and intelligently give the account owner info about what's being tracked and what isn't.

Yes, none of this is useful until we have the script that that processes the AWS cost-and-usage report to provide rolled-up summaries to account owners and project PIs.  I have been tinkering with the AWS cost-and-usage report for a few months and am confident about our ability to "roll up" costs by account, cost center, user, etc. in a variety of ways.

> The question is do we wan[t] to accept NO PROGRAM until we have something like that in place?

I think so:  I'm not sure what else we can do besides (or in advance of) building and running the envisioned accounting script.  Is there an alternative we should consider?

> I will guarantee that there will be things that won't have tags on them even if you use cloudformation to provision resources.

Yes, I agree completely.  I imagine each time we roll up costs there will be a bunch of untagged 'dark matter' that we might need to dig into and analyze.  Until we see if the $ amount is small or large I don't know whether it will be a problem.

By the way, the reason I like the "NO PROGRAM" code at the resource level is that it allows us to differentiate between (1) a resource we explicitly say is not for a funded program and (2) an unexpected cost we aren't tracking.  It's a reasonable question to ask, "How much of (2) is there in our monthly bill?"


So we need to look at how much 'untagged' cost there is and (if it's significant) dive in to figure out why.  To me the strength of the "NO PROGRAM" tag is it lets us see the difference between items we explicitly say are for no particular program and items that we have just failed to tag.

brucehoff:
> sometimes multiple people "own" accounts, should I specify a list? 

@thomasyu888 The idea is that the "Owner" is a single person who is ultimately responsible for the account costs.  One way to identify the individual is the common ancestor for the account users in the Sage org-chart.  So if you, Tess and Phil all use a 'mobile health' AWS account the the "Owner" might be you, Tom, as the director of the team.  Or it could be Brian O'Connor.  Whoever it is, it would be someone (1) with the responsibility to answer for costs incurred in the account and (2) the authority to make the other account users answer for their activity in the account.

org-formation/organization.yaml
```diff
@@ -356,6 +365,8 @@ Organization:
         <<: !Include ./_default_org_tags.yaml
         Department: CompOnc
         Project: iAtlas
+        AccountOwner: undecided@sagebase.org
```
thomasyu888, line 368: I know its documented that Andrew Lamb is the System owner, but I'm unsure how much involvement he actually has.  Would need more discussion internally first, but Brian is out.

org-formation/organization.yaml
```diff
@@ -371,6 +382,8 @@ Organization:
         <<: !Include ./_default_org_tags.yaml
         Department: CompOnc
         Project: htan
+        CostCenter: HTAN-DFCI / 120100
+        AccountOwner: undecided@sagebase.org
```
thomasyu888, line 386: I know its documented that Bruno is the System owner, but I'm unsure how much involvement he actually has. Would need more discussion internally first, but Brian O'connor is out sick.

org-formation/organization.yaml
```diff
@@ -412,6 +429,10 @@ Organization:
         <<: !Include ./_default_org_tags.yaml
         Department: SysBio
         Project: amp-ad
+        # Anna G confirmed that AMP-AD charges should technically
+        # be charged to STRIDES account but this tag works
+        CostCenter: NIA AMP-AD CC / 101500
+        AccountOwner: anna.greenwood@sagebase.org
```
thomasyu888, line 435: Will need to confirm that Anna is the correct person to be the account owner.

org-formation/organization.yaml
thomasyu888: Apologies for a slight tangent.  I was going to try to test out some of the scripts found in this repo: https://github.com/ucsc-cgl/aws-reporting to see what the output was.  I'll report back with results.

zaro0508: >  Is there an alternative we should consider

One alternative is to send the bill for the entire account to the responsible department.  We can either use `Department` tag, make up a department `CostCenter` tag, or even use the `AccountOwner` tag to identify who should get the bill.  Then let the individual department figure out specific resources costs themselves?   It looks  like @thomasyu888 is already looking for ways to slice and dice the costs.  There are also SAS services that can do this for us, no need to re-invent the wheel here.

thomasyu888: We will most likely host certain dev resources that will be utilized across by multiple projects so in those scenarios, it might even be difficult for us to split the costs, so we would probably just put it in "NO PROGRAM" code.  At the end of the day, maybe System Admins can try our best to make sure everything is tagged appropriately at the resource level. It'll be easier to tell once we get the costs of "untagged resources".

@zaro0508, agreed that we should re-use services when possible.  One thing to consider is that the system/account owners would potentially need access whatever billing information that allows us to use this service OR the reports are sent to us.  As an example, the script I linked above requires access to an S3 bucket (AWS billing can publish detailed billing results to an S3 bucket if we choose to).  Not saying we have to go that route, just some food for thought.

brucehoff (approved):

zaro0508 (approved):

PR #491, opened 2022-05-19
[IT-1703] personal AWS account

zaro0508:
Create an AWS account for @tthyer

### DIFF
--- a/org-formation/organization.yaml
+++ b/org-formation/organization.yaml
@@ -445,3 +445,20 @@ Organization:
       Alias: org-sagebase-sceptredev
       Tags:
         <<: !Include ./_default_org_tags.yaml
+
+
+  #---------- Personal User Accounts  -----------
+  hxo89cta7a0e:
+    Type: OC::ORG::Account
+    Properties:
+      AccountName: org-sagebase-hxo89cta7a0e
+      RootEmail: aws-hxo89cta7a0e@sagebase.org
+      Alias: org-sagebase-hxo89cta7a0e
+      Tags:
+        <<: !Include ./_default_org_tags.yaml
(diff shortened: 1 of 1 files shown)


thomasyu888:
Could we create one for me first. Tess is currently on leave.

zaro0508:
self service @thomasyu888, we are just waiting for your PR  to setup your own account 😄

org-formation/organization.yaml
```diff
@@ -445,3 +445,20 @@ Organization:
...
+    Type: OC::ORG::Account
+    Properties:
+      AccountName: org-sagebase-hxo89cta7a0e
+      RootEmail: aws-hxo89cta7a0e@sagebase.org
+      Alias: org-sagebase-hxo89cta7a0e
+      Tags:
+        <<: !Include ./_default_org_tags.yaml
+        Department: TBD
+        Project: TBD
+        CostCenter: TBD
```
brucehoff, line 461: Instead of "TBD" can we use a valid Cost Center, like 000000 NO PROGRAM?

brucehoff (changes requested):

org-formation/organization.yaml
zaro0508: We are just waiting for @tthyer to review and provide a valid values for these tags.

thomasyu888: Tess won't be reviewing these for awhile, so let's just leave this PR open until she's back from leave.

tthyer:
Let's close this PR for now. I understand there's a DPE account I can use if needed and I can request to have this re-opened in the future if it becomes necessary, but it's not looking likely at the moment.

PR #506, opened 2022-06-02
step to validate cfn templates

zaro0508:
Add an additional `print-task` step in CI to validate templates.
The `print-task` command will generate templates with inputs which
may find failures that the `validate` command may not

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
     "print-tasks": "npx org-formation print-tasks ./org-formation/_tasks.yaml --verbose --print-stack --output yaml --max-concurrent-stacks 100 --max-concurrent-tasks 100",
-    "print-tasks-failfast": "npx org-formation print-tasks ./org-formation/_tasks.yaml --verbose --print-stack --output yaml --max-concurrent-stacks 1 --max-concurrent-tasks 1 --failed-tasks-tolerance 0  --failed-stacks-tolerance 0",
+    "print-tasks-failfast": "npx org-formation print-tasks ./org-formation/_tasks.yaml --verbose --print-stack --output yaml --max-concurrent-stacks 1 --max-concurrent-tasks 1 --failed-tasks-tolerance 1  --failed-stacks-tolerance 1",
     "perform-tasks": "npx org-formation perform-tasks ./org-formation/_tasks.yaml --verbose --print-stack --max-concurrent-stacks 1 --max-concurrent-tasks 1",
     "ci-perform-tasks": "npx org-formation perform-tasks ./org-formation/_tasks.yaml --verbose --print-stack --max-concurrent-stacks 1 --max-concurrent-tasks 1 --perform-cleanup",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -6,6 +6,7 @@
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
     "print-tasks": "npx org-formation print-tasks ./org-formation/_tasks.yaml --verbose --print-stack --output yaml --max-concurrent-stacks 100 --max-concurrent-tasks 100",
+    "print-tasks-failfast": "npx org-formation print-tasks ./org-formation/_tasks.yaml --verbose --print-stack --output yaml --max-concurrent-stacks 1 --max-concurrent-tasks 1 --failed-tasks-tolerance 1  --failed-stacks-tolerance 1",
```
xschildw, line 9: Maybe I'm missing it but I don't see the print-tasks command in the docs (there's a print-stacks command that prints the generated cf templates). I looked at [](Maybe I'm missing it but I don't see the print-tasks command in the docs (there's a validate-tasks and a print-stacks command that prints the generated cf templates). I looked at https://github.com/org-formation/org-formation-cli/blob/master/docs/cli-reference.md#operations-on-task-files. Any concern with sensitive parameters being logged?

zaro0508: Good questions.  You are correct, `print-tasks` does not seem to be in the docs however it does exist.  I cannot find any references in org-formation that would expose sensitive data.

```
➜  npx org-formation print-tasks --help
Usage: print-tasks [options] <tasksFile>

Will print out all cloudformation templates that will be deployed by tasksFile

Options:
  --max-concurrent-tasks <max-concurrent-tasks>        maximum number of tasks to be executed concurrently (default: 1)
  --max-concurrent-stacks <max-concurrent-stacks>      maximum number of stacks (within a task) to be executed concurrently (default: 1)
  --failed-tasks-tolerance <failed-tasks-tolerance>    the number of failed tasks after which execution stops (default: 99)
  --failed-stacks-tolerance <failed-stacks-tolerance>  the number of failed stacks (within a task) after which execution stops (default: 0)
  --organization-file [organization-file]              organization file used for organization bindings
  --parameters [parameters]                            parameters used when creating build tasks from tasks file
  --output <output>                                    the serialization format used when printing stacks. Either json or yaml. (default: "yaml")
  --output-path <output-path>                          path, within the root directory, used to store printed templates (default: "./.printed-stacks/")
  --output-cross-account-exports <output-path>         when set, output well generate cross account exports as part of cfn parameter (default: false)
  --no-print-parameters                                will not print parameter files when printing stacks
  --debug-templating [debug-templating]                when set to true the output of text templating processes will be stored on disk (default: false)
  --state-bucket-name [state-bucket-name]              bucket name that contains state file (default: "organization-formation-${AWS::AccountId}")
  --state-object [state-object]                        key for object used to store state (default: "state.json")
  --profile [profile]                                  aws profile to use
  --print-stack                                        will print stack traces for errors
  --verbose                                            will enable debug logging
  --no-color                                           will disable colorization of console logs
  --master-account-id [master-account-id]              run org-formation on a build account that functions as a delegated master account
  --is-partition                                       is run on a partition
  --partition-keys                                     is to indicate to look for partition access keys
  --partition-profile [profile]                        aws partition profile to use
  --partition-region [region]                          aws partition region to use
  -h, --help                                           output usage information
```

xschildw (approved):

PR #512, opened 2022-06-06
refactor org formation accounts files

zaro0508:
Move project account definitions to a separate file to keep the
file size manageable.


brucehoff (approved):

zaro0508:
We can't do this right now due to orgf issue https://github.com/org-formation/org-formation-cli/issues/293

PR #515, opened 2022-06-07
IT-1971 Adding Dnt Dev Viewer Group

DiepThach:
adding Sso integration for Jumpcloud to allow viewer access to dntdev account.


zaro0508:
@DiepThach this is failing on pre-commit validation.  we'll need to fix this

PR #518, opened 2022-06-09
it-1874-remove-peering-authorizer-role

xschildw:
This PR removes the peering authorizer role (used to accept peering in target account). Originally targeted at nlpsandbox but after looking at other accounts listed I think it can be removed everywhere.

Depends on #517


zaro0508 (approved):
Merging should automatically remove this cfn stack from all referenced accounts.

PR #528, opened 2022-06-16
[IT-1766] Restore AWS region restriction

zaro0508:
Restrict accounts from creating resources in regions outside of the US
region.


zaro0508:
Region restriction was disabled due to issue IT-1766, we need to make sure it can work before merging this PR.

brucehoff (approved):
This seems OK as long as it omits Synapse which does, in places, require the use of resources outside the US.

xschildw (approved):
So you're putting it back so AWS can investigate?

zaro0508:
this has been replaced with PR #624

PR #529, opened 2022-06-18
[IT-1974] SCP to deny registering domains

zaro0508:
Setup a service control policy to prevent all accounts from registering
domains, except for the org-sagebase-sageit account.  We want to control
the domains that are created at Sage.


brucehoff:
@jesusaurus This is a good opportunity to start learning about org-formation and how it can use "service control policies" to limit what can be done in an AWS account we create.

brucehoff (commented):
This looks OK to me but I'd like @xschildw to approve since he does a lot of name registration:  Is it OK with you that we do it all in one centralized place like this?

xschildw (approved):

xschildw:
Yes, all the registrations in sage-it (where most of the portals reside) and DNS zones where they need to be managed.

PR #540, opened 2022-07-18
[IT-1046] Create Monthly Cost and Usage Reports

jesusaurus:
Automate the creation of a cost-and-usage report definition, copying
settings from the manually-created report definition.

This report will be analyzed by an automated process to generate
per-program cost reports.


org-formation/050-costs/_tasks.yaml
```diff
@@ -11,3 +11,14 @@ AnomalyDetectorService:
...
   Parameters:
     Subscriber: !GetAtt CurrentAccount.RootEmail
+
+CostAndUsageReports:
+  Type: update-stacks
+  Template: cur.yaml
+  StackName: !Sub '${resourcePrefix}-CUR-Monthly'
+  DefaultOrganizationBinding:
+    IncludeMasterAccount: true
+    Account: '*'
```
zaro0508, line 21: this only needs to be deployed to the Master account so what you want is `Account: !Ref MasterAccount`

org-formation/050-costs/_tasks.yaml
```diff
@@ -11,3 +11,14 @@ AnomalyDetectorService:
...
+    Region: !Ref primaryRegion
```
zaro0508, line 22: we can probably remove this everything deploys to `primaryRegion` by default.  you only need to  set this if you want to deploy to non us-east-1 regions

org-formation/050-costs/cur.yaml
```diff
@@ -0,0 +1,17 @@
+AWSTemplateFormatVersion: '2010-09-09'
+Description: 'Monthly Cost and Usage Report'
+Parameters:
+  curBucket:
+    Type: String
+Resources:
+  MonthlyCostReport:
+    Type: "AWS::CUR::ReportDefinition"
+    Properties:
+      ReportName: !Sub ""
```
zaro0508, line 10: `!Sub ""` will evaluate to empty.

org-formation/050-costs/cur.yaml
```diff
@@ -0,0 +1,17 @@
...
+      ReportVersioning: False
+      TimeUnit: "hourly"
+      Compression: "parquet"
+      Format: "parquet"
+      S3Region: !Ref primaryRegion
+      S3Prefix: "hourly"
```
zaro0508, line 16: consider parameterizing these configs so that you can create CUR with different parameters using the same template?

zaro0508 (changes requested):

org-formation/050-costs/cur.yaml
```diff
@@ -0,0 +1,17 @@
...
+Parameters:
+  curBucket:
+    Type: String
+Resources:
+  MonthlyCostReport:
+    Type: "AWS::CUR::ReportDefinition"
+    Properties:
+      ReportName: !Sub ""
+      ReportVersioning: False
+      TimeUnit: "hourly"
```
brucehoff, line 12: I think TimeUnit should be MONTHLY, not HOURLY.  I recall that I first made an HOURLY report.  At the end of the month the report was massive, spanning multiple files.  Changing to MONTHLY provided all the needed information in one file that was just a few MB.

brucehoff (changes requested):

org-formation/050-costs/_tasks.yaml
jesusaurus: Oh, we don't need to deploy it to all accounts in order to collect information on all accounts? Cool

org-formation/050-costs/cur.yaml
jesusaurus: Whoops, I meant to put simple and unique value here.

org-formation/050-costs/cur.yaml
jesusaurus: I moved the frequency to a parameter, but left the compression and format as opinionated values. Let me know if they should be parameterized as well.

org-formation/050-costs/cur.yaml
```diff
@@ -0,0 +1,24 @@
...
+Resources:
+  MonthlyCostReport:
+    Type: "AWS::CUR::ReportDefinition"
+    Properties:
+      ReportName: !Sub "${AWS::StackName}"
+      ReportVersioning: False
+      TimeUnit: !Ref Frequency
+      Compression: "parquet"
+      Format: "parquet"
+      S3Region: !Ref primaryRegion
```
zaro0508, line 22: `primaryRegion` is not a CFN parameter, it's only available to orgf configs.  you can use the [CFN pseudo parameter](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/pseudo-parameter-reference.html) `!Ref AWS::Region`

org-formation/050-costs/cur.yaml
```diff
@@ -0,0 +1,24 @@
...
+    AllowedValues:
+    - "monthly"
+    - "weekly"
+    - "hourly"
+Resources:
+  MonthlyCostReport:
+    Type: "AWS::CUR::ReportDefinition"
+    Properties:
+      ReportName: !Sub "${AWS::StackName}"
+      ReportVersioning: False
```
zaro0508, line 18: [the documentation ](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-cur-reportdefinition.html#cfn-cur-reportdefinition-s3prefix) says these are the possible options? `Allowed values: CREATE_NEW_REPORT | OVERWRITE_REPORT`

org-formation/050-costs/cur.yaml
```diff
@@ -0,0 +1,24 @@
...
+      TimeUnit: !Ref Frequency
+      Compression: "parquet"
+      Format: "parquet"
+      S3Region: !Ref primaryRegion
+      S3Prefix: !Ref Frequency
```
zaro0508, line 23: your report name is set to `${resourcePrefix}-CUR-Monthly` which evaluates to `sagebase-CUR-Monthly` so with this prefix it will be `Monthly-sagebase-CUR-Monthly`.  seems kinda redundant, no?

org-formation/050-costs/cur.yaml
```diff
@@ -0,0 +1,24 @@
...
+    Type: String
+  Frequency:
+    Type: String
+    Default: "monthly"
+    AllowedValues:
+    - "monthly"
+    - "weekly"
+    - "hourly"
+Resources:
+  MonthlyCostReport:
```
zaro0508, line 14: I think it might make more sense to just create additional resources here when we want to create a additional reports.  if you agree then maybe we don't need to parameterize `Frequency`?

org-formation/050-costs/cur.yaml
```diff
@@ -0,0 +1,24 @@
...
+    Type: "AWS::CUR::ReportDefinition"
+    Properties:
+      ReportName: !Sub "${AWS::StackName}"
+      ReportVersioning: False
+      TimeUnit: !Ref Frequency
+      Compression: "parquet"
+      Format: "parquet"
+      S3Region: !Ref primaryRegion
+      S3Prefix: !Ref Frequency
+      S3Bucket: !Ref Bucket
```
zaro0508, line 24: I think it would be better to create a custom bucket for these reports.  your thoughts?

zaro0508 (changes requested):

org-formation/050-costs/cur.yaml
jesusaurus: Yeah, it is very redundant, especially since the prefix is required but AWS will also add its own prefixes based on the date and frequency, giving us a full s3 path that looks like `cost-and-usage-reports.sagebase.org/Monthly/Monthly/year=2022/month=7/Monthly-00001.parquet` where our prefix parameter is the top-level (first) "Monthly". But I suppose I could change the prefix to "CostReport" or something similar.

org-formation/050-costs/cur.yaml
jesusaurus: We are already using a dedicated bucket for cost reports. Do you mean creating the bucket in the template rather than relying on one that has been created out-of-band?

zaro0508: yes, that is correct

jesusaurus: Yeah, I like that idea for both disaster recovery and also it makes it easier to clean up the manually-created proof-of-concept reports.

org-formation/050-costs/cur.yaml
jesusaurus: yeah, done

org-formation/040-budgets/cur.yaml
```diff
@@ -0,0 +1,42 @@
...
+Description: 'Monthly Cost and Usage Report'
+Parameters:
+  resourcePrefix:
+    Type: String
+Resources:
+  CostReportBucket:
+    Type: "AWS::S3::Bucket"
+    Properties:
+      BucketName: !Sub "${resourcePrefix}-cost-reports"
+      AccessControl: PublicRead
```
zaro0508, line 11: hmm, i think `public read` will allow anybody access to the bucket.  Do we really want that?

org-formation/040-budgets/cur.yaml
```diff
@@ -0,0 +1,42 @@
...
+      Bucket: !Ref CostReportBucket
+      PolicyDocument:
+        Version: 2012-10-17
+        Statement:
+          - Action:
+              - 's3:GetBucketAcl'
+              - 's3:GetBucketPolicy'
+              - 's3:PutObject'
+            Effect: Allow
+            Resource: !Join
```
zaro0508, line 24: AWS recommends using `!Sub` over `!Join`.  i think i agree with them.

org-formation/040-budgets/_tasks.yaml
```diff
@@ -12,3 +12,12 @@ BudgetAlarms:
     budgetName: !Sub '${resourcePrefix}-budget-${CurrentAccount.AccountId}'
     budgetAmount: !GetAtt CurrentAccount.Tags.budget-alarm-threshold
     emailAddress: !GetAtt CurrentAccount.Tags.budget-alarm-threshold-email-recipient
+
+CostAndUsageReports:
+  Type: update-stacks
+  Template: cur.yaml
+  StackName: !Sub '${resourcePrefix}-monthly-cost-report'
```
zaro0508, line 19: do you think this stack is going to contain only monthly reports or other frequencys?  if you think there will be other frequencies then i suggest a more generic name like `${resourcePrefix}-cost-report`.  Otherwise if this is going to be exclusively monthly reports then keep this name and make the resource name a little more specific, like `MonthlyCostAndUsageReports`.

zaro0508 (changes requested):

org-formation/040-budgets/cur.yaml
```diff
@@ -0,0 +1,42 @@
...
+    Type: "AWS::CUR::ReportDefinition"
+    Properties:
+      ReportName: !Sub "${resourcePrefix}-monthly-costs"
+      ReportVersioning: "OVERWRITE_REPORT"
+      RefreshClosedReports: true
+      TimeUnit: "MONTHLY"
+      Compression: "Parquet"
+      Format: "Parquet"
+      S3Region: !Ref AWS::Region
+      S3Prefix: !Ref resourcePrefix
```
zaro0508, line 41: would this not set report name to `sagebase-sagebase-monthly-costs`?

zaro0508 (changes requested):

org-formation/040-budgets/_tasks.yaml
jesusaurus: Ah, good point, if we do add more reports they would be in the same stack, so the stack shouldn't refer to the report frequency.

org-formation/040-budgets/cur.yaml
jesusaurus: No, the prefix is not added to the report name, it is added to the begging of the S3 path.

org-formation/040-budgets/cur.yaml
jesusaurus: I'm on the fence about this, do we have a need to keep the information confidential?

zaro0508: to clarify, i believe this allows anybody on the internet access to the cost reports.

org-formation/040-budgets/cur.yaml
```diff
@@ -0,0 +1,54 @@
...
+              Service: "billingreports.amazonaws.com"
+          # Grant full access to the 'organizations' account
+          - Action:
+              - 's3:*'
+            Effect: Allow
+            Resource:
+              - !GetAtt CostReportBucket.Arn
+              - !Sub '${CostReportBucket.Arn}/${resourcePrefix}/*'
+            Principal:
+              AWS: "531805629419"
```
jesusaurus, line 42: I would prefer to grant access based on the `aws-admins` group, but I can't figure out how to reference group membership as a principal.

brucehoff (approved):

org-formation/040-budgets/cur.yaml
zaro0508: I think the account already has access to it's own files so I don't think we even need this section.

org-formation/040-budgets/cur.yaml
zaro0508: hmm, are you sure?  the doc says.. `S3Prefix: The prefix that Amazon Web Services adds to the report name when Amazon Web Services delivers the report. Your prefix can't include spaces.`

zaro0508 (changes requested):

org-formation/040-budgets/cur.yaml
jesusaurus: Oh, you're right. I thought it was pulling that part from `ReportName` but the two values were very similar for the manually created report.

PR #541, opened 2022-07-21
[IT-2054] Add personal account for Joni

jesusaurus:
Create account P9hFg7I5 for Joni's org-formation testing.

### DIFF
--- a/org-formation/organization.yaml
+++ b/org-formation/organization.yaml
@@ -150,6 +150,7 @@ Organization:
         - !Ref RestrictDnsRegistrationSCP
       Accounts:
         - !Ref 8dxfh53Account
+        - !Ref P9hFg7IAccount
 
   BridgeDevAccount:
     Type: OC::ORG::Account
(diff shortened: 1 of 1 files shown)


org-formation/organization.yaml
```diff
@@ -560,3 +561,15 @@ Organization:
         AccountOwner: vijay.yadav@sagebase.org
         budget-alarm-threshold: 2000
         budget-alarm-threshold-email-recipient: aws.8dxfh53@sagebase.org
+  P9hFg7I5Account:
+    Type: OC::ORG::Account
+    Properties:
+      AccountName: org-sagebase-P9hFg7I5
```
zaro0508, line 567: non-enforced account name convention is 7 chars

zaro0508 (changes requested):

org-formation/organization.yaml
jesusaurus: oof, ok

zaro0508:
actually let's just try sharing org-sagebase-itsandbox account.  You should have access to that @jesusaurus

PR #543, opened 2022-07-26
[IT-1904] Fix policies for cloudtrail

zaro0508:
This is a follow on to PR #522 which only partially refactored
the cloudtrail for encryption.  The policies in that PR were
not quite correct.

This PR will do the following to finish it:

* Enable KMS key to encrypt the cloudtrail trail
* Fix policies to allow cloudtrail to correctly access
  the S3 bucket
* Fix policies to allow cloudtrail and users to access the KMS key

### DIFF
--- a/org-formation/060-cloudtrail/_tasks.yaml
+++ b/org-formation/060-cloudtrail/_tasks.yaml
@@ -37,6 +37,7 @@ CloudTrail:
     KmsKeyAdminPrincipalArns:
       - !Sub 'arn:aws:iam::${AWS::AccountId}:root'
       - !Sub 'arn:aws:iam::${AWS::AccountId}:role/OrganizationAccountAccessRole'
+      - 'arn:aws:iam::531805629419:role/github-oidc-sage-bionetwo-ProviderRoleorganization-1EY1CRUF48VP5'
 
 CloudTrailSumoLogicRole:
   Type: update-stacks
--- a/org-formation/060-cloudtrail/cloudtrail-trail.yaml
+++ b/org-formation/060-cloudtrail/cloudtrail-trail.yaml
@@ -1,6 +1,5 @@
 # From https://github.com/org-formation/org-formation-reference/tree/master/src/templates/060-cloud-trail
 AWSTemplateFormatVersion: '2010-09-09-OC'
-
 Parameters:
   bucketName:
     Type: String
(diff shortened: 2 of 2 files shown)


org-formation/060-cloudtrail/_tasks.yaml
```diff
@@ -37,6 +37,7 @@ CloudTrail:
     KmsKeyAdminPrincipalArns:
       - !Sub 'arn:aws:iam::${AWS::AccountId}:root'
       - !Sub 'arn:aws:iam::${AWS::AccountId}:role/OrganizationAccountAccessRole'
+      - 'arn:aws:iam::531805629419:role/github-oidc-sage-bionetwo-ProviderRoleorganization-1EY1CRUF48VP5'
```
brucehoff, line 40: Why does GitHub need to access KMS?  E.g., does it transfer secrets from its own store to KMS?

zaro0508: @brucehoff because that is the role that we setup for github actions.  it needs access to the key to make changes.

brucehoff: @zaro0508 I see.  I don't see any other references to this role, in this PR or in the https://github.com/Sage-Bionetworks-IT/organizations-infra repo'.  Does GitHub Actions not need access to other AWS services besides KMS?  (Sorry if this is a dumb question.)

org-formation/060-cloudtrail/cloudtrail-trail.yaml
```diff
@@ -122,28 +121,32 @@ Resources:
...
-            Condition:
-              StringEquals:
-                s3:x-amz-acl: bucket-owner-full-control
+        - Sid: AWSCloudTrailAclCheck20150319
+          Effect: Allow
+          Principal:
+            Service: cloudtrail.amazonaws.com
+          Action: s3:GetBucketAcl
+          Resource: !GetAtt CloudTrailBucket.Arn
+          Condition:
```
xschildw, line 133: So that ensure that only the trail can do getBucketAcl (and putObject) on this bucket?

org-formation/060-cloudtrail/_tasks.yaml
zaro0508: @brucehoff the role is defined here: https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/org-formation/650-identity-providers/_tasks.yaml 

The role has admin access, however admin access does not automatically give accesss to KMS keys.  To provide KMS access we need to explicitly set every principal.  That's how it works in AWS.

org-formation/060-cloudtrail/cloudtrail-trail.yaml
zaro0508: yes, i believe so @xschildw

jesusaurus (approved):

xschildw (approved):

PR #549, opened 2022-08-10
Give Travis CI and admins read access to the CUR bucket.

jesusaurus:
The lambda for grouping costs by cost center includes a test that
will download and process the current report without uploading the
results. Give read access to travis and to 'organizations' account
admins for running the test.

PR Checklist:
[x] Describe and explain your intentions for this change
[x] Setup pre-commit and run the validators (info in README.md)

### DIFF
--- a/org-formation/040-budgets/cur.yaml
+++ b/org-formation/040-budgets/cur.yaml
@@ -32,6 +32,15 @@ Resources:
             Resource: !Sub '${CostReportBucket.Arn}/*'
             Principal:
               Service: "billingreports.amazonaws.com"
+          # Grant AWS admins and Travis CI access to view cost reports
+          - Action:
+              - 's3:GetObject'
+            Effect: Allow
+            Resource: !Sub '${CostReportBucket.Arn}/*'
+            Principal:
+              AWS:
+                - !Sub "arn:aws:iam::${AWS::AccountId}:role/OrganizationAccountAccessRole"
+                - arn:aws:iam::745159704268:role/service-accounts-CfnDeployerServiceRole-1OJK3OAN1IAI  # Travis CI
   MonthlyCostReport:
     Type: "AWS::CUR::ReportDefinition"
     Properties:


brucehoff (approved):

zaro0508 (changes requested):
This is not necessary.  It would be better to create a policy in the lambda template and give the lambda access to the cost report bucket.

jesusaurus:
@zaro0508 the lambda has already been given access, but that doesn't affect access for Travis CI while running the lambda's tests.

org-formation/040-budgets/cur.yaml
```diff
@@ -32,6 +32,15 @@ Resources:
...
               Service: "billingreports.amazonaws.com"
+          # Grant AWS admins and Travis CI access to view cost reports
+          - Action:
+              - 's3:GetObject'
+            Effect: Allow
+            Resource: !Sub '${CostReportBucket.Arn}/*'
+            Principal:
+              AWS:
+                - !Sub "arn:aws:iam::${AWS::AccountId}:role/OrganizationAccountAccessRole"
+                - arn:aws:iam::745159704268:role/service-accounts-CfnDeployerServiceRole-1OJK3OAN1IAI  # Travis CI
```
zaro0508, line 43: This is not enough because the CfnDeployerServiceRole is [scoped](https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/sceptre/admincentral/templates/service-accounts.yaml#L20) to only allow access to the [lambda bucket](https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/sceptre/admincentral/templates/iam-policies.yaml#L19-L20).  It does not have access to any other buckets currently.  That means you will also need to give that role permission to access  `${CostReportBucket.Arn}/*` as well.   This role does not provide enough permissions to run test against AWS.  that’s one of the reasons why we only created unit tests for our lambdas thus far.  i think we’de really need to think about how we would setup functional testing with lambdas and how we want to scope the security around that.

jesusaurus:
Closing this because adding a permission model for functional lambda testing is too expensive for the minimal value we get from verifying s3 object existence in testing.

PR #552, opened 2022-08-16
Add a service catalog project tag

zaro0508:
per Pranav request, add a `recover` tag to the list of projects
for the service catalog.


jesusaurus (approved):

brucehoff (changes requested):
We need to update program codes, not just projects.

brucehoff (approved):

PR #554, opened 2022-08-18
[IT-2075] Update bucket policy generator lambda

zaro0508:
Update the service catalog bucket policy generator lambda to support
non ACL based buckets.

depends on https://github.com/Sage-Bionetworks-IT/cfn-cr-sc-bucket-policy/pull/7


jesusaurus (approved):
Looks good, presuming the dependent PR is merged and tagged

PR #558, opened 2022-08-29
IT-2030: Removed unused templates, bridge peering authorizer role and ingress from old VPN CIDR

xschildw:
This PR removes a template that used to setup a route to the VPN via peered connexion.


xschildw:
WIP while I keep adding removing peering authorizer role from other accounts

xschildw:
Looks like bridge was implemented differently than other accounts (where the role is defined in essentials.yam). Propose we merge this now and do others in separate PR.

zaro0508 (approved):

PR #562, opened 2022-08-31
Pin template version

zaro0508:
Pin to a specific template version in case we need to roll back
to a working version.


sceptre/bridge/config/prod/bridge-ops.yaml
```diff
@@ -1,6 +1,6 @@
 template:
   type: http
-  url: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/master/templates/EC2/basic-ec2.j2
+  url: https://github.com/Sage-Bionetworks/aws-infra/blob/v0.4.8/templates/EC2/basic-ec2.j2
```
jesusaurus, line 3: This needs to refer to the raw content, right?

jesusaurus (approved):

PR #565, opened 2022-09-01
[IT-1530] Add missing account CostCenter tags

jesusaurus:
Set 'Platform Infrastructure' as the default CostCenter tag, and
add necessary overrides.

### DIFF
--- a/org-formation/_default_org_tags.yaml
+++ b/org-formation/_default_org_tags.yaml
@@ -1,3 +1,4 @@
 Department: Platform
 Project: Infrastructure
 CloudwatchCloudTrailLogRetentionPeriod: 90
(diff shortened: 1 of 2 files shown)


zaro0508 (commented):
LGTM, but i'll let @brucehoff confirm.

org-formation/_default_org_tags.yaml
```diff
@@ -1,3 +1,4 @@
 Department: Platform
 Project: Infrastructure
 CloudwatchCloudTrailLogRetentionPeriod: 90
+CostCenter: 'Platform Infrastructure / 990300'
```
brucehoff, line 4: Perhaps the default should be 'NO PROGRAM / 000000'.

brucehoff (approved):

PR #574, opened 2022-09-09
[IT-1046] Deploy cost reports lambda with S3 notifications

jesusaurus:
Add the cost report lambda to the budgets tasks, and trigger the lambda with S3 notifications for files written by AWS Billing.

Depends on: Sage-Bionetworks-IT/lambda-cost-reports#4

### DIFF
--- a/org-formation/040-budgets/_tasks.yaml
+++ b/org-formation/040-budgets/_tasks.yaml
@@ -13,6 +13,16 @@ BudgetAlarms:
     budgetAmount: !GetAtt CurrentAccount.Tags.budget-alarm-threshold
     emailAddress: !GetAtt CurrentAccount.Tags.budget-alarm-threshold-email-recipient
 
+CostReportLambda:
+  Type: update-stacks
+  Template: !Sub 'https://${AdminCentralCfnBucket}.s3.amazonaws.com/lambda-cost-reports/0.0.3/lambda-cost-reports.yaml'
+  StackName: !Sub '${resourcePrefix}-cost-reports-lambda'
+  DefaultOrganizationBinding:
+    Account: !Ref MasterAccount
+    Region: !Ref primaryRegion
+  Parameters:
+    bucketName: !Sub "${resourcePrefix}-cost-reports"
+
 CostAndUsageReports:
   Type: update-stacks
   Template: cur.yaml
@@ -22,3 +32,5 @@ CostAndUsageReports:
     Region: !Ref primaryRegion
   Parameters:
     resourcePrefix: !Ref resourcePrefix
+    bucketName: !Sub "${resourcePrefix}-cost-reports"
+    lambdaArn: !CopyValue [ !Sub "${resourcePrefix}-cost-reports-lambda-CostReportFunctionArn" ]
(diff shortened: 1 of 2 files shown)


org-formation/040-budgets/_tasks.yaml
```diff
@@ -22,3 +32,5 @@ CostAndUsageReports:
     Region: !Ref primaryRegion
   Parameters:
     resourcePrefix: !Ref resourcePrefix
+    bucketName: !Sub "${resourcePrefix}-cost-reports"
+    lambdaArn: !CopyValue CostReportFunctionArn
```
zaro0508, line 36: prepend `CostReportFunctionArn` with CostReportLambda StackName or/and primaryRegion.

brucehoff (approved):

org-formation/040-budgets/_tasks.yaml
```diff
@@ -13,6 +13,16 @@ BudgetAlarms:
...
 
+CostReportLambda:
+  Type: update-stacks
+  Template: !Sub 'https://${AdminCentralCfnBucket}.s3.amazonaws.com/lambda-cost-reports/0.0.3/lambda-cost-reports.yaml'
+  StackName: !Sub '${resourcePrefix}-cost-reports-lambda'
+  DefaultOrganizationBinding:
+    Account: !Ref MasterAccount
+    Region: !Ref primaryRegion
+  Parameters:
+    bucketName: !Sub "${resourcePrefix}-cost-reports"
```
zaro0508, line 24: hmm, lambda depends on the bucket and the bucket depends on the lambda in line 36.  circular dependency?

org-formation/040-budgets/cur.yaml
```diff
@@ -3,12 +3,33 @@ Description: 'Monthly Cost and Usage Report'
...
+      NotificationConfiguration:
+        LambdaConfigurations:
+          - Event: "s3:ObjectCreated:*"
+            Function: !Ref lambdaArn
+            Filter:
+              S3Key:
+                Rules:
+                  - Name: suffix
+                    Value: ".snappy.parquet"
+  LambdaPermission:
```
zaro0508, line 25: hmm, shouldn't the lambda permission go in the lambda template?

zaro0508 (changes requested):

org-formation/040-budgets/_tasks.yaml
jesusaurus: almost, by putting the event configuration on the bucket resource rather than the lambda resource, we make the bucket dependent on the lambda; and the lambda also depends on a managed policy that is dependent on the name of the bucket. but by separating the bucket from the bucket name, we can make both stacks dependent on the bucket name without having a circular dependency between the lambda, lambda policy, and bucket

org-formation/040-budgets/cur.yaml
jesusaurus: I originally put this here to couple it with the event notification on the bucket, and even though I'm tempted to move the managed policy from the lambda template into this one to consolidate the bucket references in a single template, I agree that we should put as much lambda configuration into the lambda template as possible. I'll move this.

zaro0508 (approved):

PR #578, opened 2022-09-16
[IT-2051] Always deploy the ebs lambda

jesusaurus:
Since the template URL references the template for the master branch, the template may change without the URL changing. Force deploy the lambda since a hash match doesn't mean the template is unchanged.


zaro0508 (changes requested):
I don't think this is necessary because the [template file](https://bootstrap-awss3cloudformationbucket-19qromfd235z9.s3.amazonaws.com/index.html#lambda-ebs-cleanup/master/) always references a unique version of the lambda.  when you update the lambda the CI build will deploy a unique version to the `awss3lambdaartifactsbucket` bucket and update the `CodeUri` reference

```
CodeUri: s3://essentials-awss3lambdaartifactsbucket-x29ftznj6pqw/fdf82efa2debed2f2ce4ef8ab0163934
```

PR #580, opened 2022-09-19
[IT-1766] Fix and add service control policies

zaro0508:
* Fix SCP to restrict AWS regions from being used.
* Add a policy to not allow admins of an account to leave an organization.


org-formation/_scp.yaml
```diff
@@ -6,21 +7,55 @@
...
-              - 'support:*'
-            Resource: '*'
-            Condition:
-              StringNotEquals:
-                'aws:RequestedRegion':
-                  - us-west-1
-                  - us-west-2
-                  - us-east-1
-                  - us-east-2
+        - Sid: DenyAllOutsideEU
```
xschildw, line 10: US?

org-formation/_scp.yaml
```diff
@@ -126,3 +161,19 @@
...
+  # from https://docs.aws.amazon.com/organizations/latest/userguide/orgs_manage_policies_scps_examples_general.html#example-scp-deny-region
+  RestrictLeaveOrganizationSCP:
+    Type: OC::ORG::ServiceControlPolicy
+    Properties:
+      PolicyName: RestrictLeaveOrganization
+      Description: Administrators of member accounts can't remove their accounts from the organization
+      PolicyDocument:
+        Version: '2012-10-17'
+        Statement:
+          - Sid: DenyDnsRegistration
```
xschildw, line 174: Could you confirm? That does not match the description and we need to be able to register / transfer domains (or are you planning to remove the permissions in all accounts except sageit?).

org-formation/_scp.yaml
zaro0508: good catch

org-formation/_scp.yaml
zaro0508: oops, did copy/paste without updating description

org-formation/_scp.yaml
zaro0508: fixed

org-formation/_scp.yaml
zaro0508: fixed

org-formation/_scp.yaml
```diff
@@ -126,3 +161,19 @@
...
+    Properties:
+      PolicyName: RestrictLeaveOrganization
+      Description: Administrators of member accounts can't remove their accounts from the organization
+      PolicyDocument:
+        Version: '2012-10-17'
+        Statement:
+          - Sid: DenyLeaveOrganization
+            Effect: Deny
+            Action:
+              - 'route53domains:RegisterDomain'
```
xschildw, line 177: Should this not be 'organizations:LeaveOrganization'  ?

zaro0508: yes, of course.  fixed!

zaro0508:
could you please take another look @xschildw ?

org-formation/_scp.yaml
```diff
@@ -126,3 +161,18 @@
               - 'route53domains:RegisterDomain'
               - 'route53domains:TransferDomain'
             Resource: '*'
```
xschildw, line 163: Maybe I'm missing something... There are still some domains that need to be transferred to sageIT, and we need to register domains in sageIT. Does not this rule prohibit this? Do we need a condition here to give admins (or some role) the right to do register/transfer domains?

xschildw (approved):
Looks good, see my question re: register/transfer domain.

org-formation/_scp.yaml
zaro0508: yes, your understanding of how this SCP works is correct @xschildw.  However you also need to consider how it's used.  The application of this SCP is in [organizations.yaml](https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/org-formation/organization.yaml).  We are applying this to nearly all of our accounts except the `SageITAccount`.

PR #581, opened 2022-09-21
IT-2002: Remove bastion security group from sageVPC in synapsedev account

xschildw:
This PR is trial run for using the parameter, the security group should be removed and the corresponding securityHub finding should go away.


zaro0508 (approved):
would be nice to provide a link to the actual PR for this change in this PR's description, i usually like to add a `depends on` link.  Otherwise this looks fine to me.

xschildw:
Got it, will do for the next ones...

PR #588, opened 2022-09-26
IT-2030: remove vpnsecurity group for bridge

xschildw:
This is an attempt to move forward from previous PR. The security group does not exist on prod so it cannot be updated (and cannot be rolled back either, causing the stack to be in UPDATE_ROLLBACK_FAILED). The stack is now in UPDATE_ROLLBACK_COMPLETE with 2 drifts (deleted sg and missing subscription on a SNS topic).
This PR should resolve the first drift by deleting the VpnSecurityGroup (note: it will also delete the sg from dev).
The template already has the right config for the 2nd drift, either it will update or the drift should be fixed manually
(seen similar problem with properties of Cloudfront resource).


sceptre/bridge/templates/bridge.yaml
```diff
@@ -313,10 +296,6 @@ Outputs:
         - !GetAtt AWSS3AndroidAppBucket.DomainName
     Export:
       Name: !Sub '${AWS::StackName}-AndroidAppBucketSecureURL'
-  VpnSecurityGroup:
```
zaro0508, line 316: this is referenced in https://github.com/Sage-Bionetworks/BridgeServer2-infra/blob/develop/templates/bridgeserver2-db.yaml#L47 and https://github.com/Sage-Bionetworks/BridgeServer2-infra/blob/develop/templates/bridgeserver2.yaml

zaro0508 (changes requested):

sceptre/bridge/templates/bridge.yaml
xschildw: I think these are coming from BridgeServer2-vpc stack (based on vpc.yaml which also creates a VpnSecurityGroup). This one would be referenced as bridge-VpnSecurityGroup would it not?

zaro0508: sorry, you are correct.

zaro0508 (approved):

PR #590, opened 2022-09-28
It-2147: Remove old infra VPC

xschildw:
This PR removes the infra VPC at 10.10.0.0/16 using the new 'obsolete' feature

Depends on #587


zaro0508 (approved):
LGTM.   VPCs have a lot of dependencies that make it difficult for AWS to update/remove cleanly.  I would keep an eye on this deployment (or deletion) after merging.

PR #591, opened 2022-09-29
[IT-2089] Remove obsolete patching mechanism

zaro0508:
Forgot to remove obsolete patching mechanism in bridge and scicomp as part of PR #585


zaro0508:
this is replaced by PR #592

PR #594, opened 2022-09-29
[IT-2124] Add a Cost Category for Program Codes

jesusaurus:
Replace our CUR ETL lambda with Cost Category rules. The purpose of the lambda is simply to collate our CostCenter and CostCenterOther tags, along with a fallback account mapping, into a single meaningful value. This can also be done with Cost Categories, which would allow us to use Cost Explorer for auditing/inspecting the resulting data.

The included rules were built using the online rules builder and copying the resulting JSON. Automatically updating the rules when program codes change will be challenging, but might be possible using a custom CloudFormation macro.

### DIFF
--- a/org-formation/050-costs/program-code-rules.json
+++ b/org-formation/050-costs/program-code-rules.json
@@ -1,0 +1,398 @@
+[
+  {
+    "Type": "REGULAR",
+    "Value": "990300 Platform Infrastructure",
+    "Rule": {
+      "And": [
+        {
+          "Dimensions": {
+            "Key": "LINKED_ACCOUNT",
+            "Values": [
+              "531805629419",
+              "140124849929",
+              "254940560790",
+              "231505186444",
+              "153370007719",
+              "087547153226",
+              "797640923903",
+              "867686887310",
+              "420786776710",
+              "745159704268",
+              "325565585839",
+              "383874245509",
+              "449435941126",
+              "649232250620",
+              "804034162148",
+              "743644221192"
+            ],
+            "MatchOptions": [
+              "EQUALS"
+            ]
+          }
+        },
+        {
+          "Tags": {
+            "Key": "CostCenter",
+            "MatchOptions": [
+              "ABSENT"
+            ]
+          }
+        },
+        {
+          "Tags": {
+            "Key": "CostCenterOther",
+            "MatchOptions": [
+              "ABSENT"
+            ]
+          }
+        }
+      ]
+    }
+  },
+  {
+    "Type": "REGULAR",
+    "Value": "990300 Platform Infrastructure",
+    "Rule": {
+      "Tags": {
+        "Key": "CostCenter",
+        "Values": [
+          "990300"
+        ],
+        "MatchOptions": [
+          "ENDS_WITH"
+        ]
+      }
+    }
+  },
+  {
+    "Type": "REGULAR",
+    "Value": "30144 BMGF Ki",
+    "Rule": {
+      "Tags": {
+        "Key": "CostCenter",
+        "Values": [
+          "30144"
+        ],
+        "MatchOptions": [
+          "ENDS_WITH"
+        ]
+      }
+    }
+  },
+  {
+    "Type": "REGULAR",
+    "Value": "30144 BMGF Ki",
+    "Rule": {
+      "Tags": {
+        "Key": "CostCenterOther",
+        "Values": [
+          "30144"
+        ],
+        "MatchOptions": [
+          "ENDS_WITH"
+        ]
+      }
+    }
+  },
+  {
+    "Type": "REGULAR",
+    "Value": "30144 BMGF Ki",
+    "Rule": {
+      "Dimensions": {
+        "Key": "LINKED_ACCOUNT",
+        "Values": [
+          "464102568320"
+        ],
+        "MatchOptions": [
+          "EQUALS"
+        ]
+      }
+    }
+  },
+  {
+    "Type": "REGULAR",
+    "Value": "113300 CD2H CU",
+    "Rule": {
+      "Tags": {
+        "Key": "CostCenterOther",
+        "Values": [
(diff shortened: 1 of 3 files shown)


jesusaurus:
This approach would let us remove the cost reporting lambda entirely, obsoleting https://github.com/Sage-Bionetworks-IT/lambda-cost-reports/pull/7

org-formation/050-costs/program-code-rules.json
```diff
@@ -0,0 +1,399 @@
...
+          "797640923903",
+          "867686887310",
+          "420786776710",
+          "745159704268",
+          "325565585839",
+          "383874245509",
+          "449435941126",
+          "649232250620",
+          "804034162148",
+          "743644221192"
```
zaro0508, line 26: why not list all accounts here?

if you want to be able to reference specific account IDs by using something like `!Ref SandboxAccount` then i think you can do that with [orgformation templating](https://github.com/org-formation/org-formation-cli/blob/master/docs/task-files.md#example).  if you want to generate a list of all account IDs then you can use the [Fn::EnumTargetAccounts feature](https://github.com/org-formation/org-formation-cli/blob/master/docs/cloudformation-resources.md#fnenumtargetaccountsregions).

zaro0508 (approved):

org-formation/050-costs/program-code-rules.json
jesusaurus: 1) This is a pure json file intended to be read in as a string in the template using `!ReadFile`. 2) These are the accounts that are tagged with "Platform Infrastructure", I was planning on leaving untagged resources in other accounts as uncategorized for now; but if we want categorize absolutely everything that isn't tagged as "Platform Infrastructure", then it would be better to remove this rule entirely and use "Platform Infrastructure" as a final default value.

org-formation/050-costs/program-code-rules.json
```diff
@@ -0,0 +1,399 @@
...
+    "Type": "REGULAR",
+    "Value": "BMGF Ki",
+    "Rule": {
+      "Tags": {
+    "Key": "CostCenter",
+    "Values": [
+      "BMGF-Ki"
+    ],
+    "MatchOptions": [
+      "STARTS_WITH"
```
brucehoff, line 77: Why "STARTS_WITH" instead of an exact match?

org-formation/050-costs/program-code-rules.json
```diff
@@ -0,0 +1,399 @@
...
+    ],
+    "MatchOptions": [
+      "EQUALS"
+    ]
+      }
+    }
+  },
+  {
+    "Type": "REGULAR",
+    "Value": "CD2H CU",
```
brucehoff, line 114: Shouldn't these values be the full program code strings, e.g. `CD2H-CU / 113300`?

brucehoff (changes requested):

org-formation/050-costs/program-code-rules.json
```diff
@@ -0,0 +1,399 @@
...
+      "Platform Infrastructure"
+    ],
+    "MatchOptions": [
+      "STARTS_WITH"
+    ]
+      }
+    }
+  },
+  {
+    "Type": "REGULAR",
```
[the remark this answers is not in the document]
jesusaurus: I haven't been able to find a way, I suspect because the tags are on the organization, not the account, and the two are slightly different; but I've opened a feature request for AWS to propagate the organizations tags to resources within the organization in cost explorer.

org-formation/050-costs/program-code-rules.json
jesusaurus: In order to match tags with either `/` or `_`, and in some later cases to group subprograms under a single program category (e.g. NTAP NF, or Mobile Toolbox). But now that I'm thinking about it, it might be better to match on the numeric code.

org-formation/050-costs/program-code-rules.json
jesusaurus: The value can only have numbers, letters, or whitespace. This doesn't apply to values inherited from tags, only to values set explicitly in these rules. I stripped off the numeric code when I removed the delimiter, but maybe that was a mistake.

org-formation/050-costs/program-code-rules.json
brucehoff: @jesusaurus is there an `ENDS_WITH` operator?  I'd rather match `CD2H-CU / 11330` on the unambiguous numeric code " 113300" (perhaps with a leading space) than on the human readable prefix, `CD2H-CU`, which may not be unique.

org-formation/050-costs/program-code-rules.json
```diff
@@ -0,0 +1,398 @@
...
+            "MatchOptions": [
+              "ABSENT"
+            ]
+          }
+        }
+      ]
+    }
+  },
+  {
+    "Type": "REGULAR",
```
brucehoff, line 53: What is the purpose of `CostCenter` tag matching rules for specific tag values?  Isn't this handled by the generic 'inheritance' rule at the bottom of the file?
```
  {
    "Type": "INHERITED_VALUE",
    "InheritedValue": {
      "DimensionName": "TAG",
      "DimensionKey": "CostCenter"
    }
  }
```

jesusaurus: The purpose is to group multiple tag values into a single cost category. In some cases we have trivial differences to group, like ignoring the difference between a `/` and a `_`; and in some other cases we want to combine subprojects into a single category, like grouping 301100, 301101, and 301102 into a single cost category. And in any case where we are creating a category with a regular rule rather than inheriting a value, we have to remove punctuation from the name.

Changing our tagging format to put the code at the beginning of the string and enforcing that format would make the rules quite a bit simpler because we could just use a STARTS_WITH match and match against the significant digits of the code.

brucehoff (approved):

PR #601, opened 2022-10-03
IT-2030: Update remaining VPCs

xschildw:
This PR updates the default sage VPCs in all remaining accounts (except sandbox which is getting decommisioned). Same as #598


zaro0508 (changes requested):
this is fine but can we do `sceptre/scicomp/config/prod/computevpc.yaml` and scipool prod VPCs in a separate PR?  I would like to verify that a first batch for dev VPCs works before 2nd batch to update prod VPCs

xschildw:
Removed computervpc and scipool prod VPCs from PR.

zaro0508 (approved):

PR #606, opened 2022-10-06
[IT-1532] Grant access to the Cost Explorer

jesusaurus:
Grant the viewer-plus role self-service access to the cost explorer, including creating/deleting reports and anomaly detection subscriptions.

### DIFF
--- a/org-formation/600-access/_tasks.yaml
+++ b/org-formation/600-access/_tasks.yaml
@@ -173,6 +173,29 @@ DnTDevCIServiceAccounts:
       - !Ref DnTDevAccount
     Region: us-east-1
 
+CostExplorerAccessPolicy:
+  Type: update-stacks
+  Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.3.8/templates/IAM/managed-policy.yaml
+  StackName: cost-explorer-access-policy
+  DefaultOrganizationBinding:
+    Account: !Ref MasterAccount
+    Region: !Ref primaryRegion
+  Parameters:
+    PolicyDocument:
+      Version: 2012-10-17
+      Statement:
+        - Sid: CostExplorerAccess
+          Effect: Allow
+          Action:
+            - aws-portal:ViewUsage
+            - ce:Describe*
+            - ce:Get*
+            - ce:List*
+            - ce:*AnomalyMonitor
+            - ce:*AnomalySubscription
+            - ce:*Report
+          Resource: '*'
+
 # An access role allowing Saturn Cloud access to our AWS account running their product
 # https://dev.saturncloud.io/docs/enterprise/installation/
 SaturnCloudAccessRole:
--- a/org-formation/700-aws-sso/_tasks.yaml
+++ b/org-formation/700-aws-sso/_tasks.yaml
@@ -273,6 +273,8 @@ SsoDeveloper:
     managedPolicies:
       - 'arn:aws:iam::aws:policy/PowerUserAccess'
       - 'arn:aws:iam::aws:policy/AdministratorAccess-AWSElasticBeanstalk'
+      - 'arn:aws:iam::aws:policy/AWSBillingReadOnlyAccess'
+      - !CopyValue [ !Sub '${primaryRegion}-cost-explorer-access-policy-ManagedPolicyArn' ]
     sessionDuration: 'PT12H'
     inlinePolicy: >-
       {
@@ -308,6 +310,8 @@ SsoAuditor:
       - 'arn:aws:iam::aws:policy/AmazonInspector2ReadOnlyAccess'
       - 'arn:aws:iam::aws:policy/AWSSecurityHubReadOnlyAccess'
       - 'arn:aws:iam::aws:policy/AWSElasticBeanstalkReadOnly'
+      - 'arn:aws:iam::aws:policy/AWSBillingReadOnlyAccess'
+      - !CopyValue [ !Sub '${primaryRegion}-cost-explorer-access-policy-ManagedPolicyArn' ]
     sessionDuration: 'PT1H'
     masterAccountId: !Ref MasterAccount
 
@@ -327,7 +331,10 @@ SsoViewer:
     instanceArn: !Ref instanceArn
     principalId: !Ref adminGroup
     permissionSetName: 'Viewer'
-    managedPolicies: [ 'arn:aws:iam::aws:policy/job-function/ViewOnlyAccess' ]
+    managedPolicies:
+      - 'arn:aws:iam::aws:policy/job-function/ViewOnlyAccess'
+      - 'arn:aws:iam::aws:policy/AWSBillingReadOnlyAccess'
+      - !CopyValue [ !Sub '${primaryRegion}-cost-explorer-access-policy-ManagedPolicyArn' ]
     sessionDuration: 'PT12H'
     masterAccountId: !Ref MasterAccount
 
@@ -356,6 +363,8 @@ SsoViewerPlus:
       - arn:aws:iam::aws:policy/AWSElasticBeanstalkReadOnly
       - arn:aws:iam::aws:policy/CloudWatchLogsReadOnlyAccess
       - arn:aws:iam::aws:policy/AWSLambda_ReadOnlyAccess
+      - arn:aws:iam::aws:policy/AWSBillingReadOnlyAccess
+      - !CopyValue [ !Sub '${primaryRegion}-cost-explorer-access-policy-ManagedPolicyArn' ]
 
 SsoViewerSupporter:
   Type: update-stacks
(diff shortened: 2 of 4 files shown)


brucehoff (approved):

org-formation/700-aws-sso/_tasks.yaml
```diff
@@ -356,6 +356,26 @@ SsoViewerPlus:
       - arn:aws:iam::aws:policy/AWSElasticBeanstalkReadOnly
       - arn:aws:iam::aws:policy/CloudWatchLogsReadOnlyAccess
       - arn:aws:iam::aws:policy/AWSLambda_ReadOnlyAccess
+    inlinePolicy: >-
```
zaro0508, line 359: is there not an AWS managed policy for this?

Also why not just allow viewer role this access?

jesusaurus: No, AWS doesn't provide a managed policy, the most they offer is guidance on building your own: https://docs.aws.amazon.com/awsaccountbilling/latest/aboutv2/billing-example-policies.html 

And to be honest, I'm not exactly sure what the use-cases are for viewer vs viewer-plus, but I was presuming that viewer-plus was intended for any use-case that requires more than the standard ViewOnlyAccess policy. Does it make more sense to add this inlinePolicy to both viewer and viewer-plus?

zaro0508: don't forget about the list of `managedPolicies` for this role.  when you assign a user this role they will get all that access.  This means that if your intention is to give a user in finance only `cost explorer` access you are also giving them access to all the `managedPolicies`.

if we instead put this on viewer role then they will only get access to cost explore and  `arn:aws:iam::aws:policy/job-function/ViewOnlyAccess` policy.

which should it be?  i'm not sure myself.

brucehoff: Can we simply create a new role (separate from Viewer and Viewer-Plus), meant for people who need to look at Cost Explorer?

jesusaurus: Since one of our goals is to allow any cloud user to track their own costs, I think I'm going to add the inline policy to both the viewer and viewer plus roles so that anyone with AWS access also has cost explorer access (scoped to the member account the IAM user is logged in to).

zaro0508: @brucehoff your idea is indeed possible and i think a valid alternative solution. 

@jesusaurus what about the `developers` and `auditor` role? shouldn't those roles get access to as well?  If you are going to add duplicate inline policies to multiple roles I think it would be better to create a custom managed policy that you can then add to each role's `managedPolicies` list.

org-formation/600-access/_tasks.yaml
```diff
@@ -173,6 +173,30 @@ DnTDevCIServiceAccounts:
       - !Ref DnTDevAccount
     Region: us-east-1
 
+CostExplorerAccessRole:
```
zaro0508, line 176: this is gonna create a policy, not a role.

org-formation/600-access/_tasks.yaml
```diff
@@ -173,6 +173,30 @@ DnTDevCIServiceAccounts:
...
+    Region: !Ref primaryRegion
+  Parameters:
+    PolicyDocument:
+      Version: 2012-10-17
+      Statement:
+        - Sid: CostExplorerAccess
+          Effect: Allow
+          Action:
+            - aws-portal:ViewBilling
+            - aws-portal:ViewUsage
```
zaro0508, line 191: do you think we can replace these with the AWS managed policy `AWSBillingReadOnlyAccess`?

org-formation/700-aws-sso/_tasks.yaml
```diff
@@ -273,6 +273,7 @@ SsoDeveloper:
     managedPolicies:
       - 'arn:aws:iam::aws:policy/PowerUserAccess'
       - 'arn:aws:iam::aws:policy/AdministratorAccess-AWSElasticBeanstalk'
+      - !Sub "arn:aws:iam:${MasterAccount}:policy/CostExplorerAccess"
```
zaro0508, line 276: Probably better to just use a [CopyValue](https://github.com/org-formation/org-formation-cli/blob/master/docs/task-files.md#copyvalue) and reference the [managed-policy template output](https://github.com/Sage-Bionetworks/aws-infra/blob/master/templates/IAM/managed-policy.yaml#L30)

org-formation/600-access/_tasks.yaml
```diff
@@ -173,6 +173,30 @@ DnTDevCIServiceAccounts:
       - !Ref DnTDevAccount
     Region: us-east-1
 
+CostExplorerAccessRole:
+  Type: update-stacks
+  Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.3.8/templates/IAM/managed-policy.yaml
+  StackName: cost-explorer-access-role
```
zaro0508, line 179: policy

zaro0508 (changes requested):

jesusaurus:
@zaro0508 the way I have this now it's only going to apply to the organizations accounts, right? How do I also give access to the strides accounts?

org-formation/600-access/_tasks.yaml
jesusaurus: according to https://aws.permissions.cloud/managedpolicies/AWSBillingReadOnlyAccess it only grants `aws-portal:ViewBilling`, so I don't think it's very useful; but I'm also surprised that that isn't mentioned anywhere under https://docs.aws.amazon.com/awsaccountbilling/latest/aboutv2/security-iam.html

zaro0508 (changes requested):

zaro0508 (changes requested):
Also the `strides-ampad-workflows` account is another one of those accounts that is not under our umbrella organization.

sceptre/strides/templates/jumpcloud-idp.yaml
```diff
@@ -83,6 +83,24 @@ Resources:
               - cur:DeleteReportDefinition
               - cur:ModifyReportDefinition
             Resource: "*"
+  AWSIAMCostExplorerAccessPolicy:
```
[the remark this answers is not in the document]
jesusaurus: whoops, yeah it should, and the auditor role too

sceptre/strides/templates/jumpcloud-idp.yaml
```diff
@@ -83,6 +84,24 @@ Resources:
...
+    Type: "AWS::IAM::ManagedPolicy"
+    Properties:
+      PolicyDocument:
+        Version: "2012-10-17"
+        Statement:
+          -
+            Effect: "Allow"
+            Action:
+              - aws-portal:ViewUsage
```
zaro0508, line 96: replace with AWSBillingReadOnlyAccess?

zaro0508 (changes requested):
i think we want to do the same for  `sceptre/strides-ampad-workflows`?

sceptre/strides/templates/jumpcloud-idp.yaml
jesusaurus: done

zaro0508:
@jesusaurus i've mentioned `strides-ampad-workflows` a few times in this PR.  do you plan to do anything  about it?

jesusaurus:
@zaro0508 whoops, not sure how I missed adding that

zaro0508 (changes requested):

org-formation/700-aws-sso/_tasks.yaml
```diff
@@ -273,6 +273,8 @@ SsoDeveloper:
     managedPolicies:
       - 'arn:aws:iam::aws:policy/PowerUserAccess'
       - 'arn:aws:iam::aws:policy/AdministratorAccess-AWSElasticBeanstalk'
+      - 'arn:aws:iam::aws:policy/AWSBillingReadOnlyAccess'
+      - !CopyValue [ !Sub '${primaryRegion}-cost-explorer-access-policy-ManagedPolicyArn' ]
```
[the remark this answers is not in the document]
jesusaurus: I'm not exactly sure why its needed, but something about stringing `!CopyValue !Sub` together causes the linter to fail with `syntax error: expected <block end>, but found '<tag>' (syntax)`. I copied this format from [a few lines above](https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/org-formation/700-aws-sso/_tasks.yaml#L256) it.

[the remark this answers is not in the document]
zaro0508: ok. sounds good then

zaro0508 (approved):

PR #607, opened 2022-10-07
[IT-2089] narrow resource group for patching

zaro0508:
We apply tags to EC2 to identify which EC2s need patching.  The problem
is that the tags are also propogated to the EC2s attached volumes.
This means that the patch group query will return EBS volumes for
patching which will fail.  We narrow the scope of the query so that
only instances are returned to the patch manager to run patch update on.


brucehoff (approved):
Great sleuthing, Khai!

PR #614, opened 2022-10-15
Fix JSON syntax (again)

jesusaurus:
Add braces needed to specify we have a list of objects.

### DIFF
--- a/org-formation/600-access/_tasks.yaml
+++ b/org-formation/600-access/_tasks.yaml
@@ -184,6 +184,7 @@ CostExplorerAccessPolicy:
     PolicyDocument: '{
       "Version": "2012-10-17",
       "Statement": [
+        {
           "Sid": "CostExplorerAccess",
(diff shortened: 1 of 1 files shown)


org-formation/600-access/_tasks.yaml
```diff
@@ -184,6 +184,7 @@ CostExplorerAccessPolicy:
     PolicyDocument: '{
```
zaro0508, line 184: try using `PolicyDocument: >-`, like how it's done for [inlinePolicy in aws configs](https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/org-formation/700-aws-sso/_tasks.yaml#L811)

jesusaurus: @zaro0508 I was able to successfully test the JSON string as it is here with a manual sceptre deploy in itsandbox.

zaro0508 (approved):

PR #615, opened 2022-10-17
[IT-2139] Setup lifecycle policy on logs data

zaro0508:
Enable a storage lifecycle policy on cloudtrail logs. We setup to move data to glacier after 90 days then move to deep archive after 180 days then delete after 360 days


xschildw (approved):
Seems reasonable to me, did you check with Brad?

zaro0508:
@xschildw Brad's recommendation is in the jira issue.

PR #616, opened 2022-10-19
[IT-2268] Fix managed policies

jesusaurus:
We need to use different parameters for managed policies managed by AWS vs managed by us. Update to use a template with support for the parameter used by custom managed policies.

depends-on: https://github.com/Sage-Bionetworks/aws-infra/pull/360

### DIFF
--- a/org-formation/600-access/_tasks.yaml
+++ b/org-formation/600-access/_tasks.yaml
@@ -177,29 +177,33 @@ CostExplorerAccessPolicy:
   Type: update-stacks
   Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.3.8/templates/IAM/managed-policy.yaml
   StackName: cost-explorer-access-policy
+  DefaultOrganizationBindingRegion: !Ref primaryRegion
   DefaultOrganizationBinding:
-    Account: !Ref MasterAccount
-    Region: !Ref primaryRegion
+    IncludeMasterAccount: true
+  OrganizationBindings:
+    TargetBinding:
+      Account: '*'
   Parameters:
-    PolicyDocument: '{
-      "Version": "2012-10-17",
-      "Statement": [
-        {
-          "Sid": "CostExplorerAccess",
-          "Effect": "Allow",
-          "Action": [
-              "aws-portal:ViewUsage",
-              "ce:Describe*",
-              "ce:Get*",
-              "ce:List*",
-              "ce:*AnomalyMonitor",
-              "ce:*AnomalySubscription",
-              "ce:*Report"
-          ],
-          "Resource": "*"
-        }
-      ]
-    }'
+    PolicyDocument: >-
+      {
+        "Version": "2012-10-17",
+        "Statement": [
+          {
+            "Effect": "Allow",
+            "Action": [
+                "aws-portal:ViewUsage",
+                "ce:Describe*",
+                "ce:Get*",
+                "ce:List*",
+                "ce:*AnomalyMonitor",
+                "ce:*AnomalySubscription",
+                "ce:*Report"
+            ],
+            "Resource": "*"
+          }
+        ]
+      }
+    PolicyName: !Ref CostExplorerPolicyName
 
 # An access role allowing Saturn Cloud access to our AWS account running their product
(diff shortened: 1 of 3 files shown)


xschildw (approved):

org-formation/600-access/_tasks.yaml
```diff
@@ -181,25 +181,27 @@ CostExplorerAccessPolicy:
...
-        }
-      ]
-    }'
+    PolicyDocument:
+      !Sub
+        '{
+          "Version": "2012-10-17",
+          "Statement": [
+            {
+              "Sid": "${policyCostExplorerSid}",
```
zaro0508, line 190: where is `policyCostExplorerSid` defined for this reference?

org-formation/700-aws-sso/_tasks.yaml
```diff
@@ -278,7 +278,7 @@ SsoDeveloper:
       - 'arn:aws:iam::aws:policy/PowerUserAccess'
       - 'arn:aws:iam::aws:policy/AdministratorAccess-AWSElasticBeanstalk'
       - 'arn:aws:iam::aws:policy/AWSBillingReadOnlyAccess'
-      - !CopyValue [ !Sub '${primaryRegion}-cost-explorer-access-policy-ManagedPolicyArn' ]
+    customerManagedPolicyName: !Ref policyCostExplorerSid
```
zaro0508, line 281: i think the reference needs to be the policy name?  I don't think a policy's Sid is the same as a policy's name.  You probably need to output the policy name in the template that created the policy then reference it here.

zaro0508 (changes requested):

org-formation/600-access/_tasks.yaml
jesusaurus: I'm adding it to `_parameters.yaml`: https://github.com/Sage-Bionetworks-IT/organizations-infra/pull/616/files#diff-69957a1b2df5fa15bc2b9dc2bb4c2321375dcf5966a15b43600657e1c34c7f12R24

org-formation/700-aws-sso/_tasks.yaml
jesusaurus: Oh, hrm, I was interpreting that as the name that AWS uses, not the name that CloudFormation uses. I wonder which it is

org-formation/600-access/_tasks.yaml
zaro0508: ahh, ok.  have you tested this with `org-formation print-task`?

org-formation/700-aws-sso/_tasks.yaml
zaro0508: I believe the Sid is just a description of the policy

jesusaurus: After some testing I found that we need to set the ManagedPolicyName parameter in the template and use that value. I've updated the dependent PR to also add support for that parameter as well.

org-formation/600-access/_tasks.yaml
jesusaurus: yeah, `npm run print-tasks-failfast` is happy

zaro0508 (changes requested):

org-formation/_parameters.yaml
```diff
@@ -20,3 +20,7 @@ allRegions:
 AdminCentralCfnBucket:
   Type: String
   Default: bootstrap-awss3cloudformationbucket-19qromfd235z9
+
+CostExplorerPolicyName:
+  Type: String
+  Default: CostExplorerAccess
```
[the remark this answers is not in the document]
jesusaurus: yeah, I can do that, that name is much less likely to be duplicated

org-formation/600-access/_tasks.yaml
```diff
@@ -181,25 +181,26 @@ CostExplorerAccessPolicy:
     Account: !Ref MasterAccount
```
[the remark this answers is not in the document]
jesusaurus: whoops, yeah it does

zaro0508 (approved):

PR #622, opened 2022-10-26
[IT-2214] Deploy cost category rule generator macro

jesusaurus:
Deploy our macro for generating cost category rules into the master account and use it to build our cost category rules. This is deployed to the master account since that is the only account we can manage cost categories from.


org-formation/_tasks.yaml
```diff
@@ -11,9 +11,14 @@ Types:
   DependsOn: [ OrganizationUpdate ]
   Path: ./005-types/_tasks.yaml
 
+Macros:
```
zaro0508, line 14: do we really need a folder just for macros?  this lambda seems like it's specifically related to `050-costs` so why not just deploy it from there?

org-formation/050-costs/categories.yaml
```diff
@@ -6,4 +6,113 @@ Resources:
...
+              - "797640923903"
+              - "867686887310"
+              - "420786776710"
+              - "745159704268"
+              - "325565585839"
+              - "383874245509"
+              - "449435941126"
+              - "649232250620"
+              - "804034162148"
+              - "743644221192"
```
zaro0508, line 30: could we maybe add a comment to this file to tell users that they can find account ID to name mapping 
 at `https://d-906769aa66.awsapps.com/start#/`?

zaro0508 (changes requested):

org-formation/050-costs/categories.yaml
jesusaurus: done

org-formation/_tasks.yaml
jesusaurus: I'm trying to treat the macros similar to the types since the stack needs to complete before the macro is available. We might not ever create a second macro, but I think it still deserves its own stage so that we can manage the dependency at the top level in this file.

zaro0508: i think that would make sense if the lambda was more general purpose however this lambda is specific to cost so I would prefer to keep them in the same folder so that it's in context.  You can setup dependencies between `update-stack` deployments by using the orgf's [DependOn](https://github.com/org-formation/org-formation-cli/blob/master/docs/task-files.md#update-stacks) parameter.

jesusaurus: done

org-formation/050-costs/categories.yaml
```diff
@@ -6,4 +6,114 @@ Resources:
     Properties:
       Name: 'Program Code'
       RuleVersion: 'CostCategoryExpression.v1'
-      Rules: !ReadFile program-code-rules.json
+      Rules:
+        Fn::Transform:
+          - Name: SageCostRuleGenerator
+        # Account names can be found at https://d-906769aa66.awsapps.com/start#/
+        RegularValues:
```
zaro0508, line 13: I assume AWS CFN will trigger `SageCostRuleGenerator` macro when changes are made to the below configs?  I would definitely test to make sure that happens.

zaro0508 (approved):

org-formation/050-costs/categories.yaml
jesusaurus: I'm not sure I understand the question, but CFN will run the macro every time it processes the template whether or not it has changed, though https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/template-macros.html doesn't really specify when that processing happens

PR #623, opened 2022-10-26
[IT-2221] Restrict ssh access to scipool prod

jesusaurus:
Remove bastian ssh access from `internalpoolvpc` and `scheduledjobsvpc`, ssh access is still available from the vpn.

depends-on: https://github.com/Sage-Bionetworks/aws-infra/pull/361

### DIFF
--- a/sceptre/scipool/config/prod/internalpoolvpc.yaml
+++ b/sceptre/scipool/config/prod/internalpoolvpc.yaml
@@ -1,9 +1,10 @@
 template:
   type: http
-  url: https://{{stack_group_config.admincentral_cf_bucket}}.s3.amazonaws.com/aws-infra/v0.4.10/vpc.yaml
+  url: https://{{stack_group_config.admincentral_cf_bucket}}.s3.amazonaws.com/aws-infra/v0.5.2/vpc.yaml
 stack_name: internalpoolvpc
 dependencies:
   - prod/bootstrap.yaml
 parameters:
   VpcSubnetPrefix: "10.41"
   VpcName: internalpoolvpc
+  IncludeBastianSecurityGroup: "false"
--- a/sceptre/scipool/config/prod/scheduledjobsvpc.yaml
+++ b/sceptre/scipool/config/prod/scheduledjobsvpc.yaml
@@ -1,6 +1,6 @@
 template:
   type: http
-  url: https://{{stack_group_config.admincentral_cf_bucket}}.s3.amazonaws.com/aws-infra/v0.4.0/VPC/vpc-mini.yaml
+  url: https://{{stack_group_config.admincentral_cf_bucket}}.s3.amazonaws.com/aws-infra/v0.5.2/VPC/vpc-mini.yaml
 stack_name: scheduledjobsvpc
 stack_tags:
   Department: "Platform"
@@ -11,3 +11,4 @@ dependencies:
 parameters:
   VpcSubnetPrefix: "10.255.40"
   VpcName: "scheduledjobsvpc"
+  IncludeBastianSecurityGroup: "false"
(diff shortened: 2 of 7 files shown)


sceptre/scipool/config/prod/internalpoolvpc.yaml
```diff
@@ -7,3 +7,4 @@ dependencies:
 parameters:
   VpcSubnetPrefix: "10.41"
   VpcName: internalpoolvpc
+  IncludeBastianSecurityGroup: "false"
```
zaro0508, line 10: i think we need to keep this because we apply bastian SG to windows instances to allow users to RDP to them after connecting to the VPN.

zaro0508 (changes requested):

sceptre/scipool/config/prod/scheduledjobsvpc.yaml
```diff
@@ -11,3 +11,4 @@ dependencies:
 parameters:
   VpcSubnetPrefix: "10.255.40"
   VpcName: "scheduledjobsvpc"
+  IncludeBastianSecurityGroup: "false"
```
zaro0508, line 14: Bastian SG should be removed for all `scheduledjobsvpc` so could you please make this change to all files under `sceptre/scipool/config` folder ?

sceptre/scipool/config/prod/internalpoolvpc.yaml
xschildw: Is this not taken care of by the instance being associated with Vpn security group?

zaro0508: yes, you are correct @xschildw .  i confused bastian with vpn SG.

then we'll want to make this change for all VPC files under `sceptre/scipool/config` folder.

zaro0508 (changes requested):

sceptre/scipool/config/prod/internalpoolvpc.yaml
jesusaurus: @zaro0508 the bastian security group is specifically for ssh access from 0.0.0.0: https://github.com/Sage-Bionetworks/aws-infra/blob/master/templates/vpc.yaml#L506-L511, I think you're confusing it with the quarantine security group.

@xschildw do you mean the quarantine security group? they are both being applied, and I'm not sure about the order the security groups are processed in, but even if the quarantine group is applied first, a failure would fall back on the bastian security group which would allow the access.

xschildw: @jesusaurus now that you mention it, the quarantine SGs need to be changed (they're currently referring to 10.1.0.0/16). My understanding was that they should remove all access (i.e. to be used to isolate instances when compromised). Where did you see them being applied (I checked a few instances but no luck...)?

jesusaurus: @xschildw looking at internalpoolvpc in scipoolprod, we have sg-0bb347378d543521d allowing ssh from the vpn and we also have sg-0ac57202ffc20b43b allowing ssh from 0.0.0.0, is it possible to create a security group that isn't applied? I thought they were automatically applied across the vpc they are associated with

jesusaurus: @xschildw after doing a bunch of digging, I think you're right, these security groups exist but aren't applied anywhere. but the security group is still a CIS finding and removing it seems like an easy and secure solution.

zaro0508: @jesusaurus I would recommend removing it.

PR #643, opened 2022-11-18
IT-1756: enable SecurityHub findings suppression - fix deployment account

xschildw:
The infra (queue + policies) should be deployed to the securitycentral account. This PR should fix the deployment account.


zaro0508 (approved):
please check that both accounts are updated after the deployment.

PR #646, opened 2022-11-22
IT-1756: Add exceptions for CIS4.1/2

xschildw:
This PR adds a rule to send some findings for CIS 4.1 and 4.2 to the suppression queue (starting with VPN SGs in accounts in security scope).

### DIFF
--- a/org-formation/075-security-hub/security-hub-suppress-infra.yaml
+++ b/org-formation/075-security-hub/security-hub-suppress-infra.yaml
@@ -265,6 +265,47 @@ Resources:
           - Arn
         Id: Target0
 
+  # In every account, suppress findings for Bastion SGs
+  # CIS4.1: Ensure no security groups allow ingress from 0.0.0.0/0 to port 22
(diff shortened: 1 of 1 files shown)


org-formation/075-security-hub/security-hub-suppress-infra.yaml
```diff
@@ -265,6 +265,47 @@ Resources:
...
+            "Resources": {
+              "Id": [{
+                # scicomp - snowflakevpc
+                "equals": "arn:aws:ec2:us-east-1:055273631518:security-group/sg-017d8398957a0d199"
+              }, {
+                # scicomp - computevpc
+                "equals": "arn:aws:ec2:us-east-1:055273631518:security-group/sg-0fb7cd3ecd8c0a304"
+              }, {
+                # transit - vpn endpoint
+                "equals": "arn:aws:ec2:us-east-1:153370007719:security-group/sg-0642b65f82adbca1a"
```
zaro0508, line 294: maybe we should just remove the SG from transit account?  we shouldn't be running any instances in there anyways.

zaro0508 (approved):
please consider suggestion but in general it looks good

PR #647, opened 2022-11-30
[IT-2248] Align Security Group with Routing Table.

jesusaurus:
The security group is applied to the VPC that VPN clients land on after authenticating. This VPC does not need to allow inbound connections since the only hosts on the subnet are VPN clients.

The routing table only has a configuration for local traffic, restrict the security group similarly.


zaro0508 (approved):
seems fine to me.  please test after merging.

PR #648, opened 2022-12-06
IT-1756: Update infra to support Security Hub findings suppression

xschildw:
This PR creates the lambda that takes a finding from the queue created by #643 and does the actual suppression.

Depends on #643.


org-formation/075-security-hub/security-hub-suppress-lambda.yaml
```diff
@@ -0,0 +1,119 @@
+---
+Transform: AWS::Serverless-2016-10-31
+Resources:
+  AWSLambdaPowertoolsApplication:
+    Type: AWS::Serverless::Application
+    Properties:
+      Location:
+        ApplicationId: arn:aws:serverlessrepo:eu-west-1:057560766410:applications/aws-lambda-powertools-python-layer
+        SemanticVersion: 1.17.0
```
jesusaurus, line 9: Do we need to use this version? It looks like the current version is 2.4.0: https://awslabs.github.io/aws-lambda-powertools-python/latest/

org-formation/075-security-hub/security-hub-suppress-lambda.yaml
```diff
@@ -0,0 +1,119 @@
...
+                  logger.exception("client error")
+                  raise ConnectionError(f"Client error invoking batch update findings {error}")
+              except boto3exceptions.ParamValidationError as error:
+                  logger.exception("invalid parameters")
+                  raise ValueError(f"The parameters you provided are incorrect: {error}")
+
+              return {"statusCode": 200}
+      Role:
+        Fn::ImportValue:
+          !Sub 'sagebase-securityhub-suppression-infra-SecurityHubBatchUpdateSuppressionRoleArn'
```
jesusaurus, line 92: It doesn't look like we're performing any substitutions, can we simplify this to `!ImportValue 'sagebase-securityhub-suppression-infra-SecurityHubBatchUpdateSuppressionRoleArn'`?

org-formation/075-security-hub/security-hub-suppress-lambda.yaml
```diff
@@ -0,0 +1,119 @@
...
+  BatchUpdateLambdaFctEventSourceEventSourceMapping:
+    Type: AWS::Lambda::EventSourceMapping
+    Properties:
+      FunctionName:
+        Ref: BatchupdateLambdaFct
+      BatchSize: 10
+      Enabled: true
+      EventSourceArn:
+        Fn::ImportValue:
+          !Sub 'sagebase-securityhub-suppression-infra-SecurityHubFindingsQueueArn'
```
jesusaurus, line 118: Can we remove this `!Sub` too?

org-formation/075-security-hub/security-hub-suppress-lambda.yaml
```diff
@@ -0,0 +1,119 @@
...
+    Properties:
+      Location:
+        ApplicationId: arn:aws:serverlessrepo:eu-west-1:057560766410:applications/aws-lambda-powertools-python-layer
+        SemanticVersion: 1.17.0
+    Metadata:
+      aws:cdk:path: sechub-finding-suppression/AWSLambdaPowertoolsApplication
+  BatchupdateLambdaFct:
+    Type: AWS::Lambda::Function
+    Properties:
+      Code:
```
jesusaurus, line 15: I'm a little wary of adding python directly to this project instead of creating a separate lambda repo, but I won't block on that since this function is so small.

org-formation/075-security-hub/security-hub-suppress-lambda.yaml
```diff
@@ -0,0 +1,119 @@
+---
```
zaro0508, line 1: did you create this lambda?  if not could you please add a comment telling us where you got it from?

org-formation/075-security-hub/security-hub-suppress-lambda.yaml
xschildw: Updated to 2.5.0

org-formation/075-security-hub/security-hub-suppress-lambda.yaml
xschildw: Done

org-formation/075-security-hub/security-hub-suppress-lambda.yaml
xschildw: Added source in desc.

jesusaurus (approved):

org-formation/075-security-hub/security-hub-suppress-lambda.yaml
```diff
@@ -0,0 +1,121 @@
...
+              except boto3exceptions.ClientError as error:
+                  logger.exception("client error")
+                  raise ConnectionError(f"Client error invoking batch update findings {error}")
+              except boto3exceptions.ParamValidationError as error:
+                  logger.exception("invalid parameters")
+                  raise ValueError(f"The parameters you provided are incorrect: {error}")
+
+              return {"statusCode": 200}
+      Role:
+        !ImportValue 'sagebase-securityhub-suppression-infra-SecurityHubBatchUpdateSuppressionRoleArn'
```
zaro0508, line 95: this depends on `security-hub-suppress-infra.yaml` and `security-hub-suppress-infra.yaml` depends on this lambda which means that you have a circular dependency.  In order to fix this situation you may need to create a  role for this in here.

org-formation/075-security-hub/security-hub-suppress-lambda.yaml
```diff
@@ -0,0 +1,121 @@
...
+      MaximumRetryAttempts: 2
+  BatchUpdateLambdaFctEventSourceEventSourceMapping:
+    Type: AWS::Lambda::EventSourceMapping
+    Properties:
+      FunctionName:
+        Ref: BatchupdateLambdaFct
+      BatchSize: 10
+      Enabled: true
+      EventSourceArn:
+        Fn::ImportValue: 'sagebase-securityhub-suppression-infra-SecurityHubFindingsQueueArn'
```
zaro0508, line 120: circular dependency.

zaro0508 (changes requested):

zaro0508:
I'm wondering if we should be suppressing findings in inspector and guardduty instead of securityhub?  My thinking is that if we suppress in guardduty and inspector the findings won't make it to security hub?  Setting up suppression rules in guardduty and inspector does seem possible..
https://docs.aws.amazon.com/inspector/latest/user/findings-managing-supression-rules.html
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-inspectorv2-filter.html#cfn-inspectorv2-filter-filteraction  and set FilterAction to SUPPRESS as indicated by boto3 https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/inspector2.html#Inspector2.Client.create_filter

org-formation/075-security-hub/security-hub-suppress-infra.yaml
```diff
@@ -249,7 +249,10 @@ Resources:
           findings:
             GeneratorId:
               - 'arn:aws:securityhub:::ruleset/cis-aws-foundations-benchmark/v/1.2.0/rule/4.3'
+              - 'aws-foundational-security-best-practices/v/1.0.0/EC2.2'
```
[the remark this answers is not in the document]
xschildw: I see what you mean, I'll change the description to something like 'this rule handles suppression of controls (GeneratorIds) for all accounts'.

org-formation/075-security-hub/security-hub-suppress-infra.yaml
```diff
@@ -9,18 +9,15 @@ Resources:
...
-          Resource:
-            Fn::Join:
-            - ''
-            - - 'arn:aws:logs:'
-              - Ref: AWS::Region
-              - ":"
-              - Ref: AWS::AccountId
-              - ":log-group:/aws/lambda/security-hub-batch-update-suppression:*"
-          Sid: CloudWatchLogs
+          Resource: !Sub "arn:aws:logs:${AWS::Region}:${AWS::AccountId}:log-group:/aws/lambda/securityhub-suppress-findings:*"
```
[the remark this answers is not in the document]
xschildw: I don't think it's really a dependency, I'm just specifying the name of the CWLogs that the lambda can write to. I tried to make it use an output (and then ran into the circular dependency), but there was no way to get the name of the lambda as an output anyway (it was either the arn or the resource name, not the function name) so I reverted to this.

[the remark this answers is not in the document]
xschildw: Re: 1 vs 2 PRs you might be right, I was hoping the DependsOn would ensure the infra.yaml would go before the lambda.yaml.

org-formation/075-security-hub/security-hub-suppress-infra.yaml
```diff
@@ -276,3 +278,8 @@ Outputs:
     Description: Arn of the findings queue
     Export:
       Name: !Sub '${AWS::StackName}-SecurityHubFindingsQueueArn'
+  SecurityHubFindingsSuppressionLambdaFctName:
```
zaro0508, line 281: you are taking a passed in value and then just outputting it?  I'm not sure why you need to do this.  Is there some purpose for this output that i'm not aware of?

zaro0508 (changes requested):
Looking at this again I've realized that it would make more sense to put all aws resources in the lambda (with the event rules) and deploy the lambda as a complete package.  You had this statement as the lambda description..

>   This lambda takes the findings from the findings queue defined in 'security-hub-suppress-infra.yaml' and does the actual suppression. See https://aws.amazon.com/blogs/security/how-to-create-auto-suppression-rules-in-aws-security-hub/   for more details.

in that blog there is a reference.. 

> You can find the latest code [on GitHub](https://github.com/aws-samples/aws-security-hub-automatic-suppression-rules), where you can also contribute to the sample code

The sample code is the lambda (in CDK) which deploys a security hub suppression solution as an entire lambda package.  Why didn't you just use that solution?

org-formation/075-security-hub/security-hub-suppress-infra.yaml
xschildw: I used to have everything in the same CF stack (like the CDK), but I (over?)thought it'd be better to split the lambda out in case we need to add some logic to it (in which case we'd also move it to SAM)... The param/output is just to pass the lambda function name, which is needed in the policy (https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/4c07059b53167d7ef54132dfd0e91d18e4290693/org-formation/075-security-hub/security-hub-suppress-infra.yaml#L23) but not accessible as an output (output from lambda is either arn or resource name, not function name, besides that'd be a circular dependency as you mentioned in other PR). I can put it all back in one stack...

org-formation/075-security-hub/security-hub-suppress-infra.yaml
```diff
@@ -1,6 +1,6 @@
-AWSTemplateFormatVersion: 2010-09-09
+Transform: AWS::Serverless-2016-10-31
 Description: >-
-  Provision queues and policies to support automatic suppression of Security Hub findings
+  Provision queues, policies and a lambda to support automatic suppression of Security Hub findings
 Resources:
   SecurityHubBatchUpdateSuppressionManagedPolicy:
     Type: AWS::IAM::ManagedPolicy
```
xschildw, line 6: Add a depends on Lambda and use a ref in the arn of the log below.

org-formation/075-security-hub/security-hub-suppress-infra.yaml
```diff
@@ -9,18 +9,15 @@ Resources:
...
-          Resource:
-            Fn::Join:
-            - ''
-            - - 'arn:aws:logs:'
-              - Ref: AWS::Region
-              - ":"
-              - Ref: AWS::AccountId
-              - ":log-group:/aws/lambda/security-hub-batch-update-suppression:*"
-          Sid: CloudWatchLogs
+          Resource: !Sub "arn:aws:logs:${AWS::Region}:${AWS::AccountId}:log-group:/aws/lambda/suppress-security-hub-findings:*"
```
xschildw, line 19: Double-check how it was done in original template.

xschildw: It was hardcoded as here

org-formation/075-security-hub/security-hub-suppress-infra.yaml
```diff
@@ -234,22 +231,134 @@ Resources:
...
+      Layers:
+      - Fn::GetAtt:
+        - AWSLambdaPowertoolsApplication
+        - Outputs.LayerVersionArn
+      Runtime: python3.8
+      Timeout: 120
+  BatchupdateEventInvokeConfig:
+    Type: AWS::Lambda::EventInvokeConfig
+    Properties:
+      FunctionName: suppress-security-hub-findings
```
zaro0508, line 333: less places with hard coded names would be better therefore i would consider doing a `!Ref BatchupdateLambdaFct` here

org-formation/075-security-hub/security-hub-suppress-infra.yaml
```diff
@@ -234,22 +231,134 @@ Resources:
...
+      Qualifier: "$LATEST"
+      MaximumRetryAttempts: 2
+  BatchUpdateLambdaFctEventSourceEventSourceMapping:
+    Type: AWS::Lambda::EventSourceMapping
+    Properties:
+      FunctionName: suppress-security-hub-findings
```
zaro0508, line 339: less places with hard coded names would be better therefore i would consider doing a `!Ref BatchupdateLambdaFct` here

zaro0508 (changes requested):

org-formation/075-security-hub/security-hub-suppress-infra.yaml
```diff
@@ -76,7 +73,7 @@ Resources:
         - Action: sts:AssumeRole
           Effect: Allow
           Principal:
-            AWS: 'arn:aws:sts::140124849929:assumed-role/AWSReservedSSO_Administrator_1de765cf1298ed53/x.schildwachter@sagebase.org'
+            AWS: 'arn:aws:sts::140124849929:assumed-role/AWSReservedSSO_Administrator_1de765cf1298ed53/*'
```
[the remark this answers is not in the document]
xschildw: Do you mean pass in a list of ARNs?

org-formation/075-security-hub/security-hub-suppress-infra.yaml
```diff
@@ -1,6 +1,6 @@
-AWSTemplateFormatVersion: 2010-09-09
+Transform: AWS::Serverless-2016-10-31
 Description: >-
-  Provision queues and policies to support automatic suppression of Security Hub findings
+  Provision queues, policies and a lambda to support automatic suppression of Security Hub findings
```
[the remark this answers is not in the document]
xschildw: Done

org-formation/075-security-hub/security-hub-suppress-infra.yaml
xschildw: If I understand correctly, !Ref BatchupdateLambdaFct will return the resource name (not the function name). To make this work I can name the function the same as the resource (at line 323), then use use a Ref to the resource everywhere else. Not very elegant but should work. I've seen examples in the doc where they use !Ref for the FunctionName here, I don't understand how it works based on the doc for !Ref and AWS::Lambda::Function resource...

org-formation/075-security-hub/security-hub-suppress-infra.yaml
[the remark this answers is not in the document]
zaro0508: yes, exactly

org-formation/075-security-hub/security-hub-suppress-infra.yaml
zaro0508: i thought the function name is the same as the resource name?  if not then what would the function name look like compared to the resource name?

xschildw: When I created the lambda without a function name it generated a name starting with the resource name but ending with some (semi-)random string (same as it does for other resources where you don't specify the name). In this case it's an issue because I can't get the name as an attribute of the resource (from the doc at least), and I need it to put in the policy for the lambda. Without something like 'myFct.FctName', I don't see how I can do it without hardcoding the function name (and making it the same as the resource name is just a hack so I can use !Ref)...

zaro0508 (approved):

jesusaurus (approved):

PR #651, opened 2022-12-08
IT-2398 adding Genie Prod account

brucehoff:
Create the production account for hosting the Genie-BPC Shiny application.


zaro0508:
Is there an jira issue that we can tie to this PR?  Also do you think we should address the [issues regarding genie-bpc-infra repo](https://sagebionetworks.slack.com/archives/CEFQD0KU1/p1670454386293199) before going forward?

brucehoff:
> Is there an jira issue that we can tie to this PR? 
Yes: The PR comes from a branch titled with the Jira issue ID, IT-2398.  I added the ID to the PR title too.

> Also do you think we should address the [issues regarding genie-bpc-infra repo](https://sagebionetworks.slack.com/archives/CEFQD0KU1/p1670454386293199) before going forward?

Yes, we can address your comments before deploying the app'. But I'd like to proceed to set up the account in parallel.

org-formation/organization.yaml
```diff
@@ -563,6 +563,19 @@ Organization:
...
+    Properties:
+      AccountName: genie-prod
+      RootEmail: aws-genie-prod@sagebase.org
+      Alias: org-sagebase-genie-prod
+      Tags:
+        <<: !Include ./_default_org_tags.yaml
+        Department: DPE
+        Project: Genie
+        AccountOwner: thomas.yu@sagebase.org
+        CostCenter: Genie-AACR BPC / 312100
```
zaro0508, line 577: no budget tags for this account?

zaro0508 (changes requested):
please add this account to an org OU

org-formation/organization.yaml
```diff
@@ -566,6 +567,21 @@ Organization:
...
+        budget-alarm-threshold: 1000
```
zaro0508, line 582: RecoverDevAccount is set to 10K but this only set to 1K?

org-formation/organization.yaml
```diff
@@ -566,6 +567,21 @@ Organization:
...
+    Properties:
+      AccountName: genie-prod
+      RootEmail: aws-genie-prod@sagebase.org
+      Alias: org-sagebase-genie-prod
+      Tags:
+        <<: !Include ./_default_org_tags.yaml
+        Department: DPE
+        Project: Genie
+        AccountOwner: thomas.yu@sagebase.org
+        CostCenter: Genie-AACR BPC / 312100
```
zaro0508, line 581: these tags are different than `RecoverDevAccount`, is that intentional?

zaro0508 (changes requested):

org-formation/organization.yaml
```diff
@@ -566,6 +567,21 @@ Organization:
         <<: !Include ./_default_org_tags.yaml
         <<: !Include ./_platform_org_tags.yaml
 
+  GenieProdAccount:
+    Type: OC::ORG::Account
+    Properties:
+      AccountName: genie-prod
```
zaro0508, line 573: should be the same as the alias

org-formation/organization.yaml
brucehoff: Yes.  Since this account will only host one app', I think the threshold should be lower.

org-formation/organization.yaml
brucehoff: Yes, this is the correct cost center for the Genie BPC Shiny app', as per @thomasyu888 .

zaro0508 (approved):

PR #664, opened 2022-12-16
SC-404: Update dependency version

xschildw:
This PR updates the dependency on sc-ec2-windows-jumpcloud.yaml to v1.1.44 (remove the user association step).

Depends on Sage-Bionetworks/service-catalog-library#288


zaro0508 (approved):
thanks @xschildw .  i guess i botched up tag v1.1.43

PR #671, opened 2023-01-06
[IT-2476] bootstrap accounts for AWS CDK

zaro0508:
This will bootstrap all of our AWS accounts in our organization with resources required to deploy with the AWS CDK.

It will generate the CDK boostrap template from the AWS CDK CLI then deploy that template to each account.


brucehoff (approved):
Love it!

PR #673, opened 2023-01-09
[ETL-309] add ci service acct for recover

rxu17:
**Purpose:** 
The AWS accounts recover-dev and recover-prod needs service account users for CI. This adds it.

PR Checklist:
[X] Describe and explain your intentions for this change
[X] Setup pre-commit and run the validators (info in README.md)

### DIFF
--- a/org-formation/600-access/_tasks.yaml
+++ b/org-formation/600-access/_tasks.yaml
@@ -173,6 +173,21 @@ DnTDevCIServiceAccounts:
       - !Ref DnTDevAccount
     Region: us-east-1
 
+# Service account for https://github.com/Sage-Bionetworks/recover
+RecoverCIServiceAccounts:
+  Type: update-stacks
+  Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.3.8/templates/IAM/service-account.yaml
+  StackName: recover-ci-service-account
+  Parameters:
+    ManagedPolicyArns:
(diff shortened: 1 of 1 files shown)


thomasyu888:
This looks good to me.  I defer to @brucehoff / @zaro0508 for final review.

brucehoff:
@rxu17 I don't think a service account is needed to allow GitHub to deploy to the Recover AWS account.  Rather what's needed is OIDC integration with an IAM *role* that a certain GitHub repo' can assume.  @zaro0508 has worked out how to do this for IT deployments and can give more guidance.

thomasyu888:
@brucehoff / @zaro0508, Since there is a January 27th deadline to set up the pilot ETL, I would like this or the OIDC integration to be worked out ASAP as this is currently semi-blocking our work. (I say semi, because we can technically deploy using using our admin jumpcloud privileges, but ideally we just use CI/CD)

Does it make most sense to set up an hour meeting to work through the steps of OIDC integration?  Also, would OIDC integration work on multiple accounts?  We need to be able to deploy in both the dev and prod account?

zaro0508:
please try setting up with github OIDC access.  example can be found at https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/org-formation/650-identity-providers/_tasks.yaml

rxu17:
Sounds good. Added new PR [#674](https://github.com/Sage-Bionetworks-IT/organizations-infra/pull/674) with the updates

PR #675, opened 2023-01-10
Remove Peter and add Leon

xschildw:
This PR removes the IAM user for Peter Harvey and adds a user for Leon Li in the Synapse dev account.

### DIFF
--- a/sceptre/synapsedev/templates/accounts.yaml
+++ b/sceptre/synapsedev/templates/accounts.yaml
@@ -71,10 +71,10 @@ Resources:
       LoginProfile:
         Password: !Ref InitNewUserPassword
         PasswordResetRequired: true
-  AWSIAMPeterHarveyUser:
+  AWSIAMLeonLiUser:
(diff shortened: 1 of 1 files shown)


brucehoff (approved):

sceptre/synapsedev/templates/accounts.yaml
```diff
@@ -71,10 +71,10 @@ Resources:
       LoginProfile:
         Password: !Ref InitNewUserPassword
         PasswordResetRequired: true
-  AWSIAMPeterHarveyUser:
+  AWSIAMLeonLiUser:
     Type: 'AWS::IAM::User'
     Properties:
-      UserName: peter.harvey@sagebase.org
+      UserName: leon.li@sagebase.org
```
zaro0508, line 77: We are actively denying creation of user LoginProfile at the org root level https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/org-formation/organization.yaml#L15

PR #683, opened 2023-01-17
IT: Remove more BastianSGs

xschildw:
This PR removes more SGs that are flagged by SecurityHub

### DIFF
--- a/sceptre/bridge/config/develop/BridgeServer2-vpc.yaml
+++ b/sceptre/bridge/config/develop/BridgeServer2-vpc.yaml
@@ -7,3 +7,4 @@ dependencies:
 parameters:
   VpcName: BridgeServer2-develop-vpc
   VpcSubnetPrefix: "172.48"
+  IncludeBastianSecurityGroup: "false"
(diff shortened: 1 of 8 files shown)


sceptre/bridge/config/develop/BridgeServer2-vpc.yaml
```diff
@@ -7,3 +7,4 @@ dependencies:
 parameters:
   VpcName: BridgeServer2-develop-vpc
   VpcSubnetPrefix: "172.48"
+  IncludeBastianSecurityGroup: "false"
```
zaro0508, line 10: i would double check to make sure bridge isn't using these SG before merging

zaro0508 (approved):
LGTM, i would just double check with bridge

sceptre/bridge/config/develop/BridgeServer2-vpc.yaml
xschildw: I executed aws --region us-east-1 --profile <profile> ec2 describe-network-interfaces --filters Name=group-id,Values=<bastionSGid> in bridgedev/bridgeprod and it returned nothing. Found a couple more groups in the process (bridge-aux was also using the vpc template, and bridge was using its own template to setup the SG), I added them to this PR.

PR #684, opened 2023-01-18
IT-1756: Add lambda and update README

xschildw:
This PR creates the lambda to do the actual suppression from findings queued up by the infra in #648 .

Depends on #648 .


xschildw:
Merging in #648

PR #686, opened 2023-01-23
[IT-2430] Suppress CIS 2.1.1 for public buckets

jesusaurus:
Suppress CIS 2.1.1 for our essentials and bootstrap buckets because they are intentionally public.

depends on: #648


xschildw (approved):
Not sure why all my changes show up... The new rule looks good!

PR #687, opened 2023-01-23
IT-1756: Fix circular dependency

xschildw:
I re-introduced the dependency by having the function name == resource name and referencing the function in the policy definition. Now hardcoding the function name in both the function and the policy.

### DIFF
--- a/org-formation/075-security-hub/security-hub-suppress-infra.yaml
+++ b/org-formation/075-security-hub/security-hub-suppress-infra.yaml
@@ -22,7 +22,7 @@ Resources:
           - logs:CreateLogStream
           - logs:PutLogEvents
           Effect: Allow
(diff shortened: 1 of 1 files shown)


org-formation/075-security-hub/security-hub-suppress-infra.yaml
```diff
@@ -335,12 +336,14 @@ Resources:
       Timeout: 120
   BatchupdateEventInvokeConfig:
     Type: AWS::Lambda::EventInvokeConfig
+    DependsOn: SecurityHubFindingsSuppressionLambdaFct
```
zaro0508, line 339: hmm, i wouldn't think DependsOn would be needed here or in BatchUpdateLambdaFctEventSourceEventSourceMapping since SecurityHubFindingsSuppressionLambdaFct is  already a reference in FunctionName param?

zaro0508 (approved):

org-formation/075-security-hub/security-hub-suppress-infra.yaml
xschildw: Agree, I'm being over-explicit...

PR #689, opened 2023-01-24
IT-2308: Add suppress rule for CIS1.4

xschildw:
This PR suppresses the findings for CIS1.4 (rotate IAM keys < 90 days) for bootstrap accounts and any service accounts.
Human accounts (synapseprod/dev/dw) are being decomissioned and the associated findings will disappear as they are.

### DIFF
--- a/org-formation/075-security-hub/security-hub-suppress-infra.yaml
+++ b/org-formation/075-security-hub/security-hub-suppress-infra.yaml
@@ -383,24 +383,21 @@ Resources:
           - Arn
         Id: Target0
 
-  # This rule suppresses findings with the given generator IDs only for the bootstrap account and service accounts
-  SuppressFindingsForPublicBucketsRule:
+  # This rule suppresses findings with the given generator IDs only for the IAM accounts specified
+  SuppressFindingsForIAMsRule:
     Type: AWS::Events::Rule
     Properties:
-      Description: SecHubSuppress findings (identified by generatorId) for the bootstrap and essentials buckets in all accounts
+      Description: SecHubSuppress findings (identified by generatorId) for given IAM accounts in all AWS accounts
       EventPattern:
         detail:
           findings:
             Resources:
               Id:
                 # synapsedw IT-2308, do not exempt human users
-                - prefix: 'arn:aws:iam::383874245509:user/bootstrap-AWSIAMTravisUser-'
                 - 'arn:aws:iam::383874245509:user/prod-warehouse'
                 # synapseprod IT-2396, do not exempt human users
-                - prefix: 'arn:aws:iam::325565585839:user/bootstrap-AWSIAMTravisUser-'
                 - 'arn:aws:iam::325565585839:user/ran-uploader'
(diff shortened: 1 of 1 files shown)


org-formation/075-security-hub/security-hub-suppress-infra.yaml
```diff
@@ -382,3 +382,39 @@ Resources:
...
+    Type: AWS::Events::Rule
+    Properties:
+      Description: SecHubSuppress findings (identified by generatorId) for the bootstrap and essentials buckets in all accounts
+      EventPattern:
+        detail:
+          findings:
+            Resources:
+              Id:
+                # synapsedw IT-2308, do not exempt human users
+                - prefix: 'arn:aws:iam::383874245509:user/bootstrap-AWSIAMTravisUser-'
```
xschildw, line 397: Is this account deprecated? Looking it either does not exist or has not been used in a long while (except scicomp/scipool).

zaro0508: i believe that is the `org-sagebase-synapsedw` account.  in any case `bootstrap-AWSIAMTravisUser-*` is old and should not be used anymore.  if it's still around then it should probably just be deleted

org-formation/075-security-hub/security-hub-suppress-infra.yaml
```diff
@@ -382,3 +382,39 @@ Resources:
           - SecurityHubFindingsQueue
           - Arn
         Id: Target0
+
+  # This rule suppresses findings with the given generator IDs only for the bootstrap account and service accounts
+  SuppressFindingsForPublicBucketsRule:
+    Type: AWS::Events::Rule
+    Properties:
+      Description: SecHubSuppress findings (identified by generatorId) for the bootstrap and essentials buckets in all accounts
```
zaro0508, line 390: seems like this is doing more than just boostrap and essentials?

org-formation/075-security-hub/security-hub-suppress-infra.yaml
```diff
@@ -382,3 +382,39 @@ Resources:
           - SecurityHubFindingsQueue
           - Arn
         Id: Target0
+
+  # This rule suppresses findings with the given generator IDs only for the bootstrap account and service accounts
+  SuppressFindingsForPublicBucketsRule:
```
zaro0508, line 387: this name doesn't seem to match what's being supressed?

zaro0508 (changes requested):

org-formation/075-security-hub/security-hub-suppress-infra.yaml
xschildw: Fixed

org-formation/075-security-hub/security-hub-suppress-infra.yaml
xschildw: Put a more generic description.

zaro0508 (approved):

PR #693, opened 2023-01-25
[DCA-34] setup github OIDC for automated deployments

zaro0508:


zaro0508:
going to redo with git-machete

PR #698, opened 2023-01-25
Add docs for Github OIDC

zaro0508:
* Add docs
* Change template extension to njk (nunjucks)

depends on #697


org-formation/650-identity-providers/_tasks.yaml
```diff
@@ -21,7 +21,7 @@ GithubOidcSageBionetworks:
 GithubOidcSageBionetworksIt:
   Type: update-stacks
   DependsOn: GithubOidcSageBionetworks
-  Template: github-oidc-provider-access.j2
+  Template: github-oidc-provider-access.njk
```
zaro0508, line 24: org-formation uses the nunjucks not jinja template engine that's why we are changing the file extension.

xschildw (approved):
Looks good to me...

PR #706, opened 2023-01-26
IT-2591: Enable key rotation on CDK assets key

xschildw:
The CDK creates a key that's not auto-rotating and gets flagged. This PR patches the CDK-generated CF template and adds 'EnableKeyRotation :true' property to the key resource.


zaro0508 (approved):

zaro0508 (approved):
I have provided comments on the related jira issue

xschildw:
Worked on MacOS with gsed, may need better pattern matching...

.github/workflows/aws-deploy.yaml
```diff
@@ -46,6 +46,8 @@ jobs:
         run: npm install
       - name: Generate AWS CDK bootstrap template
         run: npm run generate-cdk-bootstrap-template
+      - name: Add EnableKeyRotation property to CDK assets key
+        run: sed -i '/.*KeyPolicy/i\ \ \ \ \ \ EnableKeyRotation:\ true' cdk-bootstrap.yaml
```
zaro0508, line 50: the previous step runs `npm run generate-cdk-bootstrap-template`,  generate-cdk-bootstrap-template in package.json executes `npx cdk bootstrap --show-template > org-formation/200-baseline/cdk-bootstrap.yaml`.  the cdk-bootstrap file is in a subdirectory not in the current directory.  therefore you probably want `sed -i '/.*KeyPolicy/i\ \ \ \ \ \ EnableKeyRotation:\ true' ./org-formation/200-baseline/cdk-bootstrap.yaml/cdk-bootstrap.yaml` ?

i think modifying a json might be easier than yaml file because you don't need to worry about the spacing and there are other tools like `jq` to change it.   if you want the previous step to generate a json just  modify the command to do this.. `npx cdk bootstrap --json --show-template > org-formation/200-baseline/cdk-bootstrap.json`

zaro0508 (changes requested):
please update the PR description to reflect what this PR is actually doing.

.github/workflows/aws-deploy.yaml
```diff
@@ -45,7 +45,9 @@ jobs:
       - name: Install dependencies
         run: npm install
       - name: Generate AWS CDK bootstrap template
-        run: npm run generate-cdk-bootstrap-template
+        run: npx cdk bootstrap --json --show-template > org-formation/200-baseline/tmptemplate.json
```
zaro0508, line 48: the actual cdk command should be in package.json, look for `generate-cdk-bootstrap-template`

`tmptemplate.json` should not be added to the repo so please add it to `.gitignore`

zaro0508 (changes requested):
Again, please update the PR description to reflect the true nature of this change.

.github/workflows/aws-deploy.yaml
```diff
@@ -45,7 +45,9 @@ jobs:
...
+      - name: Add EnableKeyRotation property to CDK assets key
+        run: jq '.Resources.FileAssetsBucketEncryptionKey.Properties += {"EnableKeyRotation" :true}' org-formation/200-baseline/tmptemplate.json > org-formation/200-baseline/cdk-bootstrap.json
```
zaro0508, line 50: much nicer than the `sed` before 😄 .  please add `cdk-bootstrap.json` to `.gitignore` as well.

.github/workflows/aws-deploy.yaml
xschildw: OK

zaro0508 (approved):

PR #707, opened 2023-01-26
Update Sage VPN certificate

zaro0508:
The VPN server certificate expired which caused the AWS client VPN to fail to connect with message..
"Connection failed because of a TLS handshake error"

The certificate was manually generated and deploy from AWS instruction: https://docs.aws.amazon.com/vpn/latest/clientvpn-admin/cvpn-getting-started.html#cvpn-getting-started-certs


xschildw (approved):
Thanks!

jesusaurus (approved):

PR #712, opened 2023-01-31
[IT-2629] Update node version in aws-deploy

jesusaurus:
Update our version of node, and also update github actions to upgrade their versions of node as well.


zaro0508 (approved):
i was hoping to get around to updating thanks for doing it.

zaro0508 (approved):

PR #714, opened 2023-02-01
IT-2591: Fix validation error

xschildw:
This PR should fix the deployment error caused by #706. Do I need something for the jq step?

### DIFF
--- a/org-formation/200-baseline/_tasks.yaml
+++ b/org-formation/200-baseline/_tasks.yaml
@@ -23,8 +23,8 @@ OrgFormationBuildAccessRole:
 
 CdkBootstrapDefault:
   Type: update-stacks
-  Template: tmptemplate.json  # auto generated from "npx cdk bootstrap --json --show-template > org-formation/200-baseline/tmptemplate.json"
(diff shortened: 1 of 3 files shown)


xschildw:
It looked like the jq step was not executed and I noticed it was the only one that was not run with 'npm run' so I re-wrote as a script run by npm...

org-formation/200-baseline/_tasks.yaml
```diff
@@ -23,7 +23,7 @@ OrgFormationBuildAccessRole:
 
 CdkBootstrapDefault:
   Type: update-stacks
-  Template: cdk-bootstrap.yaml  # auto generated from "npx cdk bootstrap --show-template > cdk-bootstrap.yaml"
+  Template: cdk-bootstrap.json  # auto generated from "npx cdk bootstrap --json --show-template > org-formation/200-baseline/tmptemplate.json and jq"
```
zaro0508, line 26: looks good but i think this was the only reason why it didn't run.

zaro0508 (approved):
LGTM

PR #715, opened 2023-02-01
[IT-2594] Add finops-api.sageit.org DNS name

jesusaurus:
Create a friendly CNAME (finops-api.sageit.org) for the lambda-mips-api Cloudfront distribution.

depends on: Sage-Bionetworks-IT/lambda-mips-api#11 Sage-Bionetworks/aws-infra#372

### DIFF
--- a/org-formation/800-redirects/_tasks.yaml
+++ b/org-formation/800-redirects/_tasks.yaml
@@ -19,3 +19,20 @@ SsoRedirect:   # Redirect sso.sageit.org to the AWS SSO Start URL
     TargetHostName: "d-906769aa66.awsapps.com"
     # and a path to our specific config
     TargetKey: "start#/"
+
+FinopsApiRedirect:  # redirect finops-api.sageit.org to lambda-mips-api cloudfront distribution
+  Type: update-stacks
+  Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.6.3/templates/R53/cname.yaml
+  StackName: !Sub '${resourcePrefix}-finops-api-redirect'
+  StackDescription: Setup a redirect to the lambda-mips-api cloudfront distribution
+  DefaultOrganizationBindingRegion: !Ref primaryRegion
+  DefaultOrganizationBinding:
+    Account: !Ref SageITAccount
+  Parameters:
+    # the endpoint we are redirecting from
+    SourceHostName: "finops-api.sageit.org"
+    SourceAcmCertificateArn: !CopyValue [!Sub '${primaryRegion}-sageit-org-acm-cert-CertificateArn']
+    # ID of the sageit.org zone (in sageit account)
+    SourceHostedZoneId: "Z0478495257GEB73WFM63"
+    # the endpoint we are redirecting to in organizations account
+    TargetHostName: !CopyValue [!Sub '${primaryRegion}-${resourcePrefix}-mips-microservice-CloudfrontDomain']
--- a/org-formation/800-redirects/_tasks.yaml
+++ b/org-formation/800-redirects/_tasks.yaml
@@ -22,7 +22,7 @@ SsoRedirect:   # Redirect sso.sageit.org to the AWS SSO Start URL
 
 FinopsApiRedirect:  # redirect finops-api.sageit.org to lambda-mips-api cloudfront distribution
   Type: update-stacks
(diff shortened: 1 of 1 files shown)


org-formation/800-redirects/_tasks.yaml
```diff
@@ -19,3 +19,20 @@ SsoRedirect:   # Redirect sso.sageit.org to the AWS SSO Start URL
...
+  Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.6.3/templates/R53/cname.yaml
+  StackName: !Sub '${resourcePrefix}-finops-api-redirect'
+  StackDescription: Setup a redirect to the lambda-mips-api cloudfront distribution
+  DefaultOrganizationBindingRegion: !Ref primaryRegion
+  DefaultOrganizationBinding:
+    Account: !Ref SageITAccount
+  Parameters:
+    # the endpoint we are redirecting from
+    SourceHostName: "finops-api.sageit.org"
+    SourceAcmCertificateArn: !CopyValue [!Sub '${primaryRegion}-sageit-org-acm-cert-CertificateArn']
```
xschildw, line 34: There's no cert as param on the template (which I think is correct). Should not the cert be on the distribution (and the distribution needs to be configured to accept the sourceHostName)?

jesusaurus: You're right, I removed it from the template but forgot to remove the parameter here; and I hadn't realized I need to set the certificate (and alias name) on the lambda cloudfront distribution but that definitely makes sense. I'll move the parameter to the lambda template.

zaro0508 (changes requested):

org-formation/800-redirects/_tasks.yaml
```diff
@@ -19,3 +19,20 @@ SsoRedirect:   # Redirect sso.sageit.org to the AWS SSO Start URL
     TargetHostName: "d-906769aa66.awsapps.com"
     # and a path to our specific config
     TargetKey: "start#/"
+
+FinopsApiRedirect:  # redirect finops-api.sageit.org to lambda-mips-api cloudfront distribution
+  Type: update-stacks
+  Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.6.3/templates/R53/cname.yaml
```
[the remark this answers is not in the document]
jesusaurus: The s3-redirector template is not appropriate in this case because we're not hosting static files out of an S3 bucket and we have already created the Cloudfront distribution in the lambda template. In the lambda use-case, we only need to create a CNAME record in Route 53 for the Cloudfront distribution we've already created in another stack, so I chose to create a new template for that use case.

org-formation/800-redirects/_tasks.yaml
```diff
@@ -19,3 +19,20 @@ SsoRedirect:   # Redirect sso.sageit.org to the AWS SSO Start URL
...
+  DefaultOrganizationBinding:
+    Account: !Ref SageITAccount
+  Parameters:
+    # the endpoint we are redirecting from
+    SourceHostName: "finops-api.sageit.org"
+    SourceAcmCertificateArn: !CopyValue [!Sub '${primaryRegion}-sageit-org-acm-cert-CertificateArn']
+    # ID of the sageit.org zone (in sageit account)
+    SourceHostedZoneId: "Z0478495257GEB73WFM63"
+    # the endpoint we are redirecting to in organizations account
+    TargetHostName: !CopyValue [!Sub '${primaryRegion}-${resourcePrefix}-mips-microservice-CloudfrontDomain']
```
zaro0508, line 38: this lookup is going to try to find `!Sub '${primaryRegion}-${resourcePrefix}-mips-microservice-CloudfrontDomain'` in the SageITAccount which probably doesn't exist.  You will need to tell CopyValue to look in the account that has that output value, take a look at the [CopyValue docs](https://github.com/org-formation/org-formation-cli/blob/master/docs/task-files.md#copyvalue)

org-formation/800-redirects/_tasks.yaml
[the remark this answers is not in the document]
zaro0508: ok

zaro0508 (changes requested):

org-formation/800-redirects/_tasks.yaml
jesusaurus: thanks, done

zaro0508 (approved):

PR #717, opened 2023-02-02
IT-2427: Add generatorId

xschildw:
This PR adds a generatorId for cis-aws-foundations-benchmark/v/1.4.0/3.7(Cloudtrail logs encryption) to be suppressed for all accounts. This is the same resolution as arn:aws:securityhub:::ruleset/cis-aws-foundations-benchmark/v/1.2.0/rule/2.7.

### DIFF
--- a/org-formation/075-security-hub/security-hub-suppress-infra.yaml
+++ b/org-formation/075-security-hub/security-hub-suppress-infra.yaml
@@ -368,6 +368,7 @@ Resources:
               - 'arn:aws:securityhub:::ruleset/cis-aws-foundations-benchmark/v/1.2.0/rule/1.14'
               - 'aws-foundational-security-best-practices/v/1.0.0/IAM.6'
               - 'arn:aws:securityhub:::ruleset/cis-aws-foundations-benchmark/v/1.2.0/rule/2.7'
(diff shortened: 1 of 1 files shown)


org-formation/075-security-hub/security-hub-suppress-infra.yaml
```diff
@@ -368,6 +368,7 @@ Resources:
               - 'arn:aws:securityhub:::ruleset/cis-aws-foundations-benchmark/v/1.2.0/rule/1.14'
               - 'aws-foundational-security-best-practices/v/1.0.0/IAM.6'
               - 'arn:aws:securityhub:::ruleset/cis-aws-foundations-benchmark/v/1.2.0/rule/2.7'
+              - 'cis-aws-foundations-benchmark/v/1.4.0/3.7'
```
zaro0508, line 371: i just noticed that some of these IDs are ARNs while others are not.  is this ok?

zaro0508 (approved):

org-formation/075-security-hub/security-hub-suppress-infra.yaml
xschildw: Yeah, it's not clear to me either but that's how they appear in the console. I'll double-check that they do get suppressed.

PR #719, opened 2023-02-03
IT-2428: Suppress CIS2.1.5.1 for securitycentral

xschildw:
This PR suppresses the finding for CIS 2.1.5.1 (Block all on S3 at account level. I put it in the same rule as the ones that work at the bucket level, it should work because only that combo generatorId/resourceId should match, renamed the description accordingly. Not sure why other accounts did not get flagged (checked SecHub). Note: S3 behavior is changing to block all access by default so I'd expect this control to disappear in future version of CIS.

### DIFF
--- a/org-formation/075-security-hub/security-hub-suppress-infra.yaml
+++ b/org-formation/075-security-hub/security-hub-suppress-infra.yaml
@@ -384,11 +384,11 @@ Resources:
           - Arn
         Id: Target0
 
-  # This rule suppresses findings with the given generator IDs only for the bootstrap and essentials buckets
+  # This rule suppresses findings with the given generator IDs only for the infra resources
   SuppressFindingsForPublicBucketsRule:
     Type: AWS::Events::Rule
     Properties:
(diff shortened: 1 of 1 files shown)


org-formation/075-security-hub/security-hub-suppress-infra.yaml
```diff
@@ -397,9 +397,12 @@ Resources:
                 - prefix: 'arn:aws:s3:::bootstrap-awss3cloudformationbucket-'
                 - prefix: 'arn:aws:s3:::essentials-awss3lambdaartifactsbucket-'
                 - prefix: 'arn:aws:s3:::cf-templates-'
+                # This should only match with 2.1.5.1
+                - 'AWS::::Account:140124849929'
```
zaro0508, line 401: if this is running on securitycentral you should be able to use `!Sub 'AWS::::Account:${AWS::AccountId}'`

zaro0508 (approved):

PR #720, opened 2023-02-07
IT-2308/2309: Update bootstrao/essentials version

xschildw:
This PR should remove the BootstrapTravisUser from the synapseprod/synapsedw accounts, it also removes the VPC peering authorizer role from the accounts (IT-2636).

Note1: The BootstrapTravisUser has not been used in these accounts since 2020 (see IT-2635)
Note2: Per comments in bootstrap.yaml, the BootstrapTravisUser should be stripped from groups etc. prior to merging/deploying.


zaro0508 (approved):
seems fine but how come you don't update in all `essentials.yaml` files?

jesusaurus (approved):

PR #726, opened 2023-02-09
[IT-2594] Create new certificate for *.sageit.org

jesusaurus:
Once this certificate has been manually validated, it can be added to the mips microservice cloudfront distribution.

### DIFF
--- a/org-formation/100-shared-dns/README.md
+++ b/org-formation/100-shared-dns/README.md
@@ -1,12 +1,40 @@
 ## Purpose
-These hosted zones will contain DNS records for general-purpose services
-such as DCA and Schematic.
+
+Manage hosted zones and wildcard certificates for those zones.
 
 Wildcard certs are generated for the apex records and are intended to be
 shared by all services hosted in the zone by passing the ARN of the
-appropriate certificate to the CDK application deploying the service.
+appropriate certificate to the hosting service (e.g. CloudFront or an ALB).
+
+Once the hosting service is deployed with the appropriate certificate, create
+a DNS CNAME record to redirect our custom domain name to the AWS domain name
+used by the hosting service.
+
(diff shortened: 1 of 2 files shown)


zaro0508 (commented):
I like this idea however certificate creation requires a manual DNS validation step therefore this is not fully automated and worse it will hang waiting for validation.

Also I think it may make more sense adding this to cfn template for the app because it's just another resource that the app needs.

xschildw (commented):
I like the idea too, we'll have to deal with the validation no matter what. It would also make it easier to reuse certs if possible (i.e. can't we create a cert covering *.sageit.org that would cover all the endpoints).

jesusaurus:
Deploy a new wildcard certificate for sageit.org in both the SageIT and
the AdminCentral accounts. This is intended to replace the certificate
created in by sceptre in the SageIT account, and to be used by the
finops lambdas.

xschildw (approved):

org-formation/010-certificates/_tasks.yaml
```diff
@@ -0,0 +1,15 @@
...
+
+FinopsApiCertificate:
+  Type: update-stacks
+  Template: !Sub 'https://${AdminCentralCfnBucket}.s3.amazonaws.com/aws-infra/v0.6.4/acm-certificate.yaml'
+  StackName: !Sub '${resourcePrefix}-sageit-cert'
+  DefaultOrganizationBinding:
+    IncludeMasterAccount: false
+    Account:
+      - !Ref AdminCentralAccount
+      - !Ref SageITAccount
```
zaro0508, line 12: i need to think about this

zaro0508 (changes requested):

zaro0508 (approved):

PR #729, opened 2023-02-18
Fix deployment failure

zaro0508:
* fix typo
* org-formation requires `TemplatingContext` parameter when passing a nunjucks file[1].

[1] https://github.com/org-formation/org-formation-cli/blob/master/docs/task-files.md#templating
```
Note: If you want templating without passing in any data you must
set TemplatingContext: {} to trigger templating.
```


brucehoff (approved):
Thank you.

PR #734, opened 2023-02-21
[DCA-8] AWS account for Schematic app

zaro0508:
Create an AWS account to run Schematica appp[1]

[1] https://github.com/Sage-Bionetworks/schematic

### DIFF
--- a/org-formation/organization.yaml
+++ b/org-formation/organization.yaml
@@ -50,6 +50,7 @@ Organization:
         - !Ref IncludeChopProdAccount
         - !Ref RecoverProdAccount
         - !Ref GenieProdAccount
+        - !Ref SchematicProdAccount
         - !Ref DCAProdAccount
         - !Ref DpeProdAccount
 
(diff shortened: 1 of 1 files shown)


xschildw (approved):
As long as TBD is valid value for costCenter...

org-formation/organization.yaml
```diff
@@ -567,6 +568,21 @@ Organization:
...
+    Properties:
+      AccountName: org-sagebase-schematic-prod
+      RootEmail: aws-schematic-prod@sagebase.org
+      Alias: org-sagebase-schematic-prod
+      Tags:
+        <<: !Include ./_default_org_tags.yaml
+        Department: FAIR
+        Project: DCA
+        AccountOwner: milen.nikolov@sagebase.org
+        CostCenter: "TBD"
```
zaro0508, line 582: @brucehoff @milen-sage please check Department/Project/AccountOwner and CostCenter are correct.

brucehoff (changes requested):
I though we agreed just to use the 'dca-prod' account for both the DCA app' and the Schematic APIs.  The account was already set up, under DCA-8.  What is the benefit of a separate AWS account fir one part of the system?

zaro0508:
@brucehoff we had discussions but i don't remember us making a decision one way or the other.  The benefits of separate account for schematic is to provide full security isolation of the two apps.  This could be important if we think that (one day) the apps would have a separate set of developers managing and maintaining them.  If that's not important then deploying both to the same AWS account is fine.

brucehoff:
> @brucehoff we had discussions but i don't remember us making a decision one way or the other. The benefits of separate account for schematic is to provide full security isolation of the two apps. This could be important if we think that (one day) the apps would have a separate set of developers managing and maintaining them. If that's not important then deploying both to the same AWS account is fine.

i think the two app's will always be maintained by the same team and so it's OK to deploy them together.  We can check with @milen-sage.

zaro0508:
OK.  no need to create new aws account so closing this PR.

PR #737, opened 2023-02-23
IT-2635: Update bootstrap version to remove boostrapTravisUser

xschildw:
This PR updates the bootstrap.yaml version, which removes the bootstrapTravisUser.
Depends on #732 to remove a dependency first.


zaro0508 (approved):

zaro0508:
i don't think the reference to the depends on PR is accurate.

PR #739, opened 2023-02-24
[DCA-37] Add CNAME for dca-test application

jesusaurus:
Now that we have created a wildcard certificate for our application subdomain and are using it in our application, set up a CNAME redirect so that the application domain matches the certificate domain.

Depends on: https://github.com/Sage-Bionetworks/data_curator-infra/pull/29

### DIFF
--- a/org-formation/800-redirects/_tasks.yaml
+++ b/org-formation/800-redirects/_tasks.yaml
@@ -1,6 +1,8 @@
 Parameters:
   <<: !Include '../_parameters.yaml'
 
+# *.sageit.org
+
 SsoRedirect:   # Redirect sso.sageit.org to the AWS SSO Start URL
   Type: update-stacks
   Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.4.8/templates/s3-redirector.yaml
@@ -35,3 +37,22 @@ FinopsApiRedirect:  # redirect finops-api.sageit.org to lambda-mips-api cloudfro
     SourceHostedZoneId: "Z0478495257GEB73WFM63"
     # the value of the CNAME record
     TargetHostName: !CopyValue [!Sub '${primaryRegion}-${resourcePrefix}-mips-microservice-CloudfrontDomain', !Ref AdminCentralAccount]
+
+
+# *.app.sagebionetworks.org
+
+DcaDevAppRedirect:  # redirect dca-test.app.sagebionetworks.org to data_curator-infra ALB
+  Type: update-stacks
+  Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.6.4/templates/R53/cname.yaml
+  StackName: !Sub '${resourcePrefix}-finops-api-redirect'
+  StackDescription: Setup a redirect to the lambda-mips-api cloudfront distribution
+  DefaultOrganizationBindingRegion: !Ref primaryRegion
+  DefaultOrganizationBinding:
+    Account: !Ref SageITAccount
+  Parameters:
+    # the name of the CNAME record
+    SourceHostName: "dca-test.app.sagebionetworks.org"
+    # ID of the app.sagebionetworks.org zone (in sageit account)
+    SourceHostedZoneId: !CopyValue [!Sub '${primaryRegion}-${resourcePrefix}-sagebio-app-zone-HostedZoneId']
+    # the value of the CNAME record
+    TargetHostName: !CopyValue [!Sub 'dca-dev-DockerFargateStack-LoadBalancerDNS', !Ref DnTDevAccount]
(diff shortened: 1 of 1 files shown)


org-formation/800-redirects/_tasks.yaml
```diff
@@ -35,3 +37,22 @@ FinopsApiRedirect:  # redirect finops-api.sageit.org to lambda-mips-api cloudfro
     SourceHostedZoneId: "Z0478495257GEB73WFM63"
     # the value of the CNAME record
     TargetHostName: !CopyValue [!Sub '${primaryRegion}-${resourcePrefix}-mips-microservice-CloudfrontDomain', !Ref AdminCentralAccount]
+
+
+# *.app.sagebionetworks.org
+
+DcaDevAppRedirect:  # redirect dca-test.app.sagebionetworks.org to data_curator-infra ALB
```
brucehoff, line 44: @jesusaurus can you confirm that, though you use the term 'redirect' there isn't literally an HTTP redirect (3xx response), but rather traffic is forwarded to the target ALB?

jesusaurus: @brucehoff correct, this is not an HTTP redirect, the redirection is happening when looking up the IP address of the host name (which I have an old habit of also calling a redirect, but I can change that terminology if its confusing)

org-formation/800-redirects/_tasks.yaml
```diff
@@ -35,3 +37,22 @@ FinopsApiRedirect:  # redirect finops-api.sageit.org to lambda-mips-api cloudfro
...
+  Type: update-stacks
+  Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.6.4/templates/R53/cname.yaml
+  StackName: !Sub '${resourcePrefix}-finops-api-redirect'
+  StackDescription: Setup a redirect to the lambda-mips-api cloudfront distribution
+  DefaultOrganizationBindingRegion: !Ref primaryRegion
+  DefaultOrganizationBinding:
+    Account: !Ref SageITAccount
+  Parameters:
+    # the name of the CNAME record
+    SourceHostName: "dca-test.app.sagebionetworks.org"
```
zaro0508, line 54: this is more likely to be dca-dev / dca-staging / dca-prod

org-formation/800-redirects/_tasks.yaml
```diff
@@ -35,3 +37,22 @@ FinopsApiRedirect:  # redirect finops-api.sageit.org to lambda-mips-api cloudfro
...
     # the value of the CNAME record
     TargetHostName: !CopyValue [!Sub '${primaryRegion}-${resourcePrefix}-mips-microservice-CloudfrontDomain', !Ref AdminCentralAccount]
+
+
+# *.app.sagebionetworks.org
+
+DcaDevAppRedirect:  # redirect dca-test.app.sagebionetworks.org to data_curator-infra ALB
+  Type: update-stacks
+  Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.6.4/templates/R53/cname.yaml
+  StackName: !Sub '${resourcePrefix}-finops-api-redirect'
```
zaro0508, line 47: finops?

org-formation/800-redirects/_tasks.yaml
```diff
@@ -35,3 +37,22 @@ FinopsApiRedirect:  # redirect finops-api.sageit.org to lambda-mips-api cloudfro
...
+  StackDescription: Setup a redirect to the lambda-mips-api cloudfront distribution
```
zaro0508, line 48: mips?

org-formation/800-redirects/_tasks.yaml
zaro0508: could we add a comment referencing the GH repo for this redirect?

zaro0508 (changes requested):

org-formation/800-redirects/_tasks.yaml
jesusaurus: @zaro0508 done

org-formation/800-redirects/_tasks.yaml
jesusaurus: good call, I've updated this name to align with the context name

org-formation/800-redirects/_tasks.yaml
jesusaurus: whoops, I guess I needed a little more coffee this morning

org-formation/800-redirects/_tasks.yaml
jesusaurus: ditto

zaro0508 (approved):

PR #743, opened 2023-02-27
[DCA-34] Allow deployment from multiple branches

zaro0508:
For deployments where there is a more involved change to the infrastructure we may want to manage the deployment of that change to one account/environment (dev) before propogating it to other accounts (prod).

For this we setup multiple git branches (dev/stage/prod) to allow us to propogate the deployments which means we need to allow OIDC to deploy from multiple branches.

### DIFF
--- a/org-formation/650-identity-providers/_tasks.yaml
+++ b/org-formation/650-identity-providers/_tasks.yaml
@@ -70,7 +70,7 @@ GithubOidcSageBionetworksDataCuratorInfra:
   TemplatingContext:
     Repositories:
       - name: "data_curator-infra"
-        branch: "main"
+        branch: "*"
   DefaultOrganizationBinding:
     Account:
       - !Ref DnTDevAccount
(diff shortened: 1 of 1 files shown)


afwillia (approved):

jesusaurus (approved):

org-formation/650-identity-providers/_tasks.yaml
```diff
@@ -70,7 +70,7 @@ GithubOidcSageBionetworksDataCuratorInfra:
   TemplatingContext:
     Repositories:
       - name: "data_curator-infra"
-        branch: "main"
+        branch: "*"
```
brucehoff, line 73: What do you think of splitting this into two entries, one to deploy from `dev` branch into `DnTDevAccount` and another to deploy from `prod` branch into `DCAProdAccount`? That would help prevent an application developer from accidentally deploying dev' changes to production.

org-formation/650-identity-providers/_tasks.yaml
```diff
@@ -90,7 +90,7 @@ GithubOidcSageBionetworksSchematicInfra:
   TemplatingContext:
     Repositories:
       - name: "schematic-infra"
-        branch: "main"
+        branch: "*"
```
brucehoff, line 93: Same comment as above, for the `schematic-infra` repo'.

PR #750, opened 2023-03-08
[IT-2774] Update Service Catalog CostCenter tag list

zaro0508:
Switch from getting valid tags values from static files to a rest
endpoint that provides updated tag values from the source of truth.

We use the sceptre-request-resolver[1] to get the cost center tags
as a string then we use the jinja fromjson[2] filter to convert it to list
object so that we can iterate on each item to setup tags for the
service catalog.

[1] https://github.com/Sceptre/sceptre-request-resolver
[2] https://pypi.org/project/jinja2-from-json-extension/

depends on https://github.com/Sceptre/sceptre-request-resolver/pull/1


jesusaurus (approved):

sceptre/scipool/templates/sc-tag-options.j2
```diff
@@ -22,7 +22,8 @@ Resources:
   {% endfor %}
 {% endif %}
 {% if sceptre_user_data.CostCenters is defined %}
-  {% for CostCenter in sceptre_user_data.CostCenters %}
+  {% set CostCenters = sceptre_user_data.CostCenters | fromjson %}
```
jesusaurus, line 25: Have you tested this locally? I'm not finding much documentation for `jinja2_from_json_extension`, but it looks like maybe this should be `from_json` instead of `fromjson`? https://github.com/metwork-framework/jinja2_from_json_extension/blob/v1.0.0/jinja2_from_json_extension/__init__.py#L20

sceptre/scipool/config/prod/sc-tag-options.yaml
```diff
@@ -14,7 +14,7 @@ dependencies:
 sceptre_user_data:
   Departments: !file sc-tag-options/internal/Departments.json
   Projects: !file sc-tag-options/internal/Projects.json
-  CostCenters: !file https://{{stack_group_config.admincentral_cf_bucket}}.s3.amazonaws.com/aws-infra/v0.3.6/tags/SageFinancialProgramCodes.json
+  CostCenters: !request https://finops-api.sageit.org/tags?limit=25
```
jesusaurus, line 17: I want to move this to `https://mips-api.finops.sageit.org/tags?limit=25` so that I can use `https://cost-rules.finops.sageit.org/` for the cost-rules lambda: https://github.com/Sage-Bionetworks-IT/organizations-infra/pull/775

sceptre/scipool/templates/sc-tag-options.j2
zaro0508: you are correct, the jinja extension package does not contain any docs.  I had to look at the code to figure out how it works.  You are also correct that it should be `from_json`, I have updated to the correct filter name.  I have tested the latest update of this PR and it does seem to work as expected.

sceptre/scipool/config/prod/sc-tag-options.yaml
zaro0508: ok, can we go ahead with this PR and update this URL after you change the endpoint?

jesusaurus (approved):

zaro0508:
closing this PR in favor of PR #827

PR #751, opened 2023-03-08
IT-2672: Fix policy

xschildw:
This PR should allow assuming the role on tag events.

### DIFF
--- a/org-formation/650-identity-providers/github-oidc-provider-access.njk
+++ b/org-formation/650-identity-providers/github-oidc-provider-access.njk
@@ -89,12 +89,15 @@ Resources:
             Condition:
               StringEquals:
                 token.actions.githubusercontent.com:aud: !Ref ClientIdList
-  {% if Repository.branch == '*' %}
               StringLike:
-                token.actions.githubusercontent.com:sub: !Sub repo:${GitHubOrg}/{{ Repository.name }}:{{Repository.branch }}
+                token.actions.githubusercontent.com:sub:
+  {% if Repository.branch == '*' %}
+                  - "!Sub repo:${GitHubOrg}/{{ Repository.name }}:{{Repository.branch }}",
   {% else %}
-                token.actions.githubusercontent.com:sub: !Sub repo:${GitHubOrg}/{{ Repository.name }}:ref:refs/heads/{{ Repository.branch }}
+                  - "!Sub repo:${GitHubOrg}/{{ Repository.name }}:ref:refs/heads/{{ Repository.branch }}",
   {% endif %}
+                  - "!Sub repo:${GitHubOrg}/{{ Repository.name }}:ref:refs/tags/*"
+
 {% endfor %}
 Outputs:
 {% for Repository in Repositories %}
--- a/org-formation/650-identity-providers/github-oidc-provider-access.njk
+++ b/org-formation/650-identity-providers/github-oidc-provider-access.njk
@@ -90,14 +90,14 @@ Resources:
               StringEquals:
                 token.actions.githubusercontent.com:aud: !Ref ClientIdList
               StringLike:
-                token.actions.githubusercontent.com:sub:
+                token.actions.githubusercontent.com:sub: [
   {% if Repository.branch == '*' %}
-                  - "!Sub repo:${GitHubOrg}/{{ Repository.name }}:{{Repository.branch }}",
+                  "!Sub repo:${GitHubOrg}/{{ Repository.name }}:{{Repository.branch }}",
   {% else %}
-                  - "!Sub repo:${GitHubOrg}/{{ Repository.name }}:ref:refs/heads/{{ Repository.branch }}",
+                  "!Sub repo:${GitHubOrg}/{{ Repository.name }}:ref:refs/heads/{{ Repository.branch }}",
   {% endif %}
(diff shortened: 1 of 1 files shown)


org-formation/650-identity-providers/github-oidc-provider-access.njk
```diff
@@ -89,12 +89,15 @@ Resources:
...
               StringLike:
-                token.actions.githubusercontent.com:sub: !Sub repo:${GitHubOrg}/{{ Repository.name }}:{{Repository.branch }}
+                token.actions.githubusercontent.com:sub: [
+  {% if Repository.branch == '*' %}
+                  "!Sub repo:${GitHubOrg}/{{ Repository.name }}:{{Repository.branch }}",
   {% else %}
-                token.actions.githubusercontent.com:sub: !Sub repo:${GitHubOrg}/{{ Repository.name }}:ref:refs/heads/{{ Repository.branch }}
+                  "!Sub repo:${GitHubOrg}/{{ Repository.name }}:ref:refs/heads/{{ Repository.branch }}",
   {% endif %}
+                  "!Sub repo:${GitHubOrg}/{{ Repository.name }}:ref:refs/tags/*"
```
xschildw, line 99: Only wildcards '*' and '?' allowed from what I see, used '*' since this is shared with other repos.

zaro0508: this is not gonna work.  `StringEquals` and `StringLike` are dict objects, it cannot take a list of conditions.    you can test it out yourself by editing the trust policy in the AWS console. Furthermore AWS does not allow adding multiple `StringEquals` and `StringLike` conditions in a policy.

zaro0508 (changes requested):

org-formation/650-identity-providers/github-oidc-provider-access.njk
xschildw: It's not a list of conditions, it's a list of values for 'token.actions.githubusercontent.com:sub'. As I understand they should be ORed. In the end I'm just adding a value to the existing StringLike clause, am I not?
Conditions:
  StringEquals:
    token.actions.githubusercontent.com:aud: !Ref ClientIdList
  StringLike: [
    "!Sub repo:${GitHubOrg}/{{ Repository.name }}:ref:refs/heads/{{ Repository.branch }}",
    "!Sub repo:${GitHubOrg}/{{ Repository.name }}:ref:refs/tags/*"
  ]

zaro0508: ahh i see.  the list is inside of `token.actions.githubusercontent.com:sub:` which does work.  so it's something like this?
```
"Condition": {
	"StringEquals": {
		"token.actions.githubusercontent.com:aud": "sts.amazonaws.com"
	},
	"StringLike": {
		"token.actions.githubusercontent.com:sub": ["repo:Sage-Bionetworks/aws-infra:ref:refs/heads/master",
		"repo:Sage-Bionetworks/aws-infra:ref:refs/tags/*"]
	}
}
```

org-formation/650-identity-providers/github-oidc-provider-access.njk
```diff
@@ -89,12 +89,15 @@ Resources:
...
                 token.actions.githubusercontent.com:aud: !Ref ClientIdList
-  {% if Repository.branch == '*' %}
               StringLike:
-                token.actions.githubusercontent.com:sub: !Sub repo:${GitHubOrg}/{{ Repository.name }}:{{Repository.branch }}
+                token.actions.githubusercontent.com:sub: [
+  {% if Repository.branch == '*' %}
+                  "!Sub repo:${GitHubOrg}/{{ Repository.name }}:{{Repository.branch }}",
   {% else %}
-                token.actions.githubusercontent.com:sub: !Sub repo:${GitHubOrg}/{{ Repository.name }}:ref:refs/heads/{{ Repository.branch }}
+                  "!Sub repo:${GitHubOrg}/{{ Repository.name }}:ref:refs/heads/{{ Repository.branch }}",
```
zaro0508, line 97: To exactly match a branch this should go under `StringEquals` not `StringLike`.

org-formation/650-identity-providers/github-oidc-provider-access.njk
xschildw: Yes, sorry I see now I forgot the key in the StringLike...

org-formation/650-identity-providers/github-oidc-provider-access.njk
xschildw: I don't think that would work. My understanding from the docs (https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_multi-value-conditions.html) is that the conditions operators are ANDed (and so are the keys under an operator). If I move the match on the branch to StringEquals, a push will match on StringEquals but will not match on StringLike (similarly a tag will match on StringLike but won't on StringEquals). With StringEquals only dealing with 'aud' and StringLike only dealing with 'sub', the StringEquals should match for both push and tag, and so should the StringLike (since it has both possible values). And the StringLike value without a wildcard is like a StringEquals.

zaro0508: i see, it's so complicated. thanks for the info.

zaro0508 (approved):

PR #757, opened 2023-03-10
Add schematic-prod CNAME

jesusaurus:
Add a CNAME record for the production schematic deployment

### DIFF
--- a/org-formation/800-redirects/_tasks.yaml
+++ b/org-formation/800-redirects/_tasks.yaml
@@ -22,7 +22,7 @@ SsoRedirect:   # Redirect sso.sageit.org to the AWS SSO Start URL
     # and a path to our specific config
     TargetKey: "start#/"
 
-FinopsApiRedirect:  # redirect finops-api.sageit.org to lambda-mips-api cloudfront distribution
+FinopsApiDnsForward:  # forward finops-api.sageit.org to lambda-mips-api cloudfront distribution
   Type: update-stacks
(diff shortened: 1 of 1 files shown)


org-formation/800-redirects/_tasks.yaml
```diff
@@ -97,3 +97,21 @@ SchematicDevAppRedirect:
     SourceHostedZoneId: !CopyValue [!Sub '${primaryRegion}-${resourcePrefix}-sagebio-api-zone-HostedZoneId']
     # the value of the CNAME record
     TargetHostName: !CopyValue ['schematic-dev-DockerFargateStack-LoadBalancerDNS', !Ref DnTDevAccount]
+
+# redirect schematic.api.sagebionetworks.org to schematic-infra ALB
+# https://github.com/Sage-Bionetworks/schematic-infra
+SchematicProdAppRedirect:
```
brucehoff, line 103: Sorry if this is picky, but could we please use a word other than 'redirect' (both here and on line 85, above)? To me "redirect" means something very specific, i.e., to return a 3xx HTTP status.   Two alternatives terms to consider are "Routing" and "Forwarding".

brucehoff (changes requested):

org-formation/800-redirects/_tasks.yaml
jesusaurus: I've renamed these from "Redirect" to "DnsForward"

brucehoff (approved):

PR #758, opened 2023-03-12
PLFM-7700: Add wafv2:* perms

xschildw:
This PR adds permissions to the deployment role. Note: this does not get deployed by infra (see PFLM-7725 to fix that).
The key rotation did not get deployed so I removed it to be in sync with deployed, will put it back after PLFM-7725 is done.


zaro0508 (approved):
hmm, why did EnableKeyRotation not get deployed?

xschildw:
I compared the deployed template to the one I had and there was a diff. Only explanation is that I did not deploy the change, the remedy is to integrate the 2 templates in infra (https://sagebionetworks.jira.com/browse/PLFM-7725).

PR #765, opened 2023-03-15
[IT-2692] Grant SSO access for SCCE dev account

jesusaurus:
PR Checklist:
[ ] Describe and explain your intentions for this change
[ ] Setup pre-commit and run the validators (info in README.md)


zaro0508 (approved):
LGTM.  I usually only create roles that we are going to actually add users to but creating them all up front is fine as well.

PR #775, opened 2023-03-22
[IT-2594] Rename host for lambda-mips-api

jesusaurus:
Prepare for having two finops lambdas under the same domain by creating a zone and wildcard cert for `finops.sageit.org`, replacing the secondary wildcard certificate for `sageit.org`.

### DIFF
--- a/org-formation/100-shared-dns/README.md
+++ b/org-formation/100-shared-dns/README.md
@@ -12,16 +12,12 @@ used by the hosting service.
 
 Each wildcard cert will need to be manually verified before it can be used.
 
-### sageit.org
+### finops.sageit.org
 
-This hosted zone was created manually and is shared between finops
-microservices and some auth redirects (SSO and VPN).
+This zone is intended for cost-tracking automation, such as lambda-mips-api.
 
-A wildcard certificate has been created for this zone in sceptre:
-./sceptre/sageit/config/prod/sageit-org-acm-cert.yaml
-
-This directory is used for creating additional certificates in other
-accounts and their associated DNS records.
+The wildcard certificate created for this zone is expected to be used elsewhere
+in org-formation, but not by external infra projects (such as CDK applications).
 
 ### app.sagebionetworks.org
 
--- a/org-formation/100-shared-dns/_tasks.yaml
+++ b/org-formation/100-shared-dns/_tasks.yaml
@@ -1,6 +1,24 @@
 Parameters:
   <<: !Include '../_parameters.yaml'
 
+# The dns zone for this certificate was created manually, and a
+# certificate for the zone is created in the sageit account in sceptre:
+# ./sceptre/sageit/config/prod/sageit-org-acm-cert.yaml
+#
+# Create a second sageit certificate in admincentral for the finops lambdas
+# This certificate is deprecated by the finops wildcard certificate below
+SageItCertificate:
+  Type: update-stacks
+  Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.6.8/templates/acm-certificate.yaml
+  StackName: !Sub '${resourcePrefix}-sageit-cert'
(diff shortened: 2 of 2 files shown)


org-formation/100-shared-dns/_tasks.yaml
```diff
@@ -1,24 +1,33 @@
...
+    Account: !Ref SageITAccount
+  Parameters:
+    DnsDomainName: "finops.sageit.org"
+
+# Wildcard certificate for '*.finops.sageit.org'
+SageITFinopsCertificate:
   Type: update-stacks
   Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.6.8/templates/acm-certificate.yaml
-  StackName: !Sub '${resourcePrefix}-sageit-cert'
+  StackName: !Sub '${resourcePrefix}-sageit-finops-cert'
```
zaro0508, line 20: we need to be very careful here.  a rename will make org-formation delete an existing stack.  We don't want to delete the existing stack because there are dependencies on it.

The better process is to add new resources then carefully deprecate existing resources.

org-formation/100-shared-dns/README.md
```diff
@@ -12,16 +12,12 @@ used by the hosting service.
 
 Each wildcard cert will need to be manually verified before it can be used.
 
-### sageit.org
+### finops.sageit.org
```
zaro0508, line 15: What is the motivation behind this change?  there's nothing in the referenced jira issue that describes why your doing this.

We still need `sageit.org` because we have `vpn.sageit.org`, `sc.sageit.org`, etc..

zaro0508 (changes requested):

org-formation/100-shared-dns/README.md
jesusaurus: We will still have both a `sageit.org` zone and wildcard certificate in the sageit account, this is only removing the use of the `sageit.org` zone in the admincentral account by removing the wildcard certificate from that account.

The motivation is to have friendly names for both finops lambdas, I had the wrong ticket in the subject but I've updated it to IT-2594

org-formation/100-shared-dns/_tasks.yaml
jesusaurus: fortunately the only dependency on this is the mips-api lambda, but since it's managed earlier than this stack (in `050-costs`) you're right we need to wait until after we've migrated the lambda to the new service before we can remove deprecated cert

zaro0508: as a separate matter, i'm wondering if would make sense to move setup of shared-dns to before setup of costs?

org-formation/100-shared-dns/_tasks.yaml
```diff
@@ -18,7 +19,33 @@ SageItCertificate:
   Parameters:
     DnsDomainName: 'sageit.org'
 
-SagebioAppZone:  # Zone for application front-end DNS names, e.g. 'dca.app.sagebionetworks.org'
+# Zone for finops lambdas in admincentral
+SageITFinopsZone:
+  Type: update-stacks
+  Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.6.8/templates/R53-hostedzone.yaml
+  StackName: !Sub '${resourcePrefix}-sageit-finops-zone'
+  StackDescription: Create a shared hosted zone for application domains
```
zaro0508, line 27: could we be more specific with this description?

zaro0508 (approved):
generally LGTM, just need to update description and i had a question

org-formation/100-shared-dns/_tasks.yaml
jesusaurus: yeah, I regret the ordering of a lambda in `050-costs` and the dns and certificates in `100-shared-dns`, but I'm not sure which would be better to move

PR #782, opened 2023-03-24
IT-2735: OIDC for service-catalog-library repo

xschildw:
This PR sets OIDC for the service-catalog-repo in AdminCentral account

### DIFF
--- a/org-formation/650-identity-providers/_tasks.yaml
+++ b/org-formation/650-identity-providers/_tasks.yaml
@@ -207,3 +207,37 @@ GithubOidcSageBionetworksScicompProvisioner:
   DefaultOrganizationBinding:
     Account: !Ref ScicompAccount
     Region: us-east-1
+
+GithubOidcSageBionetworksSvcCatalogLibray:
+  Type: update-stacks
+  DependsOn: GithubOidcSageBionetworks
+  Template: github-oidc-provider-access.njk
+  StackName: !Sub ${resourcePrefix}-${appName}-sage-bionetworks-svc-catalog-library
+  Parameters:
+    ProviderArn: !CopyValue [ !Sub '${resourcePrefix}-${appName}-ProviderArn' ]
+    PolicyDocument: |
+      {
+        "Version": "2012-10-17",
+        "Statement": [
+          {
+              "Sid": "ListObjectsInBucket",
+              "Effect": "Allow",
+              "Action": [ "s3:GetBucketLocation", "s3:ListBucket" ],
+              "Resource": [ "arn:aws:s3:::bootstrap-awss3cloudformationbucket-19qromfd235z9" ]
+          },
+          {
+              "Sid": "AllObjectActions",
+              "Effect": "Allow",
+              "Action": [ "s3:PutObject", "s3:GetObject" ],
+              "Resource": ["arn:aws:s3:::bootstrap-awss3cloudformationbucket-19qromfd235z9/*"]
+          }
+        ]
+      }
+  TemplatingContext:
+    GitHubOrg: "Sage-Bionetworks"
+    Repositories:
+      - name: "service-catalog-library"
+        branch: "master"
+  DefaultOrganizationBinding:
+    Account: !Ref AdminCentralAccount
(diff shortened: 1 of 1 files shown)


jesusaurus (approved):

org-formation/650-identity-providers/_tasks.yaml
```diff
@@ -207,3 +207,37 @@ GithubOidcSageBionetworksScicompProvisioner:
   DefaultOrganizationBinding:
     Account: !Ref ScicompAccount
     Region: us-east-1
+
+GithubOidcSageBionetworksSvcCatalogLibrary:
+  Type: update-stacks
+  DependsOn: GithubOidcSageBionetworks
+  Template: github-oidc-provider-access.njk
+  StackName: !Sub ${resourcePrefix}-${appName}-sage-bionetworks-svc-catalog-library
```
zaro0508, line 215: not a big deal, but i think i prefer the stack name to contain the exact repo name, so i think `!Sub ${resourcePrefix}-${appName}-sage-bionetworks-service-catalog-library` would be better.

org-formation/650-identity-providers/_tasks.yaml
```diff
@@ -212,7 +212,7 @@ GithubOidcSageBionetworksSvcCatalogLibrary:
   Type: update-stacks
   DependsOn: GithubOidcSageBionetworks
   Template: github-oidc-provider-access.njk
-  StackName: !Sub ${resourcePrefix}-${appName}-sage-bionetworks-svc-catalog-library
+  StackName: !Sub ${resourcePrefix}-${appName}-sage-bionetworks-service-catalog-library
```
zaro0508, line 215: if you are changing `svc` to `service` then please change it in all of the other locations.

zaro0508 (changes requested):

org-formation/650-identity-providers/_tasks.yaml
```diff
@@ -207,3 +207,37 @@ GithubOidcSageBionetworksScicompProvisioner:
   DefaultOrganizationBinding:
     Account: !Ref ScicompAccount
     Region: us-east-1
+
+GithubOidcSageBionetworksServiceCatalogLibrary:
```
zaro0508, line 211: hmm, this is exactly the same as `GithubOidcSageBionetworksAwsInfra` above, except for the repo in the `Repositories` param.   Maybe we can just create one 
 `GithubOidcSharedCloudformationS3Bucket` resource and pass multiple repos to the  `Repositories` parameter?  it would be something like this..

```
  TemplatingContext:
    GitHubOrg: "Sage-Bionetworks"
    Repositories:
      - name: "aws-infra"
        branch: "master"
      - name: "service-catalog-library"
        branch: "master"
  DefaultOrganizationBinding:
    Account: !Ref AdminCentralAccount
    Region: us-east-1
```
I'm not sure that would work because I haven't tested it yet.  I'll test and let you know.

xschildw: That'd be nice! Would we not have to relax the policy a bit ('arn:aws:s3:::bootstrap-awss3cloudformationbucket-19qromfd235z9/service-catalog-library/*' -> 'arn:aws:s3:::bootstrap-awss3cloudformationbucket-19qromfd235z9/*' (at the risk of overwriting another repos artifacts)?

zaro0508: I ran a few tests.  Currently the `github-oidc-provider-access.njk` file does not support multiple repositories.  I attempted to change the nunjucks template to add support for multiple repos however I was not able to get it to work. i can't tell if it's something with nunjucks itself or the integration between org formation and nunjucks but something is not working.  I've create issue IT-2768 to try another time.  For now let's just move forward with one of these configs per repo.

zaro0508 (approved):

org-formation/650-identity-providers/_tasks.yaml
xschildw: OK. Thanks!

PR #789, opened 2023-03-28
IT-2752: Need service-user instead of service-account

xschildw:
The previous attempt (#773, #774) created a user, a role and the permission for the user to assume the role. For Redash, the permissions need to be directly on the user. This PR uses the service-user.yaml template which does just that...


zaro0508 (commented):
This seems fine but i'm wondering why you need a user instead of a role?  is it because the app that sets up redash uses a IAM user account instead of a role?

xschildw:
Yes that's it, the doc says the permissions are on the user itself...

zaro0508 (approved):

PR #790, opened 2023-03-29
[IT-2748] Create org-sagebase-cnb-dev account

jesusaurus:
Create a development account for challenges and benchmarks.

### DIFF
--- a/org-formation/organization.yaml
+++ b/org-formation/organization.yaml
@@ -33,6 +33,7 @@ Organization:
         - !Ref DnTDevAccount
         - !Ref RecoverDevAccount
         - !Ref ScceDevAccount
+        - !Ref CnbDevAccount
 
   ScienceProdOU:
     Type: OC::ORG::OrganizationalUnit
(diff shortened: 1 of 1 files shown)


org-formation/organization.yaml
```diff
@@ -363,6 +364,21 @@ Organization:
...
+      RootEmail: aws-cnb-dev@sagebase.org
+      Alias: org-sagebase-cnb-dev
+      Tags:
+        <<: !Include ./_default_org_tags.yaml
+        Department: CNB
+        Project: Challenges
+        AccountOwner: thomas.schaffter@sagebionetworks.org
+        CostCenter: NIH-ITCR / 101600
+        budget-alarm-threshold: 10000
+        budget-alarm-threshold-email-recipient: aws-cnb-dev@sagebase.org
```
tschaffter, line 380: I just checked on Google Groups and it looks like Joni is monitoring budget alarms. Thanks!

tschaffter (approved):

org-formation/organization.yaml
brucehoff: It shouldn't be an IT member who monitors budget alarms but rather the account owner, in this case someone from "CnB".

jesusaurus: @tschaffter I'm the current owner of the google group until the AWS account is created and I can set its root password. Once that's done, I'll transfer ownership of the google group to you as the account owner.

zaro0508 (approved):

PR #797, opened 2023-03-31
IT-2767:  Adding budget for RECOVER Service Catalog users

brucehoff:
Adding budget for RECOVER Service Catalog users
Larsson Omberg will be the 'community manager' and the budget will be $500/year.

### DIFF
--- a/sceptre/scipool/templates/prod/lambda-budgets/budget-rules.yaml
+++ b/sceptre/scipool/templates/prod/lambda-budgets/budget-rules.yaml
@@ -1,13 +1,13 @@
 # budget rules per user by synapse team id
 teams:
-  - '3469335': # Schwannomatosis Open Research Collaborative Service Catalog Users
+  '3469335': # Schwannomatosis Open Research Collaborative Service Catalog Users
     amount: '750'
     period: MONTHLY
     unit: USD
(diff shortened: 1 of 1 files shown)


sceptre/scipool/templates/prod/lambda-budgets/budget-rules.yaml
```diff
@@ -1,9 +1,15 @@
 # budget rules per user by synapse team id
 teams:
-  '3469335': # Schwannomatosis Open Research Collaborative Service Catalog Users
+  - '3469335': # Schwannomatosis Open Research Collaborative Service Catalog Users
```
jesusaurus, line 3: I think this should be a dict with a new key, not converted to a list. https://github.com/Sage-Bionetworks-IT/lambda-budgets/blob/master/budget/config.py#L9

jesusaurus (changes requested):

sceptre/scipool/templates/prod/lambda-budgets/budget-rules.yaml
brucehoff: Good catch, thanks.

jesusaurus (approved):

PR #799, opened 2023-04-04
IT-2651: Test key access by infra

xschildw:
This is a test PR to create a key in the SynapseDev account that should be accessible by the old AWSIAMAdminRole (to be deleted), myself logged in as SSO and the infra (the deployment should fail if the deployer would not be able to manage the key). If it works, the changes will be merged back in the original script used to deploy the Synapse master key, and that script will be added to the infra

### DIFF
--- a/sceptre/synapsedev/templates/testCMK.json
+++ b/sceptre/synapsedev/templates/testCMK.json
@@ -1,0 +1,104 @@
+{
+    "AWSTemplateFormatVersion": "2010-09-09",
+    "Description": "The master encryption key used to encypt/decypt all Synapse master secrets",
+    "Parameters": {
+        "Stack": {
+            "Description": "The stack",
+            "Type": "String"
+        }
+    },
+    "Resources": {
+        "SynapseCMK": {
+            "Type": "AWS::KMS::Key",
+            "Properties": {
+                "Description": "A master encryption key",
+                "EnableKeyRotation": false,
+                "KeyPolicy": {
+                    "Version": "2012-10-17",
+                    "Id": "key-default-1",
+                    "Statement": [
+                        {
+                            "Sid": "Allow administration of the key",
+                            "Effect": "Deny",
+                            "Principal": {
+                                "AWS": "*"
+                            },
+                            "Action": [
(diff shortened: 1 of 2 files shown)


sceptre/synapsedev/templates/testCMK.json
```diff
@@ -0,0 +1,104 @@
...
+            "Type": "AWS::KMS::Key",
+            "Properties": {
+                "Description": "A master encryption key",
+                "EnableKeyRotation": false,
+                "KeyPolicy": {
+                    "Version": "2012-10-17",
+                    "Id": "key-default-1",
+                    "Statement": [
+                        {
+                            "Sid": "Allow administration of the key",
```
zaro0508, line 21: isn't this a deny policy?

sceptre/synapsedev/templates/testCMK.json
```diff
@@ -0,0 +1,104 @@
...
+                                                "arn:aws:iam::",
+                                                {
+                                                    "Ref": "AWS::AccountId"
+                                                },
+                                                ":root"
+                                            ]
+                                        ]
+                                    },
+                                    "arn:aws:sts::449435941126:assumed-role/AWSReservedSSO_Administrator_693a85eb20cd5043/x.schildwachter@sagebase.org",
+                                    "arn:aws:iam::449435941126:role/sagebase-github-oidc-sage-ProviderRoleorganization-1K1MGUYPL5JUO"
```
zaro0508, line 71: you just denied all except for root in the previous policy.  deny overrides allow therefore I don't think this will allow these principals.

zaro0508 (commented):
i'm fine if you want to merge and try this change but i don't think this is gonna work.  Also I think testing something like this would be much easier and faster if you just do it manually in the AWS console.

sceptre/synapsedev/templates/testCMK.json
xschildw: At line 45, any role should be excepted as well. It worked for SSO, I want to see if it's gonna work for oidc.

zaro0508 (approved):

PR #803, opened 2023-04-06
PLFM-7774: Add synapsedev-specific admin group

xschildw:
This PR creates an AWS admin group in the SynapseDev account (currently we have an admin group that spans the OU). This is to be used by devs in the dev account as we discover which permissions should be added to the developer group.
Also reduced the target of the existing group to SynapseProd only (should rename/replace later for clarity).

### DIFF
--- a/org-formation/700-aws-sso/_tasks.yaml
+++ b/org-formation/700-aws-sso/_tasks.yaml
@@ -77,6 +77,10 @@ Parameters:
     Type: String
     Default: '906769aa66-1c66a9b6-ae5e-473c-a51c-4b02918c4454'
 
+  synapseDevAdminGroup:   #JC aws-synapsedev-admins
+    Type: String
+    Default: '44d844b8-60a1-70e5-cb5a-50137c984398'
+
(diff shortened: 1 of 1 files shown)


org-formation/700-aws-sso/_tasks.yaml
```diff
@@ -754,12 +758,30 @@ SsoSynapseAdmin:
   OrganizationBindings:
     TargetBinding:
       OrganizationalUnit:
-        - !Ref SynapseOU
+        - !Ref SynapseProdAccount
```
zaro0508, line 761: I think we should probably replace the above `synapseAdminGroup:      #JC aws-synapse-admins` with a more specific `synapseProdAdminGroup`?  it can be done in a separate PR but I just wanted to see if that was in your plans.

zaro0508 (approved):
This generally looks fine, just had a question regarding separation of groups.

org-formation/700-aws-sso/_tasks.yaml
xschildw: Yes, that's the plan.

PR #805, opened 2023-04-07
PLFM-7774: Followup - create synapseprod-specific admin group

xschildw:
This PR is a followup to #803, remove the existing shared synapse admin group and replace it with one specific to synapseprod.

### DIFF
--- a/org-formation/700-aws-sso/_tasks.yaml
+++ b/org-formation/700-aws-sso/_tasks.yaml
@@ -81,6 +81,10 @@ Parameters:
     Type: String
     Default: '44d844b8-60a1-70e5-cb5a-50137c984398'
 
+  synapseProdAdminGroup:  #JC aws-synapseprod-admins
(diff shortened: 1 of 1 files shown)


org-formation/700-aws-sso/_tasks.yaml
```diff
@@ -73,14 +73,14 @@ Parameters:
     Type: String
     Default: '906769aa66-6cdd2cc1-6d3d-49d8-b105-37c22bb3d402'
 
-  synapseAdminGroup:      #JC aws-synapse-admins
```
zaro0508, line 76: please make sure this is removed from JC

zaro0508 (approved):

org-formation/700-aws-sso/_tasks.yaml
xschildw: The user group is deleted in JC.

PR #808, opened 2023-04-10
[IT-1655] Move source of Other value to sceptre

jesusaurus:
In order to avoid listing "Other / 000001" as a valid program code in the list we provide to users, stop configuring the lambda to add the code, and instead add it in the nunjucks template generating Tag Options for each valid tag.

### DIFF
--- a/sceptre/scipool/config/develop/sc-tag-options.yaml
+++ b/sceptre/scipool/config/develop/sc-tag-options.yaml
@@ -14,7 +14,7 @@ dependencies:
 sceptre_user_data:
   Departments: !file sc-tag-options/internal/Departments.json
   Projects: !file sc-tag-options/internal/Projects.json
-  CostCenters: !file https://{{stack_group_config.admincentral_cf_bucket}}.s3.amazonaws.com/aws-infra/v0.3.6/tags/SageFinancialProgramCodes.json
+  CostCenters: !rcmd 'wget -qO- https://mips-api.finops.sageit.org/tags?limit=24'
   ProductIDs:
     - !stack_output_external sc-portfolio-ec2::SCEC2portfolioId
     - !stack_output_external sc-portfolio-s3-basic::SCS3portfolioId
--- a/sceptre/scipool/templates/sc-tag-options.j2
+++ b/sceptre/scipool/templates/sc-tag-options.j2
@@ -30,6 +30,12 @@ Resources:
       Key: "CostCenter"
       Value: {{ CostCenter }}
   {% endfor %}
+  OtherCostCenterTag:
+    Type: AWS::ServiceCatalog::TagOption
+    Properties:
+      Active: true
+      Key: "CostCenter"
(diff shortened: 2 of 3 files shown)


zaro0508 (changes requested):
I don't understand the idea around this PR.  Could you please provide more explanation?

jesusaurus:
@zaro0508 the idea here is to hard-code the Other value in the nunjucks template for the Tag Options instead of configuring the lambda to inject the value so that it's not listed in the lambda tag list but still shows up in the drop-down menu

sceptre/scipool/config/develop/sc-tag-options.yaml
```diff
@@ -14,7 +14,7 @@ dependencies:
 sceptre_user_data:
   Departments: !file sc-tag-options/internal/Departments.json
   Projects: !file sc-tag-options/internal/Projects.json
-  CostCenters: !file https://{{stack_group_config.admincentral_cf_bucket}}.s3.amazonaws.com/aws-infra/v0.3.6/tags/SageFinancialProgramCodes.json
+  CostCenters: !rcmd 'wget -qO- https://mips-api.finops.sageit.org/tags?limit=24'
```
zaro0508, line 17: i don't think this is going to work because `!rcmd` will pass the response into `CostCenter` as a string. We are expecting CostCenter to be a list in templates/sc-tag-options.j2

sceptre/scipool/templates/sc-tag-options.j2
```diff
@@ -30,6 +30,12 @@ Resources:
       Key: "CostCenter"
       Value: {{ CostCenter }}
   {% endfor %}
+  OtherCostCenterTag:
+    Type: AWS::ServiceCatalog::TagOption
+    Properties:
+      Active: true
+      Key: "CostCenter"
+      Value: "Other / 000001"
```
zaro0508, line 38: this seems like a bad idea to me.  i think all id/codes should come from a central source of truth.  this effectively breaks that pattern.  What if we change the ID name or number?

zaro0508 (changes requested):

sceptre/scipool/config/develop/sc-tag-options.yaml
jesusaurus: Oh, I didn't realize the file resolver was doing implicit deserialization. hrm.

sceptre/scipool/templates/sc-tag-options.j2
jesusaurus: the "other" value is only used by service catalog as a place holder. now that i've split the two lambdas apart i was thinking about what needs the "other" value: we don't need it for the cost rules lambda because we don't want an "other" cost category, and we don't want it in the list we show to users either. but i see your point about the source of truth. i guess i'll need to modify the lambda to treat "other" and "no program" differently since we do want "no program" in the other cases too.

PR #809, opened 2023-04-11
[IT-2460] Migrate from custom transform to lambda 

jesusaurus:
Switch from using a custom Transform to calling a lambda with
an org-formation function.

depends on: #807

### DIFF
--- a/org-formation/050-costs/_tasks.yaml
+++ b/org-formation/050-costs/_tasks.yaml
@@ -61,3 +61,5 @@ CostCategories:
   DefaultOrganizationBinding:
     Account: !Ref MasterAccount
     Region: !Ref primaryRegion
+  Parameters:
+    ProgramCodeRulesJSON: !Cmd 'wget -qO- https://cost-rules.finops.sageit.org/program-codes'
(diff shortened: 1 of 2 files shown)


org-formation/050-costs/_tasks.yaml
```diff
@@ -61,3 +61,6 @@ CostCategories:
   DefaultOrganizationBinding:
     Account: !Ref MasterAccount
     Region: !Ref primaryRegion
+  Parameters:
+    OwnerEmailRulesJSON: !Cmd 'wget -qO- https://cost-rules.finops.sageit.org/owner-emails'
```
zaro0508, line 65: how about using the [sceptre-request-resolver](https://pypi.org/project/sceptre-request-resolver) instead?

also i get `name 'boto3' is not defined` when attempting to access that url?

org-formation/050-costs/_tasks.yaml
```diff
@@ -61,3 +61,6 @@ CostCategories:
...
+    ProgramCodeRulesJSON: !Cmd 'wget -qO- https://cost-rules.finops.sageit.org/program-codes'
```
zaro0508, line 66: same here

zaro0508 (changes requested):

org-formation/050-costs/_tasks.yaml
jesusaurus: I've resolved the missing import for boto3, and plan on working on the !Request function as follow-up work

zaro0508 (approved):

PR #813, opened 2023-04-12
[IT-2750] Move CDK admin instructions to org-formation

jesusaurus:
Migrate some instructions for CDK certificate and DNS setup from confluence to github.

### DIFF
--- a/org-formation/100-shared-dns/README.md
+++ b/org-formation/100-shared-dns/README.md
@@ -39,10 +39,11 @@ hosted zone with a value pointing to the AWS domain name of the ALB.
 
 Certificates must exist in the same account they will be used in, and so we
 must create a copy of our wildcard certificate in an application account before
-the application can be deployed to that account. For example, [the list of
-accounts for `app.sagebionetworks.org`](https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/org-formation/100-shared-dns/_tasks.yaml#L70).
-When merging a change that creates new certificates, the deploy pipeline will
-fail if the certificates are not validated within the 30-minute timeout.
+the application can be deployed to that account. Ensure that the stack creating
+the desired wildcard certificate has an account binding for the application
+account in `_tasks.yaml`. When merging a change that creates new certificates,
+the deploy pipeline will fail if the certificates are not manually validated
+within the 30-minute timeout.
 
 ### Validating New Certificates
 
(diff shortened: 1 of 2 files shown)


org-formation/100-shared-dns/README.md
```diff
@@ -34,3 +34,35 @@ This zone is intended for application-facing applications, such as Schematic.
...
 name for the created ALB. Create a DNS CNAME record in the appropriate
 hosted zone with a value pointing to the AWS domain name of the ALB.
+
+## Per-Account Certificates
+
+Certificates must exist in the same account they will be used in, and so we
+must create a copy of our wildcard certificate in an application account before
+the application can be deployed to that account. For example, [the list of
+accounts for `app.sagebionetworks.org`](https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/org-formation/100-shared-dns/_tasks.yaml#L70).
+
```
brucehoff, line 44: Need to clarify that the GitHub workflow will fail, pending validation, below.

brucehoff (changes requested):

brucehoff (approved):

org-formation/100-shared-dns/README.md
```diff
@@ -34,3 +34,37 @@ This zone is intended for application-facing applications, such as Schematic.
 CDK applications will export a CloudFormation output with the AWS domain
 name for the created ALB. Create a DNS CNAME record in the appropriate
 hosted zone with a value pointing to the AWS domain name of the ALB.
+
+## Per-Account Certificates
+
+Certificates must exist in the same account they will be used in, and so we
+must create a copy of our wildcard certificate in an application account before
+the application can be deployed to that account. For example, [the list of
+accounts for `app.sagebionetworks.org`](https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/org-formation/100-shared-dns/_tasks.yaml#L70).
```
zaro0508, line 43: i'm not a fan of specific line references in docs because it will very likely point to something else at a later time.  how about we just reference the entire file instead?

zaro0508 (changes requested):

org-formation/100-shared-dns/README.md
jesusaurus: in that case I don't think it makes sense to have a link, I'll just quote the name of the only other file in this directory

zaro0508 (approved):

PR #834, opened 2023-04-21
[IT-2703] Remove Project and Department tags

jesusaurus:
Remove the Project and Department tags from Service Catalog, we only use the CostCenter tag for cost tracking.


sceptre/scipool/config/bmgfki/sc-tag-options-external.yaml
```diff
@@ -9,7 +9,5 @@ dependencies:
   - "bmgfki/sc-portfolio-ec2-external.yaml"
   - "bmgfki/sc-portfolio-s3-basic.yaml"
 sceptre_user_data:
-  Departments: !file sc-tag-options/external/Departments.json
```
zaro0508, line 12: hmm, we use project and department to identify resources belonging to [external SC users](https://github.com/Sage-Bionetworks/synapse-login-aws-infra/blob/develop/config/prod/synapse-login-scipoolprod-params.yaml#L28).
Screen Shot 2023-04-21 at 10 17 12 AM

Before removing we probably need a different way to identify resources belonging to external users.

zaro0508 (changes requested):

sceptre/scipool/config/bmgfki/sc-tag-options-external.yaml
jesusaurus: hm, good catch, this will take some thinking, but I'm wondering if we can do something similar to the synapse tagger lambda.

jesusaurus: after discussing in the team meeting, we decided it's okay to remove these without a replacement tag

zaro0508: @jesusaurus I don't see explanation in the issue regarding the rational behind the this decision.  Would you be able to explain?

jesusaurus: @zaro0508 yeah, the short version is that these tags aren't used for tracking ownership. in the case of external users the tags are set to `collaborator` and `sage-external` like in your screenshot, and the tag that has useful ownership info is `aws:servicecatalog:provisioningPrincipalArn`.

jesusaurus: @zaro0508 we can use `aws:servicecatalog:provisioningPrincipalArn` to get the synapse user id of the user that created the resource, but why do we need to know if the synapse user is sage or non-sage?

zaro0508: I remembered that we wanted to know the cost breakdown between sage and non-sage usage of the service catalog. Tag was the way we were doing it.   Do we still have a way to do this after removing these tags?

jesusaurus: I think we can still use `aws:servicecatalog:provisioningPrincipalArn` which will have either `ServiceCatalogEndusers` or `ServiceCatalogExternalEndusers` in it. I also don't think the Project and Department tags are the appropriate place for that info, if we need a specific tag to track that we can add a new one like `SageExternal:true` or something

zaro0508 (approved):

PR #842, opened 2023-04-26
IT-2736: Add repo to existing role?

xschildw:
#841 failed because the provider was already in another stack (the org-infra one), try to add the repo to the list of repos on that stack (though I think we tried that with orgformation and it did not work...).


jesusaurus:
This depends on IT-2768

PR #859, opened 2023-05-08
[IT-2768] make OIDC template support multi repos

zaro0508:
Our current GH OIDC template only sets up GH OIDC for a single repository in an GH org.  This causes lots of duplication in the orgf _task.yaml file.  This update will allow setting up OIDC for multiple repos and branches which will allow us to reduce the number of entries in the _task.yaml file.

### DIFF
--- a/org-formation/650-identity-providers/_tasks.yaml
+++ b/org-formation/650-identity-providers/_tasks.yaml
@@ -423,7 +423,7 @@ GithubOidcSciPoolDevSageBionetworksSynapseLoginAwsInfra:
   StackName: !Sub ${resourcePrefix}-${appName}-scipool-dev-synapse-login-aws-infra
   Parameters:
     ProviderArn: !CopyValue [ !Sub '${resourcePrefix}-${appName}-ProviderArn' ]
-    ProviderRoleName: !Sub ${resourcePrefix}-${appName}-scipool-dev-sage-bionetworks-synapse-login-aws-infra
+    ProviderRoleName: !Sub ${resourcePrefix}-${appName}-scipool-dev-synapse-login-aws-infra
     ManagedPolicyArns:
       - "arn:aws:iam::aws:policy/AdministratorAccess"
   TemplatingContext:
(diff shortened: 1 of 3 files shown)


xschildw (approved):

org-formation/650-identity-providers/github-oidc-provider-access.njk
```diff
@@ -66,10 +66,10 @@ Resources:
     Type: 'AWS::IAM::ManagedPolicy'
     Properties:
       PolicyDocument: !Ref PolicyDocument
-{% for Repository in Repositories %}
-  ProviderRole{{ Repository.name | replace('-','') | replace('_','') }}:
+  ProviderRole:
```
zaro0508, line 69: This change will cause this role to be renamed.

org-formation/650-identity-providers/github-oidc-provider-access.njk
```diff
@@ -66,10 +66,10 @@ Resources:
...
     Type: AWS::IAM::Role
     Properties:
+      RoleName: !Sub '${AWS::StackName}-ProviderRole'
```
zaro0508, line 72: Set to a specific role name so that we can create dependent PRs to assume
this role for CI purposes.  This will reduce the downtime from deploying
this change.

jesusaurus: The role name needs to be unique to the account, so this means that we can only deploy this template once per account, right? Should we also add the repository name to the role name?

jesusaurus (changes requested):

org-formation/650-identity-providers/github-oidc-provider-access.njk
xschildw: I think it should just update the policy document each time we run.

zaro0508: @jesusaurus yes, role names are unique per account.  CFN stack names are also unique.  You can deploy multiple instances as long as the stack name is different.

jesusaurus (approved):

xschildw (approved):

org-formation/650-identity-providers/github-oidc-provider-access.njk
zaro0508: This won't work because IAM role names can only be 64 chars max.  CFN deployment will fail if the name is set to something longer than the max length.  Some of the stack names defined in [650-identity-provider/_task.yaml](https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/org-formation/650-identity-providers/_tasks.yaml) will make the RoleName longer than 64 chars.

org-formation/650-identity-providers/_tasks.yaml
```diff
@@ -222,13 +232,14 @@ GithubOidcSageBionetworksDccValidatorPECInfra:
   StackName: !Sub ${resourcePrefix}-${appName}-sage-bionetworks-dccvalidator-pec-infra
   Parameters:
     ProviderArn: !CopyValue [ !Sub '${resourcePrefix}-${appName}-ProviderArn' ]
+    ProviderRoleName: !Sub ${resourcePrefix}-${appName}-sage-bionetworks-dccvalidator-1kD-infra
```
xschildw, line 235: Should it not be pec-infra?

xschildw: Or can't these 4 repos in dccvalidatordev account be combined ?

PR #860, opened 2023-05-08
IT-2651: Infra takes over deployment of synapse-dev-cmk stack

xschildw:
The synapse-dev-cmk stack sets up the deployer role assumed by the Jenkins system in the dev account, it also sets up the root encryption key used to encrypt the secrets (in SecretMgr) used by the Synapse Stack builder. The key was managed by an IAM role (to be deleted), the policy has been modified to give access to users with admin SSO access, to root (as before), to the deployment role (as before) and to the OIDC role (so we can deploy using infra).
This template was run already from the IAM role to give permissions to SSO/OIDC.

This PR is just to setup the deployment in the infra (Cloudformation should not detect any change).

### DIFF
--- a/sceptre/synapsedev/templates/SynapseCMK-template.json
+++ b/sceptre/synapsedev/templates/SynapseCMK-template.json
@@ -258,21 +258,9 @@
                             "Condition": {
                                 "StringNotLike": {
                                     "aws:PrincipalArn": [
-                                        {
-                                            "Fn::Join": [
-                                                "",
-                                                [
-                                                    "arn:aws:iam::",
-                                                    {
-                                                        "Ref": "AWS::AccountId"
-                                                    },
-                                                    ":root"
-                                                ]
-                                            ]
-                                        },
-                                        "arn:aws:iam::449435941126:role*",
-                                        "arn:aws:sts::449435941126:assumed-role*"
-
+                                        { "Fn::Sub": "arn:aws:iam::${AWS::AccountId}:root" },
+                                        { "Fn::Sub": "arn:aws:iam::${AWS::AccountId}:role*" },
+                                        { "Fn::Sub": "arn:aws:iam::${AWS::AccountId}:assumed-role*" }
                                     ]
                                 }
                             }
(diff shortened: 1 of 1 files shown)


sceptre/synapsedev/templates/SynapseCMK-template.json
```diff
@@ -335,12 +303,12 @@
                                             ]
                                         ]
                                     },
-                                    {
-                                        "Fn::ImportValue": "us-east-1-jumpcloud-idp-SynapseDevDeveloperSamlProviderRoleArn"
-                                    },
-                                    {
-                                        "Fn::ImportValue": "us-east-1-jumpcloud-idp-SynapseDevAdminSamlProviderRoleArn"
-                                    }
+                                    "arn:aws:sts::449435941126:assumed-role/AWSReservedSSO_Administrator_693a85eb20cd5043/x.schildwachter@sagebase.org",
```
zaro0508, line 306: instead of listing every user i believe you should be able to give the admin group access to this key.  The [documentation](https://docs.aws.amazon.com/kms/latest/developerguide/key-policy-overview.html) says..
```
(Required) The principal is the identity that gets the permissions specified in the policy statement.
You can specify AWS accounts, IAM users, IAM roles, and some AWS services as principals in a key
policy. IAM user groups are not a valid principal in any policy type.
```

I think that means you can give the SSO admin role access to this key, so probably all you need here is `arn:aws:sts::449435941126:assumed-role/AWSReservedSSO_Administrator_693a85eb20cd5043`

zaro0508 (changes requested):

sceptre/synapsedev/templates/SynapseCMK-template.json
xschildw: Thx, I'll give it a try (I tried with '/*' but did not think about just the group...).

zaro0508 (commented):
have you tested this and does it work?

xschildw:
Replaced by #911

PR #865, opened 2023-05-12
[IT-2460] Use our lambda-generated cost rules

jesusaurus:
Replace our custom transform macro with auto-generated rules written to S3 so that we no longer need to manually manage code-deduplication rules or account mappings.


org-formation/050-costs/categories.yaml
```diff
@@ -28,151 +28,7 @@ Resources:
...
-              - CostCenterOther
-              - CostCenter
-            TagEndsWith:
-              - "000000"
-        InheritedValues:
-          TagOrder:
-            - CostCenterOther
-            - CostCenter
-          RulePosition: Last
+      'Fn::Transform':
```
zaro0508, line 31: hmm, i must have missed it but why did you include `RuleVersion` into program-code-rules.yaml?  Wouldn't it be better to keep them separate so that, if you ever choose, you can set them independently?

org-formation/050-costs/categories.yaml
```diff
@@ -28,151 +28,7 @@ Resources:
...
+        Name: 'AWS::Include'
+        Parameters:
+          Location: 's3://sagebase-rules-microservice-outputbucket-17vzcnk2xnq0/cost-categories/program-code-rules.yaml'
```
zaro0508, line 34: why not include owner-email-rules.yaml for `OwnerEmailCostCategory` resource as well?

org-formation/050-costs/categories.yaml
jesusaurus: i don't think it makes sense to ever set them separately, `RuleVersion` is setting the schema used by `Rules`

org-formation/050-costs/categories.yaml
jesusaurus: because it's generating invalid rules. we can create rules that inherit email addresses from tags, but trying to create account-based rules where the category name is an email address fails a regex constraint (regex details are in it-2369). i have an open aws support ticket but i'm not hopeful that the category name regex will be expanded to match the tag value regex any time soon :(

org-formation/050-costs/categories.yaml
zaro0508: There are other properties in `AWS::CE::CostCategory`.  This implementation  suggests that the lambda will provide all of the properties for the module correct? What if you wanted to configure other parameters for this CostCategory module?  Or what if parameters get added/removed from it?

org-formation/050-costs/categories.yaml
zaro0508: so i thought i saw some code in the lambda to supply OwnerEmailCostCategory?  what's the plan for that?  are we going to keep that code in the lambda and just not use it?

jesusaurus: yeah I would like to use it eventually, if AWS ever relaxes the category regex constraint to match the regex constraint for tags. so my plan is to leave the functionality in place with hopes of eventually being able to use it

org-formation/050-costs/categories.yaml
jesusaurus: no, my intention isn't to provide all of the properties. originally the lambda was only going to manage the value of the `Rules` property, but the `Rules` and `RuleVersion` properties are so tightly coupled that I decided to implement both. though I don't see a need to implement any other properties

and I'm not sure exactly what you mean by parameters getting added/removed, but if cloudformation makes a breaking change to a resource schema, then that would force us to update our templates no matter what.

org-formation/050-costs/categories.yaml
zaro0508: I'm not a fan of unused code dangling around.  if you must keep it in then could you either comment it out completely or at least provide a comment explaining that it's not used for whatever reason?

zaro0508 (changes requested):

org-formation/050-costs/categories.yaml
jesusaurus: No not necessarily, it depends on what the parameter does. If the parameter is tightly coupled to the rules generated then it would have to go in the lambda, but if it's an independent configuration I would prefer to set it in org-formation alongside where we set the category `Name`.

org-formation/050-costs/categories.yaml
jesusaurus: is it enough to comment out the lambda in the template? https://github.com/Sage-Bionetworks-IT/lambda-finops-cost-rules/pull/8

or do i need to remove `cost_emails/owner_emails.py` from the lambda entirely?

zaro0508: It's up to you.  It should either be completely removed or completely commented out with an explanation as to why it's not used.

org-formation/050-costs/categories.yaml
zaro0508: we should discuss, I think we might have a different understand of how this works.

zaro0508 (approved):

PR #870, opened 2023-05-15
[IT-2703] Remove invalid tag option templates

jesusaurus:
Now that we don't create Project or Department tags on our Service Catalog products, we only need to create Tag Options for products that set CostCenters. Remove the stack in cases where it is not set.

Currently, the jinja template is producing an invalid Cloudformation template when no CostCenters are set.


zaro0508:
oops, i didn't notice this PR.  My solution to this failure was to set the cost center tag to "No Program" in PR #871

zaro0508:
my solution in PR #871 didn't work due to error.. 
```
develop/sc-tag-options-external NOPROGRAM000000CostCenterTag 
AWS::ServiceCatalog::TagOption CREATE_FAILED CostCenter|NO PROGRAM / 000000 already exists in stack
arn:aws:cloudformation:us-east-1:465877038949:stack/sc-tag-options/f8245460-6dc8-11ec-a4e0-1289ac4902b3
```
We can't create duplicate tags for internal and external so let's go with this solution.

zaro0508 (approved):

PR #876, opened 2023-05-17
[IT-2838] Setup SSO access for personal aws account

zaro0508:
Allow Elias access to his personal aws account as an admin.

depends on #875

### DIFF
--- a/org-formation/700-aws-sso/_tasks.yaml
+++ b/org-formation/700-aws-sso/_tasks.yaml
@@ -1576,7 +1576,7 @@ SsoBuA2aDwAdmin:
   DependsOn: SsoAdministrator
   Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.2.11/templates/SSO/aws-sso.yaml
   StackName: !Sub '${resourcePrefix}-${appName}-8dxfh53-admin'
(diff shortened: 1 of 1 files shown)


org-formation/700-aws-sso/_tasks.yaml
```diff
@@ -1567,6 +1571,23 @@ Sso8dxfh53Admin:
     principalId: !Ref 8dxfh53AdminGroup
     permissionSetArn: !CopyValue [ !Sub '${resourcePrefix}-${appName}-admin-permission-set-arn' ]
 
+SsoBuA2aDwAdmin:
+  Type: update-stacks
+  DependsOn: SsoAdministrator
+  Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.2.11/templates/SSO/aws-sso.yaml
+  StackName: !Sub '${resourcePrefix}-${appName}-8dxfh53-admin'
```
jesusaurus, line 1578: The stack name needs to be updated with the new random string

jesusaurus (changes requested):
one missing update, otherwise it looks good

org-formation/700-aws-sso/_tasks.yaml
zaro0508: fixed

jesusaurus (approved):

PR #877, opened 2023-05-17
consolidate GH OIDC definitions

zaro0508:
aws-infra and service-catalog-library repos deploy to the same location.  We consolidate the definition to reduce duplication.

depends on #859

### DIFF
--- a/org-formation/650-identity-providers/_tasks.yaml
+++ b/org-formation/650-identity-providers/_tasks.yaml
@@ -321,14 +321,15 @@ GithubOidcSageBionetworksSecuityHubToJira:
       - !Ref SecurityCentralAccount
     Region: us-east-1
 
-GithubOidcSageBionetworksAwsInfra:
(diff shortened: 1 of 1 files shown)


xschildw (approved):

org-formation/650-identity-providers/_tasks.yaml
```diff
@@ -352,6 +353,8 @@ GithubOidcSageBionetworksAwsInfra:
     Repositories:
       - name: "aws-infra"
         branches: ["master"]
+      - name: "service-catalog-library"
+        branch: "master"
```
jesusaurus, line 357: this should be "branches" right?

jesusaurus (changes requested):

org-formation/650-identity-providers/_tasks.yaml
zaro0508: fixed.  good catch, thanks.

jesusaurus (approved):

PR #878, opened 2023-05-17
Consolidate GH OIDC defintions for DCC

zaro0508:
DCC apps all deploy to the same accounts so we can consolidate the definition to reduce duplication.

depends on #859


xschildw (approved):
Nice reduction of boilerplate

jesusaurus (approved):

PR #884, opened 2023-05-25
[IT-1532] Give SC users access to Cost Explorer

jesusaurus:
We need to give the Service Catalog users read-only access to the Cost Explorer before we start sending them emails to check their expenses in it.

### DIFF
--- a/sceptre/scipool/templates/sc-enduser-iam.yaml
+++ b/sceptre/scipool/templates/sc-enduser-iam.yaml
@@ -12,6 +12,7 @@ Resources:
         - arn:aws:iam::aws:policy/AWSServiceCatalogEndUserFullAccess
         - arn:aws:iam::aws:policy/CloudWatchLogsReadOnlyAccess
         - !Ref SCEnduserExpandedPolicy
+        - !Ref SCCostExplorerPolicy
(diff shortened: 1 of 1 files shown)


sceptre/scipool/templates/sc-enduser-iam.yaml
```diff
@@ -32,6 +33,7 @@ Resources:
       ManagedPolicyArns:
         - arn:aws:iam::aws:policy/AWSServiceCatalogEndUserFullAccess
         - !Ref SCEnduserExpandedPolicy
+        - !Ref SCCostExplorerPolicy
```
zaro0508, line 36: This role is for non sage staff users.  I'm not sure if should provide external users access to cost explorer?  Your thoughts @brucehoff ?

brucehoff: This is a issue that I haven't thought about:  The goal is to give Sage workers (not all Service Catalog users) the ability to track costs.   As @zaro0508  suggests, we should NOT be giving non-Sage workers access to Cost Explorer and, therefore we should not be sending them the same spend alerts.  Will elaborate in https://sagebionetworks.jira.com/browse/IT-1532

PR #888, opened 2023-06-06
Allow sage-bionetworks/synapserutils repo' to write documentation to [staging-]ran.sagebase.org buckets as part of build

thomasyu888:
PR Checklist:

This will allow the synapserutils repo to push the package build into [staging-]ran.sagebase.org

### DIFF
--- a/org-formation/650-identity-providers/_tasks.yaml
+++ b/org-formation/650-identity-providers/_tasks.yaml
@@ -469,9 +469,9 @@ GithubOidcSageBionetworksSynapseR:
     GitHubOrg: "Sage-Bionetworks"
     Repositories:
       - name: "synapser"
-        branches: ["*"]
+        branches: ["master", "main", "develop"]
       - name: "synapserutils"
-        branches: ["*"]
+        branches: ["master", "main", "develop"]
   DefaultOrganizationBinding:
     Account: !Ref SynapseProdAccount
     Region: us-east-1
--- a/org-formation/650-identity-providers/_tasks.yaml
+++ b/org-formation/650-identity-providers/_tasks.yaml
@@ -470,6 +470,38 @@ GithubOidcSageBionetworksSynapseR:
     Repositories:
       - name: "synapser"
         branches: ["*"]
+      - name: "synapserutils"
+        branches: ["*"]
+  DefaultOrganizationBinding:
+    Account: !Ref SynapseProdAccount
+    Region: us-east-1
+
+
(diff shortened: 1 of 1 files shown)


org-formation/650-identity-providers/_tasks.yaml
```diff
@@ -470,6 +470,38 @@ GithubOidcSageBionetworksSynapseR:
     Repositories:
       - name: "synapser"
         branches: ["*"]
+      - name: "synapserutils"
```
thomasyu888, line 473: I wasn't sure if this would be better...

brucehoff (approved):

org-formation/650-identity-providers/_tasks.yaml
thomasyu888: @brucehoff just tagging you here to make sure this isn't missed. There's two ways to do this, just wanted to check it there was a preferable method.

zaro0508: either option would work but you only need one of them.  since it's the same policy going with this one is the better approach.  BTW, to be more secure you can also specify specific branches, for example:  `branches: ["master","dev", "prod", "etc"]`

zaro0508 (changes requested):

org-formation/650-identity-providers/_tasks.yaml
thomasyu888: We are using tags to release, so let's keep as this for now.

zaro0508: By default we allow  all tags to access accounts @thomasyu888.  Think of it as `tags: ["*"]`  always.

thomasyu888: Thanks @zaro0508 - i did not know that!  can you do a re-review?  I made modifications!

org-formation/650-identity-providers/_tasks.yaml
```diff
@@ -469,7 +469,9 @@ GithubOidcSageBionetworksSynapseR:
     GitHubOrg: "Sage-Bionetworks"
     Repositories:
       - name: "synapser"
-        branches: ["*"]
+        branches: ["master", "main", "develop"]
```
zaro0508, line 472: that's weird mast and main? whatever floats your boat i guess

zaro0508 (approved):

brucehoff (approved):

PR #900, opened 2023-06-19
IT-2693: Remove sso access to personal AWS account

xschildw:
This PR removes sso admin access to a personal AWS account that we are deleting.
See Account Teardown at https://sagebionetworks.jira.com/wiki/spaces/IT/pages/745373697/Setup+AWS+Accounts .

### DIFF
--- a/org-formation/700-aws-sso/_tasks.yaml
+++ b/org-formation/700-aws-sso/_tasks.yaml
@@ -302,10 +302,6 @@ Parameters:
     Default: '6498b478-20e1-7031-9648-00c20c410359'
 
   #------------- personal AWS accounts ------------------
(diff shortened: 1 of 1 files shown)


org-formation/700-aws-sso/_tasks.yaml
```diff
@@ -302,10 +302,6 @@ Parameters:
     Default: '6498b478-20e1-7031-9648-00c20c410359'
 
   #------------- personal AWS accounts ------------------
-  8dxfh53AdminGroup:             #JC aws-8dxfh53-admins
```
zaro0508, line 305: don't forget to remove this JC group.

zaro0508 (approved):

PR #909, opened 2023-06-24
temporarily remove SchematicDevAppDnsForward

zaro0508:
The schematica-infra deployment failure is causing the reference in this repo to break which is causing the CI for this repo to fail.  Remove the route53 record set until the schematica-infra deployment is fixed.


zaro0508:
please take a look @linglp

PR #916, opened 2023-06-29
[IT-2736] Use OIDC templates from aws-infra

zaro0508:
With PR https://github.com/Sage-Bionetworks/aws-infra/pull/394 we can now use the OIDC templates from our cloudformation templates repo.  The template in aws-infra works with both org-formation and sceptre.  We remove local templates to reduce duplication.


zaro0508:
I have tested this by using org-formation to generate the cloudformation templates using the njk file and  j2 file then I compared the generated cloudformation templates and both were identical.

xschildw (approved):

PR #921, opened 2023-07-12
[IT-2984] Add rules for inactive codes

jesusaurus:
Add cost category rules for inactive program codes so that we can track ongoing costs related inactive programs.

Depends on Sage-Bionetworks-IT/lambda-mips-api#19


zaro0508 (commented):
The two endpoints seem to return the exact same result?

jesusaurus:
@zaro0508 apparently cloudfront had cached the old value. they're returning different results now

zaro0508 (approved):

PR #922, opened 2023-07-12
PLFM-7886: add bucket permissions required by AWS Quicksight

brucehoff:
In PLFM-7886 we are evaluation AWS Quicksight as a data visualization tool for Athena tables, in the Synapse Dev account.  Quicksight uses a particular bucket naming pattern for the queries it saves. This PR adds the requested bucket names to the AWS policy used with Quicksight.

### DIFF
--- a/org-formation/600-access/_tasks.yaml
+++ b/org-formation/600-access/_tasks.yaml
@@ -304,6 +304,8 @@ SynapseAthenaUserAccessPolicy:
                 "Resource": [
                     "arn:aws:s3:::aws-athena-query-results-${CurrentAccount.AccountId}-us-east-1",
                     "arn:aws:s3:::aws-athena-query-results-${CurrentAccount.AccountId}-us-east-1/*",
+                    "arn:aws:s3:::aws-athena-query-results-us-east-1-${CurrentAccount.AccountId}",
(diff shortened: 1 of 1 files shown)


org-formation/600-access/_tasks.yaml
```diff
@@ -304,6 +304,8 @@ SynapseAthenaUserAccessPolicy:
                 "Resource": [
                     "arn:aws:s3:::aws-athena-query-results-${CurrentAccount.AccountId}-us-east-1",
                     "arn:aws:s3:::aws-athena-query-results-${CurrentAccount.AccountId}-us-east-1/*",
+                    "arn:aws:s3:::aws-athena-query-results-us-east-1-${CurrentAccount.AccountId}",
```
zaro0508, line 307: the two ARNs just above are also athena-query-results buckets.  Do these new ARNs replace the existing ARNs?

zaro0508 (changes requested):

org-formation/600-access/_tasks.yaml
brucehoff: No, the new ARN doesn't replace the current ARN but rather complements it.  Different tools (Athena and Quicksight) seem to require different bucket names.

zaro0508 (approved):

PR #926, opened 2023-07-17
[IT-3012] Tag platform accounts with an owner

jesusaurus:
Add an AccountOwner tag to the platform accounts.

### DIFF
--- a/org-formation/_platform_org_tags.yaml
+++ b/org-formation/_platform_org_tags.yaml
@@ -1,2 +1,3 @@
 CloudwatchCloudTrailLogRetentionPeriod: 90
 CostCenter: 'Platform Infrastructure / 990300'
+AccountOwner: 'it@sagebase.org'


zaro0508 (commented):
this PR seems fine although i'm wondering why there's a `_platform_org_tags.yaml` and a `_default_org_tags.yaml` containing the same tags?  To reduce duplication do you think it would make sense to drop `_platform_org_tags.yaml` in favor of this configuration?:

```
      Tags:
        AccountOwner: it@sagebase.org
```

jesusaurus:
The values being set in `_default_org_tags.yaml` and `_platform_org_tags.yaml` are slightly different, the default CostCenter is `000000`, but the platform CostCenter is `990300`. And the main reason I created `_platform_org_tags.yaml` is to keep the configuration DRY so that if we ever wanted to change the platform tag value we could do it with a one-line change instead of an identical line for each platform account.

zaro0508 (approved):
ooh, sorry i thought the contents were the same.

PR #938, opened 2023-08-01
[IT-3021] Prioritize the RECOVER code

jesusaurus:
Prioritize the RECOVER code so that it is listed near the top of the lambda output, ensuring that it is available in the Service Catalog drop-down menu.

Also update mips-api parameters to match name changes happening in the dependent PR.

Depends on Sage-Bionetworks-IT/lambda-mips-api#21

### DIFF
--- a/org-formation/050-costs/_tasks.yaml
+++ b/org-formation/050-costs/_tasks.yaml
@@ -42,7 +42,7 @@ CostCategoryRulesMicroservice:
   Parameters:
     AcmCertificateArn: !CopyValue [!Sub '${primaryRegion}-${resourcePrefix}-sageit-finops-cert-CertificateArn']
     DnsName: "cost-rules.finops.sageit.org"
-    ChartOfAccountsURL: "https://mips-api.finops.sageit.org/accounts?enable_code_filter&disable_inactive_filter"
+    ChartOfAccountsURL: "https://mips-api.finops.sageit.org/accounts?show_inactive_codes=on"
     ProgramCodeTagList: "CostCenterOther,CostCenter"
 
 # The underlying template uses AWS::Include, but CloudFormation will not detect changes in the included
--- a/sceptre/scipool/config/develop/sc-tag-options.yaml
+++ b/sceptre/scipool/config/develop/sc-tag-options.yaml
@@ -11,7 +11,7 @@ dependencies:
   - "develop/sc-portfolio-scheduled-jobs.yaml"
 sceptre_user_data:
   CostCenters: !from_json
-    - !request 'https://mips-api.finops.sageit.org/tags?only_codes=101300,101400,101600,112501,119400,121700,120100,122000,122100,122300,122500,312000,314900,401900,506800,613500,990100,990300,990400&enable_other_code=true'
+    - !request 'https://mips-api.finops.sageit.org/tags?limit=25&enable_other_code=true&priority_codes=122500'
   ProductIDs:
     - !stack_output_external sc-portfolio-ec2::SCEC2portfolioId
     - !stack_output_external sc-portfolio-s3-basic::SCS3portfolioId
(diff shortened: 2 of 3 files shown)


sceptre/scipool/config/develop/sc-tag-options.yaml
```diff
@@ -11,7 +11,7 @@ dependencies:
   - "develop/sc-portfolio-scheduled-jobs.yaml"
 sceptre_user_data:
   CostCenters: !from_json
-    - !request 'https://mips-api.finops.sageit.org/tags?limit=25&enable_other_code=true'
+    - !request 'https://mips-api.finops.sageit.org/tags?only_codes=101300,101400,101600,112501,119400,121700,120100,122000,122100,122300,122500,312000,314900,401900,506800,613500,990100,990300,990400&enable_other_code=true'
```
zaro0508, line 14: i'm not sure i like this idea because i think enumerating every code we want on the list is difficult to maintain which is really no better than a static list.  what if one of these codes becomes inactive?

zaro0508 (changes requested):

sceptre/scipool/config/develop/sc-tag-options.yaml
brucehoff: @zaro0508 Can you answer @jesusaurus 's question?  I think you are saying that you'd prefer the list to be maintained somewhere else.  Can you suggest an alternative?

jesusaurus: I've re-worked the underlying PR to be a priority list

sceptre/scipool/config/develop/sc-tag-options.yaml
```diff
@@ -11,7 +11,7 @@ dependencies:
   - "develop/sc-portfolio-scheduled-jobs.yaml"
 sceptre_user_data:
   CostCenters: !from_json
-    - !request 'https://mips-api.finops.sageit.org/tags?limit=25&enable_other_code=true'
+    - !request 'https://mips-api.finops.sageit.org/tags?limit=25&enable_other_code=true&priority_codes=122500'
```
zaro0508, line 14: could you please add a comment explaining why we are doing this?  maybe just a link to the issue might be good enough

zaro0508 (changes requested):

org-formation/050-costs/_tasks.yaml
```diff
@@ -42,7 +42,7 @@ CostCategoryRulesMicroservice:
   Parameters:
     AcmCertificateArn: !CopyValue [!Sub '${primaryRegion}-${resourcePrefix}-sageit-finops-cert-CertificateArn']
     DnsName: "cost-rules.finops.sageit.org"
-    ChartOfAccountsURL: "https://mips-api.finops.sageit.org/accounts?enable_code_filter&disable_inactive_filter"
+    ChartOfAccountsURL: "https://mips-api.finops.sageit.org/accounts?show_inactive_codes=on"
```
zaro0508, line 45: what is the purpose of this change?  also i don't see any documentation for `show_inactive_codes` option in the lambda-finops-cost-rules lambda docs.

zaro0508 (changes requested):

org-formation/050-costs/_tasks.yaml
jesusaurus: We need this to match the parameter name changes happening in https://github.com/Sage-Bionetworks-IT/lambda-mips-api/pull/21

sceptre/scipool/config/develop/sc-tag-options.yaml
jesusaurus: done

org-formation/050-costs/_tasks.yaml
zaro0508: ahh ok.

zaro0508 (approved):

PR #943, opened 2023-08-04
Add AccountOwner tag to scipool-dev account

jesusaurus:
Inherit platform tags on the scipool-dev account so that Sage IT is tagged as the account owner.


brucehoff (changes requested):
The update should be made to ScipoolDevAccount, not ScicompAccount.

brucehoff (approved):

PR #946, opened 2023-08-17
SC-383: Allow OIDC from branch v1.1.45

xschildw:
This PR is similar to #945, it fixes the failure to deploy in https://github.com/Sage-Bionetworks/service-catalog-library/pull/305 (job cannot assume role).

### DIFF
--- a/org-formation/650-identity-providers/_tasks.yaml
+++ b/org-formation/650-identity-providers/_tasks.yaml
@@ -259,7 +259,7 @@ GithubOidcCfnTemplateDeploy:
       - name: "aws-infra"
         branches: ["master"]
       - name: "service-catalog-library"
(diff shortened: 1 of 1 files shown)


org-formation/650-identity-providers/_tasks.yaml
```diff
@@ -259,7 +259,7 @@ GithubOidcCfnTemplateDeploy:
       - name: "aws-infra"
         branches: ["master"]
       - name: "service-catalog-library"
-        branches: ["master"]
+        branches: [ "master", "v1.1.45" ]
```
brucehoff, line 262: Do you really want to give the specific branch ( "v1.1.45")? What happens when you go to the next version, e.g.,  "v1.1.46"?

PR #950, opened 2023-08-22
Add Org-Formation Overview

jesusaurus:
Add a table of contents describing the grouping of stacks in the subdirectories under `org-formation/`.

This is intended to make the stacks easier to navigate and to simplify the decision of where to add new stacks.


brucehoff (commented):
I'll defer to @zaro0508 to approve, since he is the expert, but this looks _great_ to me.  Thanks for doing this!

zaro0508 (approved):

PR #952, opened 2023-08-24
[SC-383] Update service catalog notebook product

zaro0508:
update notebook product version and fix description in previous versions.

depends on https://github.com/Sage-Bionetworks/service-catalog-library/pull/308


xschildw (approved):
OK after new tag in place.

PR #955, opened 2023-08-28
IT-2997: Add rules in all accounts for CIS 4.1/4.2

xschildw:
This PR adds exceptions for CIS 4.1/4.2 (access to 22 and 3389 from 0.0.0.0/0). Also adds NOTIFIED to workflow status to catch findings that were manually changed.

### DIFF
--- a/org-formation/075-security-hub/security-hub-suppress-infra.yaml
+++ b/org-formation/075-security-hub/security-hub-suppress-infra.yaml
@@ -369,12 +369,15 @@ Resources:
               - 'aws-foundational-security-best-practices/v/1.0.0/IAM.6'
               - 'arn:aws:securityhub:::ruleset/cis-aws-foundations-benchmark/v/1.2.0/rule/2.7'
               - 'cis-aws-foundations-benchmark/v/1.4.0/3.7'
-              # SecHub Recommendation: no security group allows unrestricted ingress access to port 22.
+              # SecHub Recommendation: no security group allows unrestricted ingress access to port 22/3389.
               # Supression Reason: Allow for bastian hosts and testing
               - 'cis-aws-foundations-benchmark/v/1.2.0/rule/4.1'
+              - 'arn:aws:securityhub:::ruleset/cis-aws-foundations-benchmark/v/1.2.0/rule/4.1'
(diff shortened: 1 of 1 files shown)


org-formation/075-security-hub/security-hub-suppress-infra.yaml
```diff
@@ -369,12 +369,15 @@ Resources:
               - 'aws-foundational-security-best-practices/v/1.0.0/IAM.6'
               - 'arn:aws:securityhub:::ruleset/cis-aws-foundations-benchmark/v/1.2.0/rule/2.7'
               - 'cis-aws-foundations-benchmark/v/1.4.0/3.7'
-              # SecHub Recommendation: no security group allows unrestricted ingress access to port 22.
+              # SecHub Recommendation: no security group allows unrestricted ingress access to port 22/3389.
               # Supression Reason: Allow for bastian hosts and testing
               - 'cis-aws-foundations-benchmark/v/1.2.0/rule/4.1'
```
xschildw, line 374: Not sure about this one, I don't see it in findings when I search but there are still others with the same format so leaving for now.

org-formation/075-security-hub/security-hub-suppress-infra.yaml
```diff
@@ -369,12 +369,15 @@ Resources:
...
+              - 'arn:aws:securityhub:::ruleset/cis-aws-foundations-benchmark/v/1.2.0/rule/4.1'
```
zaro0508, line 375: if these are ARNs then shouldn't `cis-aws-foundations-benchmark/v/1.4.0/3.7` be an ARN as well? or should we just remove that entry?

xschildw: That's what I would have thought, but there are some findings that have generatorID like 'cis-aws-foundations-benchmark/v/1.4.0/2.1.5.2' (i.e. not ARNs) so I'm not sure we can remove them (or rename them as ARNs). I also had questions about the version of CIS... The generatorIDs above refer to CIS v1.2.0 but now're we're following CIS v1.4.0... I looked for v1.4.0 versions of the generators but did not see any in the findings so I'm just leaving everything as-is (if something needs to change, we'll see new findings pop up.).

zaro0508 (approved):

PR #961, opened 2023-09-06
[IT-3000] Supress public bucket rule

zaro0508:
Buckets used for cost reporting can be public so we are suppressing the security hub notifications for those buckets.

### DIFF
--- a/org-formation/075-security-hub/security-hub-suppress-infra.yaml
+++ b/org-formation/075-security-hub/security-hub-suppress-infra.yaml
@@ -404,7 +404,7 @@ Resources:
                 - prefix: 'arn:aws:s3:::essentials-awss3lambdaartifactsbucket-'
                 - prefix: 'arn:aws:s3:::cf-templates-'
                 - prefix: 'arn:aws:s3:::cloudformationmanageduploadinfrast-'
-                - prefix: 'arn:aws:s3:::cost-and-usage-reports.sagebase.orggi'
+                - prefix: 'arn:aws:s3:::cost-and-usage-reports.sagebase.org'
                 # This should only match with 2.1.5.1
                 - 'AWS::::Account:140124849929'
             GeneratorId:
--- a/org-formation/075-security-hub/security-hub-suppress-infra.yaml
+++ b/org-formation/075-security-hub/security-hub-suppress-infra.yaml
@@ -403,7 +403,7 @@ Resources:
                 - prefix: 'arn:aws:s3:::bootstrap-awss3cloudformationbucket-'
                 - prefix: 'arn:aws:s3:::essentials-awss3lambdaartifactsbucket-'
                 - prefix: 'arn:aws:s3:::cf-templates-'
-                - prefix: 'arn:aws:s3:::cloudformationmanageduploadinfrast-'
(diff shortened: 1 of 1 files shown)


org-formation/075-security-hub/security-hub-suppress-infra.yaml
```diff
@@ -403,6 +403,8 @@ Resources:
                 - prefix: 'arn:aws:s3:::bootstrap-awss3cloudformationbucket-'
                 - prefix: 'arn:aws:s3:::essentials-awss3lambdaartifactsbucket-'
                 - prefix: 'arn:aws:s3:::cf-templates-'
+                - prefix: 'arn:aws:s3:::cloudformationmanageduploadinfrast-'
```
jesusaurus, line 406: I'm not sure what this bucket is for, but I don't think it's used by cost reporting. It seems to be tagged with the stack arn:aws:cloudformation:us-east-1:531805629419:stack/CloudFormationManagedUploadInfrastructure/87ed7c10-866c-11eb-b672-12ff31b78b4f

org-formation/075-security-hub/security-hub-suppress-infra.yaml
```diff
@@ -403,6 +403,8 @@ Resources:
...
+                - prefix: 'arn:aws:s3:::cost-and-usage-reports.sagebase.orggi'
```
jesusaurus, line 407: there's a small typo in this name: `.orggi` should be `.org`

jesusaurus (changes requested):

org-formation/075-security-hub/security-hub-suppress-infra.yaml
zaro0508: fixed

org-formation/075-security-hub/security-hub-suppress-infra.yaml
zaro0508: i believe it's created by the [AWS CLI to store AWS custom resource types](https://awstip.com/using-the-aws-cloudformation-registry-to-manage-your-own-custom-cloudformation-resource-types-part-aa1b298ed22e)..

```
Submitting the Custom Resource Type
You can use the cfn submit command for this. The command will package your code and then create a [managed upload infrastructure stack](https://github.com/aws-cloudformation/cloudformation-cli/blob/master/src/rpdk/core/data/managed-upload-infrastructure.yaml). The stack creates an Amazon S3 bucket encrypted using an AWS KMS key, and the packaged code is uploaded to this bucket. You will be charged for the AWS KMS key, Amazon S3 storage charges, and other charges associated with both of these services. Currently, this is a minimum of $1.00/month because of the KMS key created, and the rest of the charges should be minimal in most scenarios. After uploading the packaged code to S3, the command registers the Resource Type with the CloudFormation registry.
```

The [template to create this bucket is in the AWS CLI repo](https://github.com/aws-cloudformation/cloudformation-cli/blob/master/src/rpdk/core/data/managed-upload-infrastructure.yaml)

jesusaurus: Ohhh, that's where that comes from, thanks

jesusaurus (approved):

PR #965, opened 2023-09-16
[IT-3090] Use AWS config to enable VPC flow logs

zaro0508:
Security hub says we should enable VPC flow logs.  The solution we are going to use to do that is outlined in the AWS blog post[1] on how to enable vpc flow logs automatically using aws config rules. Their implementation uses control tower, ours will use org-formation.

[1] https://aws.amazon.com/blogs/mt/how-to-enable-vpc-flow-logs-automatically-using-aws-config-rules/

### DIFF
--- a/org-formation/725-vpc-flow-logs/README.md
+++ b/org-formation/725-vpc-flow-logs/README.md
@@ -1,0 +1,9 @@
+### Purpose of these templates
+
+The templates in this folder is used to implement the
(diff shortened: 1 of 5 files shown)


brucehoff:
The obvious question is: Why put Config in the loop?  Why wait for Config to flag a VPC as lacking flow logging rather than just updating any VPC creation scripts we have to enable flow logging so that they are never flagged by Config in the first place?  I am guessing that the answer is that there are a variety of processes that create VPCs, including ones that are outside of the control of "IT", and the approach used in this PR ensures flow logging is enabled regardless of how the VPC was created.

brucehoff (approved):

PR #966, opened 2023-09-19
IT-3162: Add hosted zone in sageit

xschildw:
This PR adds a hosted zone for 'bridgedigital.health' in the sageit account.
Once in, we can manually change the name servers at the registrar to use the zone.


zaro0508 (changes requested):
would it be possible to do this in `org-formation/100-shared-dns` instead?

zaro0508 (approved):

PR #969, opened 2023-09-19
[IT-3109] remove service user

zaro0508:
We have switch all repos that use this service user from travis to github action (OIDC) therefore we should remove the service user. The only difference in the updated template is the removal of the service user.


zaro0508:
> Did you mean to include this `khai` directory?

fixed.  thanks.

jesusaurus (approved):

PR #975, opened 2023-09-21
Setup OIDC access to github for SC login app

zaro0508:
In anticipation of moving the service catalog login app to github actions we are setting up GH OIDC access for the repos.

The idea is to be able to allow the login app repos access to the AWS accounts it needs to deploy to (bmgfki, scipooldev, scipoolprod, and strides)

### DIFF
--- a/org-formation/650-identity-providers/_tasks.yaml
+++ b/org-formation/650-identity-providers/_tasks.yaml
@@ -697,3 +697,51 @@ GithubOidcSageBionetworksMTBSageIT:
   DefaultOrganizationBinding:
     Account: !Ref SageITAccount
     Region: us-east-1
+
+# for https://github.com/Sage-Bionetworks/synapse-login-aws-infra
+GithubOidcSageBioNetworksScSynapseLoginInfra:
+  Type: update-stacks
+  DependsOn: GithubOidcSageBionetworks
+  Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.7.6/templates/IAM/github-oidc-provider.j2
+  StackName: !Sub ${resourcePrefix}-${appName}-sage-bionetworks-sc-synapse-login-infra
+  Parameters:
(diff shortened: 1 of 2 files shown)


sceptre/strides/config/prod/github-oidc-sage-bionetworks.yaml
```diff
@@ -14,3 +14,7 @@ sceptre_user_data:
   Repositories:
     - name: "synapse-login-aws-infra"
       branches: ["dev", "prod"]
+    - name: "synapse-login-aws-infra"
```
jesusaurus, line 17: there's already an entry for `synapse-login-aws-infra` so I think this just needs to update the branches list. other than that, this change looks good to me.

jesusaurus (changes requested):

org-formation/650-identity-providers/_tasks.yaml
```diff
@@ -697,3 +697,51 @@ GithubOidcSageBionetworksMTBSageIT:
   DefaultOrganizationBinding:
     Account: !Ref SageITAccount
     Region: us-east-1
+
+# for https://github.com/Sage-Bionetworks/synapse-login-aws-infra
+GithubOidcSageBioNetworksScSynapseLoginInfra:
```
zaro0508, line 702: this setup already exists at https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/org-formation/650-identity-providers/_tasks.yaml#L443-L501

just need to add appropriate accounts

sceptre/strides/config/prod/github-oidc-sage-bionetworks.yaml
zaro0508: fixed, thanks

org-formation/650-identity-providers/_tasks.yaml
zaro0508: fixed

jesusaurus (approved):

PR #990, opened 2023-10-11
PLFM-8054: Temporary access for Snowflake to datawarehouse S3 bucket

xschildw:
This PR sets up a role with trust relationship for Snowflake to access the datawarehouse S3 bucket.

Needs SSM param in place before merging, then needs to be merge again with actual values per process described at https://docs.snowflake.com/en/user-guide/data-load-s3-config-storage-integration (verified with @thomasyu888 by running template manually in dev account).


sceptre/synapseprod/config/prod/snowflake-access.yaml
```diff
@@ -0,0 +1,6 @@
+template:
+  path: snowflake-access.yaml
+stack_name: snowflake-accesss
+parameters:
+  SnowflakeAccountArn: !ssm /infra/SnowflakeAccountArn
```
thomasyu888, line 5: I'm not as knowledgable about this, but where are these values pulled from?

sceptre/synapseprod/templates/snowflake-access.yaml
```diff
@@ -0,0 +1,61 @@
+Description: Setup cross account IAM access.  Give user in another AWS account access to resources.
```
zaro0508, line 1: is this setup from snowflake docs?  if so it might be good to add a comment in this file referencing the doc.

zaro0508 (changes requested):

sceptre/synapseprod/config/prod/snowflake-access.yaml
xschildw: Agree about account, will actually make things easier as we first have to setup the role with fake trust relationship values, then update with actual values.

sceptre/synapseprod/templates/snowflake-access.yaml
xschildw: Done.

sceptre/synapseprod/templates/snowflake-access.yaml
```diff
@@ -0,0 +1,61 @@
...
+        {
+          "Version": "2012-10-17",
+          "Statement": [
+              {
+                  "Effect": "Allow",
+                  "Action": [
+                      "s3:GetObject",
+                      "s3:GetObjectVersion"
+                  ],
+                  "Resource": "arn:aws:s3:::prod.datawarehouse.sagebase.org/warehouse/*"
```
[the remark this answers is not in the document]
xschildw: Correct, we did a POC in the dev account. There's only a prod datawarehouse (in the prod account) so I think this is fine. When/if we move forward to actual production, the plan is to update the cross-account-access.yaml template to support optional externalId parameter.

sceptre/synapseprod/templates/snowflake-access.yaml
```diff
@@ -0,0 +1,61 @@
...
+              - sts:AssumeRole
+              - sts:TagSession
+            Condition:
+              StringEquals:
+                sts:ExternalId: !Sub ${SnowflakeAccountExternalId}
+Outputs:
+  SnowflakeServiceRoleArn:
+    Value: !GetAtt SnowflakeServiceRole.Arn
+    Export:
+      Name: !Sub '${AWS::Region}-${AWS::StackName}-SnowflakeServiceRoleArn'
```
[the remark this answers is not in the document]
xschildw: Done

thomasyu888 (approved):

PR #1001, opened 2023-10-17
Deploy one merge at a time

zaro0508:
We want the CI to only deploy once per merge otherwise the deployments might get into a bad state.

From GH action docs, this change should do the following..

  "When a concurrent job or workflow is queued, if another job or
  workflow using the same concurrency group in the repository is
  in progress, the queued job or workflow will be pending."

[1] https://docs.github.com/en/actions/using-jobs/using-concurrency

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -6,6 +6,9 @@ on:
   push:
     branches: ['*']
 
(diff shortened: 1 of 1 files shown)


xschildw (approved):
Lol was looking at same thing...

jesusaurus (approved):

PR #1005, opened 2023-10-23
IT-2396: Exempt redash svc user

xschildw:
This PR adds the IAM user for the redash service user in the synapse prod account to the exceptions.


org-formation/075-security-hub/security-hub-suppress-infra.yaml
```diff
@@ -444,6 +444,8 @@ Resources:
                 - 'arn:aws:iam::325565585839:user/ran-uploader'
                 # scipool IT-2315
                 - prefix: 'arn:aws:iam::237179673806:user/synapse-login-scipoolprod-ServiceUser-'
+                # synapseprod IT-2396, service account for redash
```
zaro0508, line 447: this is the same situation as https://sagebionetworks.jira.com/browse/IT-3123  which, i think, @brucehoff prefers to rotate instead of suppress.  You might want to check with him first because i'm not sure if he envisions his preference to be the general case.

xschildw: Good point, rotated the key.

PR #1007, opened 2023-10-25
[IT-3224] Split finance and security auditor groups

jesusaurus:
Create separate groups for finance auditors and security auditors. Once these new groups are in place and populated with users, the old auditor group can be deprecated and users removed.

### DIFF
--- a/org-formation/700-aws-sso/_tasks.yaml
+++ b/org-formation/700-aws-sso/_tasks.yaml
@@ -301,6 +301,14 @@ Parameters:
     Type: String
     Default: '6498b478-20e1-7031-9648-00c20c410359'
 
+  financeAuditorGroup:  # JC aws-finance-auditors
+    Type: String
+    Default: 'd438e4d8-90e1-7030-6998-cad12c0d3296'
+
+  securityAuditorGroup:     # JC aws-security-auditors
+    Type: String
+    Default: '2448e4e8-50b1-70e5-def0-07e0f4fcd60e'
+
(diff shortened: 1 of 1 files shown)


org-formation/700-aws-sso/_tasks.yaml
```diff
@@ -418,6 +426,58 @@ SsoAuditor:
...
+    TargetBinding:
+      Account: '*'
+  Parameters:
+    instanceArn: !Ref instanceArn
+    principalId: !Ref financeAuditorGroup
+    permissionSetName: 'FinanceAuditor'
+    managedPolicies:
+      - 'arn:aws:iam::aws:policy/job-function/SupportUser'
+      - 'arn:aws:iam::aws:policy/AWSBillingReadOnlyAccess'
+    sessionDuration: 'PT1H'
```
zaro0508, line 451: 1hr is pretty short, wondering if we should just bump this up slightly?  maybe 2-3 hrs?

zaro0508 (approved):
would be nice to also see a PR for the removal of the general `auditorGroup`.  it could just be dependent PR of this one.

org-formation/700-aws-sso/_tasks.yaml
jesusaurus: good point, I'm not sure why the original auditor role is so short, but it looks like most of the other roles are 8 hours, so I'll use that for the finance role (but keep the security role at 1 hour).

PR #1028, opened 2023-11-16
Create role for snoflake access to synapsedev data with fake extID to get roleArn

xschildw:
This PR will create a new role for the snowflake access to Synapse dw bucket.
First called with fake externalID to get the role ARN.

Depends on https://github.com/Sage-Bionetworks/aws-infra/pull/398

### DIFF
--- a/sceptre/synapsedev/config/prod/snowflake-access.yaml
+++ b/sceptre/synapsedev/config/prod/snowflake-access.yaml
@@ -1,0 +1,4 @@
+template:
+  type: http
+  url: https://{{stack_group_config.admincentral_cf_bucket}}.s3.amazonaws.com/aws-infra/v0.7.8/IAM/snowflake-synapse-access.yaml
+stack_name: snowflake-access.yaml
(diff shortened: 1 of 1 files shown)


zaro0508 (approved):

sceptre/synapsedev/config/prod/snowflake-access.yaml
```diff
@@ -0,0 +1,9 @@
+template:
+  type: http
+  url: https://{{stack_group_config.admincentral_cf_bucket}}.s3.amazonaws.com/aws-infra/v0.7.8/IAM/snowflake-bucket-access.yaml
+stack_name: snowflake-access.yaml
+parameters:
+  BucketArn: arn:aws:s3:::dev.datawarehouse.sagebase.org
```
zaro0508, line 6: this makes me wonder why the role for this bucket isn't created in the same template that created the bucket?

xschildw: Interesting thought, the goal was for org-infra to adopt resources already deployed into the with a shared template (with the benefit that the template can be used for to give Snowflake access to any bucket/prefix in the org)

PR #1029, opened 2023-11-16
Snoflake access: Switch to shared script

xschildw:
This PR replaces the existing script by the shared script in aws-infra to create the role for snowflake to access Synapse dw data.

Check that everything is OK with #1028 before merging.


sceptre/synapseprod/config/prod/snowflake-access.yaml
```diff
@@ -1,6 +1,8 @@
 template:
-  path: snowflake-access.yaml
+  type: http
+  url: https://{{stack_group_config.admincentral_cf_bucket}}.s3.amazonaws.com/aws-infra/v0.7.8/IAM/snowflake-synapse-access.yaml
```
zaro0508, line 3: where is `snowflake-synapse-access.yaml`? i only see `templates/IAM/snowflake-bucket-access.yaml` in PR https://github.com/Sage-Bionetworks/aws-infra/pull/398

zaro0508 (changes requested):

PR #1034, opened 2023-11-22
IT-3318: Update version

xschildw:
This PR adds version 1.2.13 to test if *.6a instance types work.


xschildw:
It looks like the problem is on the strides account. We should push to that only.

PR #1038, opened 2023-11-29
IT-3322 undo breaking change

brucehoff:
I had changed `IncludeMasterAccount` from `true` to `false` to stop account specific roles from showing up under the Organizations account, but the change also removed the roles from the intended accounts.  I am unclear on how `IncludeMasterAccount` is supposed to work.  For now I am submitting this PR to undo the breaking change.


jesusaurus (approved):
I also find it confusing why this is needed, but I expect it to resolve the issue

PR #1045, opened 2023-11-30
IT-3335: Add Agora repo to OIDC

xschildw:
This PR enables AWS access for GH workflows from the Sage-Bionetworks/Agora repo.

### DIFF
--- a/org-formation/650-identity-providers/_tasks.yaml
+++ b/org-formation/650-identity-providers/_tasks.yaml
@@ -735,6 +735,8 @@ GithubOidcAgoraInfraDeploy:
     Repositories:
       - name: "agora2-infra"
         branches: ["main"]
+      - name: "Agora"
+        branches: ["develop", "prod", "staging"]
   DefaultOrganizationBinding:
     Account:
       - !Ref AgoraDevAccount


org-formation/650-identity-providers/_tasks.yaml
```diff
@@ -735,6 +735,8 @@ GithubOidcAgoraInfraDeploy:
     Repositories:
       - name: "agora2-infra"
         branches: ["main"]
+      - name: "Agora"
```
zaro0508, line 738: `Sage-Bionetworks/Agora` repo is currently using this role for deployment of the agora app: https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/org-formation/600-access/_tasks.yaml#L5-L18

We want to replace that role with a more scoped policy. The Agora repo only needs to deploy to beanstalk so i would create a separate resource here with a more scoped (non-admin) policy then delete the one in `600-access folder/_tasks.yaml` folder.

xschildw: OK

zaro0508:
@xschildw could you also prepare a PR for the removal `AgoraCIServiceAccount` and make it a dependency of this PR?

zaro0508 (approved):

jesusaurus (approved):

PR #1068, opened 2024-01-17
[IT-3403] new s3 viewer group

zaro0508:
Create a new S3 viewer group to allow the digital analytics team access to download objects from S3 buckets in the AWS recover accounts.


thomasyu888 (approved):
LGTM!  Thanks for tackling this so quickly!

PR #1069, opened 2024-01-18
IT-3317:  Add multi-notebook product to prod, stride, bmgfki

brucehoff:
This PR adds the multi-notebook product to prod, stride, and bmgfki service catalog instances.
It also removes the 'old' notebook product from the dev' SC instance.

Depends on https://github.com/Sage-Bionetworks-IT/organizations-infra/pull/1071

Depends on https://github.com/Sage-Bionetworks/service-catalog-library/pull/322

Note: After https://github.com/Sage-Bionetworks/service-catalog-library/pull/322 is merged we must tag that repo' with `v1.3.5`.

Depends on https://github.com/Sage-Bionetworks-IT/cfn-cr-alb-rule/pull/16

Note: After https://github.com/Sage-Bionetworks-IT/cfn-cr-alb-rule/pull/16 is merged with must tag that repo' with `0.1.5`.


xschildw (approved):

sceptre/scipool/config/develop/sc-product-ec2-linux-jumpcloud-notebook.yaml
```diff
@@ -1,114 +0,0 @@
-template:
```
zaro0508, line 1: i would recommend keeping this until everything is working.  remove all of these in one PR.

sceptre/scipool/config/prod/sc-product-ec2-linux-docker-notebook.yaml
```diff
@@ -0,0 +1,19 @@
+template:
+  path: "sc-product-ec2-linux-docker-notebook.j2"
+stack_name: "sc-product-ec2-linux-docker-notebook"
```
zaro0508, line 3: you'll need to add workaround here as well, https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/.github/workflows/aws-scipool.yaml#L54-L62

zaro0508 (changes requested):

sceptre/scipool/config/prod/sc-product-ec2-linux-docker-notebook.yaml
brucehoff: @zaro0508 Why does `aws-scipool.yaml` only refer to `prod` and not to `develop`, `strides`, or `bmgfki`?

zaro0508: we forgot to enable it for all prod accounts, the fix is in PR #1071.  it doesn't need to be setup for develop because we want to keep older versions around for debugging in develop account.

.github/workflows/aws-scipool.yaml
```diff
@@ -58,5 +58,6 @@ jobs:
           mkdir -p templates/remote
           sceptre delete --yes prod/sc-product-assoc-ec2-linux-docker.yaml
           sceptre delete --yes prod/sc-product-assoc-ec2-linux-jumpcloud-notebook.yaml
+          sceptre delete --yes prod/sc-product-assoc-ec2-linux-docker-notebook.yaml
```
zaro0508, line 61: you added the product for all production environments (prod, bmgfki, strides) so you will need an entry for each.

Also i would consider only enabling this product on prod for further testing then promote it to  bmgfki and strides so that it gets testing from internal sage folks before making available to more external users.   Just a thought but it's up to you.

zaro0508 (changes requested):

.github/workflows/aws-scipool.yaml
brucehoff: > you added the product for all production environments (prod, bmgfki, strides) so you will need an entry for each.

This is the change you made in #1071 , correct?

> Also i would consider only enabling this product on prod for further testing then promote it to bmgfki and strides so that it gets testing from internal sage folks before making available to more external users. Just a thought but it's up to you.

Once we've tested in 'develop' I'd like to promote to all 3 production environments.

zaro0508: > This is the change you made in #1071 , correct?
no, that PR does not add the `linux-docker-notebook` product.

zaro0508 (changes requested):

.github/workflows/aws-scipool.yaml
zaro0508: > This is the change you made in #1071 , correct?

no, that PR does not add this product (linux-docker-notebook) to the list.

PR #1070, opened 2024-01-18
Disable GH action concurrency

zaro0508:
* We want to only deploy from master branch
* We do not want concurrent cloudformation deployments to AWS accounts.
* We also don't want to interrupt cloudformation deployments otherwise it
  may get into a bad state.  Therefore we are disabling concurrency[1]

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idconcurrency

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -4,10 +4,12 @@ on:
   pull_request:
     branches: ['*']
   push:
-    branches: ['*']
+    branches: ['master']
 
 # Ensures that only one deploy task per branch/environment will run at a time.
-concurrency: ci-${{ github.ref }}
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: false
 
 jobs:
   pre-commit:


.github/workflows/main.yaml
```diff
@@ -4,10 +4,12 @@ on:
   pull_request:
     branches: ['*']
   push:
-    branches: ['*']
+    branches: ['master']
 
 # Ensures that only one deploy task per branch/environment will run at a time.
-concurrency: ci-${{ github.ref }}
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
```
brucehoff, line 11: Wouldn't it be even safer to make this more restrictive, like:
```
  group: ${{ github.workflow }}
```

?
This would keep multiple 'triggers' of this workflow from running concurrently.

brucehoff (changes requested):

.github/workflows/main.yaml
zaro0508: mehh, i've changed the above `on branches` from `*` to `master` therefore we really only have one branch (or environment) that deploys.

brucehoff (approved):

PR #1076, opened 2024-01-25
[IT-3414] update SSO S3 viewer role

zaro0508:
Give S3 viewer role read only access to cloudwatch logs.


thomasyu888 (approved):
Thanks @zaro0508 !

PR #1084, opened 2024-02-01
IT-3394: Remove 'old' RStudio notebook product from develop SC instance

brucehoff:
Removing the product-portfolio association for the "old" RStudio notebook product, develop only.


zaro0508 (changes requested):
Why are you only removing the association file?  don't we want to remove the [deprecated notebook product file](https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/sceptre/scipool/config/develop/sc-product-ec2-linux-jumpcloud-notebook.yaml)?

Please remove the corresponding one in [github actions](https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/.github/workflows/aws-scipool.yaml#L54-L70) as well. 

Please be aware that Sceptre will not delete stacks this way.  If you want to delete a stack you must delete the file and also manually delete it from the AWS cloudfomation console.  An alternative way to delete stacks thru the CI is to use the [Sceptre obsolete](https://docs.sceptre-project.org/latest/docs/stack_config.html?highlight=obsolete#obsolete) parameter.  The CI is already setup to use the [Sceptre prune action](https://docs.sceptre-project.org/latest/docs/cli.html?highlight=prune#sceptre-prune)  so you only need to mark it as obsolete.

xschildw:
Not critical in dev but does removing the product affect ability to work with deployed instances (start/stop?)?

brucehoff:
> Why are you only removing the association file? don't we want to remove the [deprecated notebook product file](https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/sceptre/scipool/config/develop/sc-product-ec2-linux-jumpcloud-notebook.yaml)?

DONE

> 
> Please remove the corresponding one in [github actions](https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/.github/workflows/aws-scipool.yaml#L54-L70) as well.

This PR is to remove from `develop` only. (Once I verify the removal worked correctly in `develop` I will move on to remove the product from the other SC instances.). There is no reference to the develop instance in that workflow.

> 
> Please be aware that Sceptre will not delete stacks this way. If you want to delete a stack you must delete the file and also manually delete it from the AWS cloudfomation console. 

By "stack" are you referring to the stack which defines the Service Catalog product?

@xschildw asks:

> Not critical in dev but does removing the product affect ability to work with deployed instances (start/stop?)?

That's a great question. I have provisioned an RStudio product in `develop` and will verify I can still use it after this PR is merged.  (If not, then we can't apply the same change to the production SCs until people stop using the product.)

zaro0508 (approved):
> By "stack" are you referring to the stack which defines the Service Catalog product?

no, it's the cloudformation stacks (sc-product-ec2-linux-jumpcloud-notebook & sc-product-assoc-ec2-linux-jumpcloud-notebook) that you are removing in this PR.  You can see them in the AWS cloudformation console page.

PR #1090, opened 2024-02-07
Reduce privileges for application managers

zaro0508:
I think the application manager role is a bit over privileged. They should probably not have unilateral access to the secrets manger in the entire AWS account.

### DIFF
--- a/org-formation/700-aws-sso/_tasks.yaml
+++ b/org-formation/700-aws-sso/_tasks.yaml
@@ -574,7 +574,6 @@ SsoApplicationManager:
     managedPolicies:
       - arn:aws:iam::aws:policy/job-function/ViewOnlyAccess
       - arn:aws:iam::aws:policy/CloudWatchReadOnlyAccess
-      - arn:aws:iam::aws:policy/SecretsManagerReadWrite
       - arn:aws:iam::aws:policy/AWSCertificateManagerReadOnly
     inlinePolicy: >-
       {


brucehoff (changes requested):
This change looks like it would remove the ability for application managers to manage the secrets for their applications, doesn't it?

zaro0508:
> This change looks like it would remove the ability for application managers to manage the secrets for their applications, doesn't it?

yes, that is correct @brucehoff.  There are lots of app managers now including people who are not app developers. My concern is that it could be deleted/changed by non developers.  if you feel that this is fine then we can just close this PR.

brucehoff:
> > This change looks like it would remove the ability for application managers to manage the secrets for their applications, doesn't it?
> 
> yes, that is correct @brucehoff. There are lots of app managers now including people who are not app developers. My concern is that it could be deleted/changed by non developers. if you feel that this is fine then we can just close this PR.

No, I do NOT feel it is fine to give people excessive access.  We have to give people the right level of access, meaning that non-developers should not have access to the secret store but neither should we remove access to the secret store from the application mangers who require it.

zaro0508:
> No, I do NOT feel it is fine to give people excessive access. We have to give people the right level of access, meaning that non-developers should not have access to the secret store but neither should we remove access to the secret store from the application mangers who require it.

@brucehoff for further context, here's @milen-sage request https://sagebionetworks.slack.com/archives/CEFQD0KU1/p1706290847735899

zaro0508:
We accomplished least privilege with an alternative approach.  We give [Application Manager role](https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/org-formation/700-aws-sso/_tasks.yaml#L555) to a few developers while we give everyone else access to the [Viewer Plus](https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/org-formation/700-aws-sso/_tasks.yaml#L507) role.

PR #1091, opened 2024-02-12
[IT-3386] Send a copy of cost reports to a new google group.

jesusaurus:
Update the email report lambda to a version that supports sending copies to additional recipients, and add a newly-created google group as an additional recipient.

Depends on Sage-Bionetworks-IT/lambda-finops-email-totals#19

### DIFF
--- a/org-formation/050-costs/_tasks.yaml
+++ b/org-formation/050-costs/_tasks.yaml
@@ -61,13 +61,14 @@ CostCategories:
 # use the payer account so that cost explorer aggregates member account data
 CostNotificationMicroservice:
   Type: update-stacks
-  Template: !Sub 'https://${AdminCentralCfnBucket}.s3.amazonaws.com/lambda-finops-email-totals/1.0.0/lambda-finops-email-totals.yaml'
+  Template: !Sub 'https://${AdminCentralCfnBucket}.s3.amazonaws.com/lambda-finops-email-totals/1.1.0/lambda-finops-email-totals.yaml'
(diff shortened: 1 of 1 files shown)


org-formation/050-costs/_tasks.yaml
```diff
@@ -61,13 +61,14 @@ CostCategories:
...
   Type: update-stacks
-  Template: !Sub 'https://${AdminCentralCfnBucket}.s3.amazonaws.com/lambda-finops-email-totals/1.0.0/lambda-finops-email-totals.yaml'
+  Template: !Sub 'https://${AdminCentralCfnBucket}.s3.amazonaws.com/lambda-finops-email-totals/1.1.0/lambda-finops-email-totals.yaml'
   StackName: !Sub '${resourcePrefix}-cost-notifications'
   DefaultOrganizationBinding:
     Account: !Ref MasterAccount
     Region: !Ref primaryRegion
   Parameters:
     AdminEmail: "it@sagebase.org"
+    CopyRecipients: "cloud-cost-audit@sagebase.org"
```
zaro0508, line 71: i think this google group is misconfigured, you need to set `Who can post` to anyone

zaro0508 (changes requested):

org-formation/050-costs/_tasks.yaml
jesusaurus: good catch, I've updated the setting

zaro0508 (approved):

PR #1116, opened 2024-04-11
[IT-3471] Setup dashboard for iAtlas

zaro0508:
Setup cross account cloudwatch dashboard for IAtlas app.

depends on #1115

### DIFF
--- a/org-formation/740-cloudwatch-dashboard/_tasks.yaml
+++ b/org-formation/740-cloudwatch-dashboard/_tasks.yaml
@@ -60,3 +60,19 @@ CrossAccountDashboards:
       - Name: "agora-prod"
         ElasticBeanstalkEnvName: "agora-prod"
         AccountId: "681175625864"
+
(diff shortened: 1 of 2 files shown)


jesusaurus:
This looks identical to the nextflow dashboards. Is it likely that the two dashboards will diverge once they are actively used, or does it make sense to rename `nextflow-cw-dashboards.njk` to `ecs-cw-dashboards.njk` and reuse it in both places?

zaro0508:
> Is it likely that the two dashboards will diverge once they are actively used, or does it make sense to rename `nextflow-cw-dashboards.njk` to `ecs-cw-dashboards.njk` and reuse it in both places?

I have only added the basic metrics to each project's dashboard.  Each project can have very custom dashboards with an very different configurations and there are many parameters that can be set in the json.  Therefore I think each project will have a custom dashboard and they will most definitely diverge from each other.

jesusaurus (approved):

PR #1122, opened 2024-04-17
[IT-3456] Deploy EC2 terminator lambda to imagecentral

jesusaurus:
Deploy EC2 terminator lambda to imagecentral


zaro0508 (changes requested):
the stuff in `sceptre` folders is around for legacy.  In this case we can probably remove everything in `sceptre/imagecentral` because i don't think it's necessary anymore.

please move the deployment of this lambda to `org-formation/300-account-defaults`

PR #1126, opened 2024-04-18
[IT-3554] Allow SC Enduser roles to tag resources

jesusaurus:
Add a product constraint allowing SC users to update tags on their provisioned products.

### DIFF
--- a/sceptre/scipool/templates/sc-product-ec2-linux-docker-notebook.j2
+++ b/sceptre/scipool/templates/sc-product-ec2-linux-docker-notebook.j2
@@ -89,6 +89,14 @@ Resources:
       ProductId: !Ref 'ScEc2LinuxNotebookProduct'
       RoleArn: !ImportValue
         'Fn::Sub': '${AWS::Region}-sc-ec2vpc-launchrole-LaunchRoleArn'
+  ConstraintTagUpdate:
+    Type: AWS::ServiceCatalog::ResourceUpdateConstraint
+    DependsOn: AssociatesExternalPortfolio
+    Properties:
+      PortfolioId: !ImportValue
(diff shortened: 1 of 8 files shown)


sceptre/scipool/templates/sc-enduser-iam.yaml
```diff
@@ -57,7 +57,18 @@ Resources:
...
+          - Effect: Allow
+            Action:
+              - "ec2:*Tags"
+              - "s3:*Tags"
+              - "*:TagResource"
+              - "*:UntagResource"
+            Resource: "*"
+            Condition:
+              StringEquals:
+                "aws:ResourceTag/aws:servicecatalog:provisioningPrincipalArn": "${aws.PrincipalArn}"
```
zaro0508, line 71: tags prefixed with `aws:` are [AWS reserved tags and cannot be deleted or updated](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Tags.html#tag-restrictions), even by admin users. 
 More info at https://sagebionetworks.jira.com/browse/IT-3504?focusedCommentId=202798

jesusaurus: I understand that we can't change the value of `aws:servicecatalog:provisioningPrincipalArn`, preventing us from changing ownership of a provisioned product. But we might be able to compare that tag value with the Principal ARN from the request to allow the original product owner to change user tags like `CostCenter` on their own resources.

jesusaurus: but I'm not entirely sure this will work and I'm not sure how to test it other than merging and seeing what happens

zaro0508 (approved):

PR #1129, opened 2024-04-26
[IT-3554] Add additional Service Catalog permissions

jesusaurus:
Updating provisioned products with changes to tags is failing with permissions errors. RestrictBucketDownloadRegionFunction updates fail on 'lambda:ListTags', and the associated roles fail on'iam:UntagResource'. Allow SC roles to list tags on any lambda, and to update tags on their own provisioned resources.

The 'aws:ResourceTag/foo' condition key lets us compare against the value of the 'foo' tag on the target resource. Here we are comparing the 'provisioningPrincipalArn' SC service tag value against the 'PrincipalArn' from the request.

https://docs.aws.amazon.com/IAM/latest/UserGuide/access_tags.html#access_tags_control-resources https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_condition-keys.html#condition-keys-resourcetag https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_condition-keys.html#condition-keys-principalarn

### DIFF
--- a/sceptre/scipool/templates/sc-enduser-iam.yaml
+++ b/sceptre/scipool/templates/sc-enduser-iam.yaml
@@ -62,8 +62,7 @@ Resources:
             Resource: "*"
           - Effect: Allow
             Action:
-              - "ec2:*Tags"
-              - "s3:*Tags"
+              - "*:*Tags"
               - "*:TagResource"
               - "*:UntagResource"
             Resource: "*"
--- a/sceptre/scipool/templates/sc-enduser-iam.yaml
+++ b/sceptre/scipool/templates/sc-enduser-iam.yaml
@@ -57,7 +57,25 @@ Resources:
               - organizations:DescribeOrganization
               - servicecatalog:DescribeProvisionedProduct
               - cloudwatch:GetMetricData  # access to EC2 monitoring info
+              - ec2:DescribeImages  # service catalog status info
+              - lambda:ListTags  # service catalog tag updates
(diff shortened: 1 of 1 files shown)


sceptre/scipool/templates/sc-enduser-iam.yaml
```diff
@@ -57,7 +57,25 @@ Resources:
               - organizations:DescribeOrganization
               - servicecatalog:DescribeProvisionedProduct
               - cloudwatch:GetMetricData  # access to EC2 monitoring info
+              - ec2:DescribeImages  # service catalog status info
+              - lambda:ListTags  # service catalog tag updates
             Resource: "*"
+          - Effect: Allow
+            Action:
+              - "ec2:*Tags"
+              - "s3:*Tags"
```
zaro0508, line 66: hmm, a service catalog product template can create multiple AWS resources and all of the resources are tagged when the product is provisioned, correct?  Therefore   Should this maybe be `*:*Tags` instead?  Also don't forget about the [Scheduled Jobs](https://github.com/Sage-Bionetworks/service-catalog-library/blob/master/templates/batch/sc-batch-fargate.yaml) product.

zaro0508 (changes requested):

zaro0508:
Would this allow a SC user to change tags on a resource that they don't actually own?

zaro0508 (approved):
lets merge to test

PR #1131, opened 2024-04-26
[IT-3554] Add additional Service Catalog permissions (second attempt)

jesusaurus:
Updating provisioned products with changes to tags is failing with permissions errors. RestrictBucketDownloadRegionFunction updates fail on 'lambda:ListTags', and the associated role fails on'iam:UntagResource'. Allow SC roles to list tags on any lambda, and to update tags on their own provisioned resources.

Create a separate role in scipool-dev for manually testing permission changes before applying the changes to scipool-prod.

The 'aws:ResourceTag/foo' condition key lets us compare against the value of the 'foo' tag on the target resource. Here we are comparing the 'provisioningPrincipalArn' SC service tag value against the 'PrincipalArn' from the request.

https://docs.aws.amazon.com/IAM/latest/UserGuide/access_tags.html#access_tags_control-resources https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_condition-keys.html#condition-keys-resourcetag https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_condition-keys.html#condition-keys-principalarn


zaro0508 (approved):

zaro0508 (changes requested):
consider making one one jinja template to reduce duplication?

sceptre/scipool/templates/sc-enduser-iam.j2
```diff
@@ -58,6 +69,48 @@ Resources:
               - servicecatalog:DescribeProvisionedProduct
               - cloudwatch:GetMetricData  # access to EC2 monitoring info
             Resource: "*"
+{% if test %}
+  ScTestPolicy:
+    Type: AWS::IAM::ManagedPolicy
+    Properties:
+      Description: Expand permissions SC users.
```
zaro0508, line 76: how about a more specific description?

zaro0508 (approved):
LGTM, although policy description could be more specific to the policies.

PR #1134, opened 2024-05-02
[IT-3554] Fix jinja logic

jesusaurus:
Variable scoping doesn't allow updating a variable within an if block. Also, Sceptre is interpreting the input as a boolean value, not a string.


zaro0508 (approved):
FYI. you can validate jinja rendering with `sceptre generate` command.

jesusaurus:
Thanks, I had forgotten about that command until yesterday.

PR #1137, opened 2024-05-06
IT-3606: Activate cis-aws-foundations-benchmark SecurityHub standard

brucehoff:
SecurityHub has a list of six 'standards' which are sets of controls that it will collect findings for.  We have selected `CIS AWS Foundations Benchmark v1.4.0` for our security program.  The standard seems to have been de-activated in Security Hub.  This PR enables the standard programmatically which should ensure it remains enabled going forward.

### DIFF
--- a/org-formation/075-security-hub/security-hub.yaml
+++ b/org-formation/075-security-hub/security-hub.yaml
@@ -3,6 +3,10 @@ Description: "Setup AWS Security Hub"
 Resources:
   SecurityHub:
     Type: "AWS::SecurityHub::Hub"
+  SecurityHubStandard:
+    Type: "AWS::SecurityHub::Standard"
+    Properties:
+      StandardsArn: arn:aws:securityhub:us-east-1::standards/cis-aws-foundations-benchmark/v/1.4.0
 Outputs:
   SecurityHubArn:
     Description: "The security hub ARN"


org-formation/075-security-hub/security-hub.yaml
```diff
@@ -3,6 +3,10 @@ Description: "Setup AWS Security Hub"
 Resources:
   SecurityHub:
     Type: "AWS::SecurityHub::Hub"
+  SecurityHubStandard:
+    Type: "AWS::SecurityHub::Standard"
+    Properties:
+      StandardsArn: arn:aws:securityhub:us-east-1::standards/cis-aws-foundations-benchmark/v/1.4.0
```
zaro0508, line 9: The examples from [the docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-securityhub-standard.html#aws-resource-securityhub-standard--examples) contain parameters for partition and region which we should probably follow..

```
StandardsArn: !Sub 'arn:${AWS::Partition}:securityhub:${AWS::Region}::standards/cis-aws-foundations-benchmark/v/1.4.0'
```

org-formation/075-security-hub/security-hub.yaml
```diff
@@ -3,6 +3,10 @@ Description: "Setup AWS Security Hub"
 Resources:
   SecurityHub:
     Type: "AWS::SecurityHub::Hub"
+  SecurityHubStandard:
```
zaro0508, line 6: We currently have `aws-foundational-security-best-practices/v/1.0.0` standard  enabled.  I think this change may enable `cis-aws-foundations-benchmark/v/1.4.0` and disable the security best practices one.  I think you may need to create a resource for each standard to enable both standards.

zaro0508 (changes requested):

org-formation/075-security-hub/security-hub.yaml
brucehoff: Yes, we ca add this.

zaro0508 (changes requested):

org-formation/075-security-hub/security-hub.yaml
```diff
@@ -3,6 +3,12 @@ Description: "Setup AWS Security Hub"
...
+    Type: "AWS::SecurityHub::Standard"
+    Properties:
+      StandardsArn:
+      - arn:${AWS::Partition}:securityhub:${AWS::Region}::standards/cis-aws-foundations-benchmark/v/1.4.0
```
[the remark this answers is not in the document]
brucehoff: Based on the doc's, I believe you are correct. I did try to validate like so:

> aws cloudformation validate-template --template-body file:///Users/bhoff/Documents/organizations-infra/org-formation/075-security-hub/security-hub.yaml 

> {

    "Parameters": [],
    "Description": "Setup AWS Security Hub"
}

Maybe it's just checking the generic CF syntax, and not that `StandardsArn` can't accept a list of strings as an argument.

zaro0508 (approved):

PR #1143, opened 2024-05-07
IT-3606: Use the SecurityCentralAccount role to select the SecHub security standards

brucehoff:
The last build failed with:

> This account is currently associated with a central configuration policy, so only the administrator can access this operation.

So this PR runs the CF that selects the SecHub standards as the SecurityCentralAccount role.


zaro0508 (changes requested):
i guess this will work, you just need to also change the `accountId` default to `Default: !Ref SecurityCentralAccount`

PR #1147, opened 2024-05-09
IT-3606: switch security-hub account to SecurityCentralAccount

brucehoff:
We want the security-hub CF stack to be created in the delegated admin' account rather than in the Organizations account.

### DIFF
--- a/org-formation/075-security-hub/_tasks.yaml
+++ b/org-formation/075-security-hub/_tasks.yaml
@@ -8,7 +8,7 @@ Parameters:
   accountId:
     Type: String
     Description: The identifier from the account used to manage this service
-    Default: !Ref SecurityCentralAccount
+    Default: !Ref MasterAccount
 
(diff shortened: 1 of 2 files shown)


zaro0508 (changes requested):
you will need to [setup securitycentral account as the delegated admin for security hub](https://docs.aws.amazon.com/securityhub/latest/userguide/designate-orgs-admin-account.html#designate-admin-instructions) before securitycentral can enable and manage securityhub.

brucehoff:
> you will need to [setup securitycentral account as the delegated admin for security hub](https://docs.aws.amazon.com/securityhub/latest/userguide/designate-orgs-admin-account.html#designate-admin-instructions) before securitycentral can enable and manage securityhub.

Yes, this is done.  Thank you for linking the instructions.

xschildw (approved):

PR #1153, opened 2024-05-28
Add EKS view policy

thomasyu888:
PR Checklist:

Add the ability to view k8 deployments (port forward) deployments.
cc @BryanFauble

### DIFF
--- a/org-formation/700-aws-sso/_tasks.yaml
+++ b/org-formation/700-aws-sso/_tasks.yaml
@@ -387,6 +387,7 @@ SsoDeveloper:
       - 'arn:aws:iam::aws:policy/PowerUserAccess'
       - 'arn:aws:iam::aws:policy/AdministratorAccess-AWSElasticBeanstalk'
       - 'arn:aws:iam::aws:policy/AWSBillingReadOnlyAccess'
+      - 'arn:aws:eks::aws:cluster-access-policy/AmazonEKSViewPolicy'
     sessionDuration: 'PT12H'
     inlinePolicy: >-
       {
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,3 +1,3 @@
 PR Checklist:
-- [] Describe and explain your intentions for this change
-- [] Setup pre-commit and run the validators (info in README.md)
+- [ ] Describe and explain your intentions for this change
+- [ ] Setup pre-commit and run the validators (info in README.md)
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,3 +1,3 @@
 PR Checklist:
-[ ] Describe and explain your intentions for this change
-[ ] Setup pre-commit and run the validators (info in README.md)
+- [] Describe and explain your intentions for this change
+- [] Setup pre-commit and run the validators (info in README.md)


org-formation/700-aws-sso/_tasks.yaml
```diff
@@ -387,6 +387,7 @@ SsoDeveloper:
       - 'arn:aws:iam::aws:policy/PowerUserAccess'
       - 'arn:aws:iam::aws:policy/AdministratorAccess-AWSElasticBeanstalk'
       - 'arn:aws:iam::aws:policy/AWSBillingReadOnlyAccess'
+      - 'arn:aws:eks::aws:cluster-access-policy/AmazonEKSViewPolicy'
```
zaro0508, line 390: this is NOT a IAM policy therefore it will probably not work.  were you thinking of this instead? https://docs.aws.amazon.com/eks/latest/userguide/connector-grant-access.html

zaro0508 (changes requested):

org-formation/700-aws-sso/_tasks.yaml
BryanFauble: https://docs.aws.amazon.com/eks/latest/userguide/access-policies.html#access-policy-permissions

The idea is going to be adding this `arn:aws:eks::aws:cluster-access-policy/AmazonEKSViewPolicy` 

Cluster mode authentication is set to `API`: https://docs.aws.amazon.com/eks/latest/userguide/grant-k8s-access.html - So that all external access is controlled through AWS specific mechanisms.

For our initial use case we created a policy association for `AmazonEKSClusterAdminPolicy` to a role `role/eks_admin_role`.

I think this is the related cloud formation concept: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-eks-accessentry.html

BryanFauble: One of the things I see is that it requires a specific cluster to be specified. We're currently using `dpe-k8` for the cluster name. However, in the future we may specify other clusters (For example if we're testing out something we don't want to apply to the main k8s cluster yet)

BryanFauble: Out of curiosity - When we provision the EKS cluster. Could I at that point grant the appropriate permissions to the `Developer` role, or (Like we are doing for the `Administrator` role) allow `Developer` to assume a `role/eks_viewer_role`.

The problem I see here though is that the IAM role would be created solely within the context of `org-sagebase-dpe-prod - Administrator`

BryanFauble: For reference in the AWS UI what shows up:
image

thomasyu888: @zaro0508 what would your recommendation be here?  If we didn't have the `developer` jumpcloud role have K8 access, is there another approach?

BryanFauble: @zaro0508 It's not the same thing. The EKS cluster connector documentation you are looking at is to connect your self-hosted cluster with AWS.

For our AWS EKS cluster:
Permissions to the EKS cluster are managed entirely through IAM roles and what you are currently logged in as.

I have 2 roles I created in IAM called:
`eks_admin_role_dpe-k8-dev` and `eks_viewer_role_dpe-k8-dev`

In order to grant an Access policy to connect to the EKS cluster I am creating an https://docs.aws.amazon.com/eks/latest/APIReference/API_CreateAccessEntry.html that links the EKS cluster permissions to an IAM principal.

When I am logged in as Administrator I am able to assume `arn:aws:iam::766808016710:role/eks_admin_role_dpe-k8-dev` because of it's trust relationship:
```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Principal": {
                "AWS": "arn:aws:iam::766808016710:root"
            },
            "Action": "sts:AssumeRole"
        }
    ]
}
```

However, when logged in as Developer I cannot assume the above role - I am denied. I had tried setting up the Trust Relationship to be:
```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Principal": {
                "AWS": "arn:aws:sts::766808016710:assumed-role/AWSReservedSSO_Developer_92af2c086e7e7f38/bryan.fauble@sagebase.org"
            },
            "Action": "sts:AssumeRole"
        }
    ]
}
```

But I'm not sure AWS ARN is correct in this case. Do you have a suggestion on what it should be?

TLDR: I want Developer to be able to assume an IAM role I create that is given a policy to access the EKS cluster.

BryanFauble:
@thomasyu888 Can you close this pull request in favor of:
https://github.com/Sage-Bionetworks-IT/organizations-infra/pull/1191

PR #1158, opened 2024-06-10
AG-1469: Add OIDC role for AgoraDevAccount ECR repo

hallieswan:
PR Checklist:

We're adding a GitHub action in [Agora](https://github.com/Sage-Bionetworks/Agora/) that will 1) authenticate Docker to the Agora dev ECR using [aws ecr get-login-password](https://docs.aws.amazon.com/cli/latest/reference/ecr/get-login-password.html), 2) check whether an image already exists using [aws ecr list-images](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/ecr/list-images.html), 3) if the image does not exist, build and push the image to the Agora dev ECR repo, and 4) pull the image for use in e2e tests. Based on this, I included a policy document with permissions based on [this doc](https://github.com/aws-actions/amazon-ecr-login?tab=readme-ov-file#ecr-private) and [this doc](https://docs.aws.amazon.com/AmazonECR/latest/userguide/security_iam_id-based-policy-examples.html).

### DIFF
--- a/org-formation/650-identity-providers/_tasks.yaml
+++ b/org-formation/650-identity-providers/_tasks.yaml
@@ -816,3 +816,56 @@ GithubOidcAgoraEBDeploy:
       - !Ref AgoraDevAccount
       - !Ref AgoraProdAccount
     Region: us-east-1
+
+GithubOidcSageBionetworksAgoraDevECR:
+  Type: update-stacks
+  DependsOn: GithubOidcSageBionetworks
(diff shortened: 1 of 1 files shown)


zaro0508 (changes requested):
This PR generally looks fine however for your test that is testing thru the agora app web ui it would be easier to just use GHCR instead of ECR.  For instance, this role isn't even needed if you use GHCR.

hallieswan:
We've decided to use GitHub Container Registry instead of AWS ECR, so I'm closing this PR. Thanks for your feedback and guidance, @zaro0508!

PR #1160, opened 2024-06-12
[IT-3675] Personal AWS account for Susheel

zaro0508:
Provision a personal AWS account for Susheel

### DIFF
--- a/org-formation/organization.yaml
+++ b/org-formation/organization.yaml
@@ -655,6 +655,6 @@ Organization:
       Alias: org-sagebase-BWmErzk
       Tags:
         <<: !Include ./_default_org_tags.yaml
-        AccountOwner: susheel.varma@sagebase.org@sagebase.org
+        AccountOwner: susheel.varma@sagebase.org
         budget-alarm-threshold: 2000
         budget-alarm-threshold-email-recipient: aws.BWmErzk@sagebase.org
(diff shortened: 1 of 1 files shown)


org-formation/organization.yaml
```diff
@@ -645,3 +646,15 @@ Organization:
...
+
+  BWmErzkAccount:
+    Type: OC::ORG::Account
+    Properties:
+      AccountName: org-sagebase-BWmErzk
+      RootEmail: aws.BWmErzk@sagebase.org
+      Alias: org-sagebase-BWmErzk
+      Tags:
+        <<: !Include ./_default_org_tags.yaml
+        AccountOwner: susheel.varma@sagebase.org@sagebase.org
```
brucehoff, line 658: Typo? (`susheel.varma@sagebase.org@sagebase.org`)

org-formation/organization.yaml
```diff
@@ -138,6 +138,7 @@ Organization:
         - !Ref RestrictDnsRegistrationSCP
       Accounts:
         - !Ref BuA2aDwAccount
+        - !Ref BWmErzkAccount
```
brucehoff, line 141: is there a reason to use a random string (BWmErzk) rather than the user's name when creating a personal account?

brucehoff (changes requested):

org-formation/organization.yaml
zaro0508: in case an employee leaves we can assign the account to someone else by updating the AccountOwner tag.

org-formation/organization.yaml
zaro0508: fixed

brucehoff (approved):

PR #1164, opened 2024-06-26
IT-3730: remove DNS names for DCC Shiny app's; remove OIDC integration for GitHub repo's

brucehoff:
First step for decommissioning the DCC Validator and DCC Monitor applications hosted in AWS ECS:
Remove DNS names and OIDC integration.


zaro0508 (approved):
Generally LGTM.  Just a note that the request in IT-3730 was only to remove the apps, not actually remove the accounts as well.  Are we supposed to also remove the AWS accounts?

brucehoff:
> Generally LGTM. Just a note that the request in IT-3730 was only to remove the apps, not actually remove the accounts as well. Are we supposed to also remove the AWS accounts?

That is a good question:  I originally created the account for the express/sole purpose of hosting the app's.  So it makes sense to delete the accounts along with the app's now.

PR #1166, opened 2024-06-28
IT-3730: Remove integration with DCC* AWS accounts

brucehoff:
Final PR for decommissioning DCC app's and AWS accounts: Remove integration between org-formation and the two AWS accounts to be deleted.


zaro0508 (approved):
Remember to manually close these `member` accounts, instructions here https://docs.aws.amazon.com/accounts/latest/reference/manage-acct-closing.html

PR #1171, opened 2024-07-05
[IT-3729] Setup Guardduty s3 malware protection

zaro0508:
Enable Guardduty S3 malware protection on a Synapse bucket. Allow Guardduty to automatically tag scanned objects in the bucket with a GuardDutyMalwareScanStatus tag.

Setup AWS Guardduty Malware Protection for S3 to scan and apply a pre-defined "GuardDutyMalwareScanStatus"[1] tag
to S3 objects with possible tag values ["NO_THREATS_FOUND","THREATS_FOUND","UNSUPPORTED","FAILED"] then send a Guardduty notification to an SNS topic.

Note: The bucket being scanned should contain a bucket policy[2] to restrict access to Malicious objects.

[1] https://docs.aws.amazon.com/guardduty/latest/ug/tag-based-access-s3-malware-protection.html
[2] https://docs.aws.amazon.com/guardduty/latest/ug/tag-based-access-s3-malware-protection.html#apply-tbac-s3-malware-protection

### DIFF
--- a/org-formation/070-guard-duty/_tasks.yaml
+++ b/org-formation/070-guard-duty/_tasks.yaml
@@ -88,7 +88,7 @@ GuarddutyS3MalwareProtection:
   DefaultOrganizationBinding:
     IncludeMasterAccount: false
     Account:
-      - !Ref SynapseProdAccount
+      - !Ref SynapseDevAccount
   Parameters:
     BucketName: "TBD"
 
--- a/org-formation/070-guard-duty/_tasks.yaml
+++ b/org-formation/070-guard-duty/_tasks.yaml
@@ -73,3 +73,23 @@ GuardDutyConnectOpenData:
     MemberRootEmail: "aws.opendata@sagebase.org"
     MasterDectector: !CopyValue [ !Sub '${resourcePrefix}-${appName}-detector-id' ]
 # ==================================================================================================
+
+
+
+# ==================================================================================================
+# S3 Malware protection
+
+GuarddutyS3MalwareProtection:
+  Type: update-stacks
+  Template: ./guardduty-s3-malware-protection.yaml
+  StackName: !Sub '${resourcePrefix}-${appName}-s3-malware-protection'
+  StackDescription: Setup AWS Guardduty S3 Malware Protection
+  DefaultOrganizationBindingRegion: !Ref primaryRegion
+  DefaultOrganizationBinding:
+    IncludeMasterAccount: false
+    Account:
+      - !Ref SynapseProdAccount
+  Parameters:
+    BucketName: "TBD"
+
+# ==================================================================================================
--- a/org-formation/070-guard-duty/guardduty-s3-malware-protection.yaml
+++ b/org-formation/070-guard-duty/guardduty-s3-malware-protection.yaml
@@ -124,8 +124,3 @@ Resources:
         - Id: NotificationTopic
           Arn: !Sub 'arn:aws:sns:${AWS::Region}:${AWS::AccountId}:${SnsTopicName}'
       State: ENABLED
-Outputs:
-  SnsTopic:
-    Value: !Ref SnsTopic
-    Export:
-      Name: !Sub '${AWS::StackName}-SnsTopic'
--- a/org-formation/070-guard-duty/guardduty-s3-malware-protection.yaml
+++ b/org-formation/070-guard-duty/guardduty-s3-malware-protection.yaml
@@ -19,7 +19,7 @@ Resources:
     Properties:
       Actions:
         Tagging:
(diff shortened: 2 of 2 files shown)


org-formation/070-guard-duty/guardduty-s3-malware-protection.yaml
```diff
@@ -0,0 +1,110 @@
...
+# S3 objects with possible tag values ["NO_THREATS_FOUND","THREATS_FOUND","UNSUPPORTED","FAILED"]
+# https://docs.aws.amazon.com/guardduty/latest/ug/tag-based-access-s3-malware-protection.html
+#
+# Note: The bucket being scanned should contain a bucket policy to restrict access to Malicious objects.
+# https://docs.aws.amazon.com/guardduty/latest/ug/tag-based-access-s3-malware-protection.html#apply-tbac-s3-malware-protection
+
+AWSTemplateFormatVersion: 2010-09-09
+Description: Guardduty Malware protection plan for S3 bucket
+Parameters:
+  BucketName:
```
jesusaurus, line 11: I'm not sure if I'm reading this right, is this the bucket being scanned and so we would need to deploy a separate stack for every bucket we want to scan?

marcomarasca: I agree, additionally we have different buckets in different accounts. I would first deploy and test this to the synapse dev account.

org-formation/070-guard-duty/guardduty-s3-malware-protection.yaml
```diff
@@ -0,0 +1,147 @@
...
+          - Sid: AllowDecryptForMalwareScan
+            Effect: Allow
+            Action:
+              - 'kms:GenerateDataKey'
+              - 'kms:Decrypt'
+            Resource: !Sub 'arn:aws:kms:${AWS::Region}:${AWS::AccountId}:key/*'   # bucket is encrypted using an AWS KMS key
+            Condition:
+              StringLike:
+                'kms:ViaService': s3.us-east-1.amazonaws.com
+  SnsTopic:
```
marcomarasca, line 111: I'm not sure if creating the topic in this stack is ideal, should the topic be a parameter instead? E.g. would setting up the topic subscriptions be a problem?

zaro0508: @marcomarasca would passing in an existing SnsTopic and using that instead work for you?

org-formation/070-guard-duty/guardduty-s3-malware-protection.yaml
zaro0508: @marcomarasca instead of deploying with org-formation you can take the `guardduty-s3-malware-protection.yaml` template and deploy it to synapse dev with the synapse java deployer.

marcomarasca: I wouldn't want to have a different deployment for dev, we have to test if the integration works.

org-formation/070-guard-duty/guardduty-s3-malware-protection.yaml
marcomarasca: Sure I think that would work.

xschildw: Should we split the role from the plan (i.e. setup the account once such that multiple plans can be added later)?

org-formation/070-guard-duty/_tasks.yaml
```diff
@@ -73,3 +73,23 @@ GuardDutyConnectOpenData:
...
+GuarddutyS3MalwareProtection:
+  Type: update-stacks
+  Template: ./guardduty-s3-malware-protection.yaml
+  StackName: !Sub '${resourcePrefix}-${appName}-s3-malware-protection'
+  StackDescription: Setup AWS Guardduty S3 Malware Protection
+  DefaultOrganizationBindingRegion: !Ref primaryRegion
+  DefaultOrganizationBinding:
+    IncludeMasterAccount: false
+    Account:
+      - !Ref SynapseProdAccount
```
xschildw, line 91: I'd start with SynapseDevAccount

zaro0508: done

org-formation/070-guard-duty/guardduty-s3-malware-protection.yaml
zaro0508: I think this template is ready for you to test now @marcomarasca.  Just pass in the bucket and topic name and it should be good to go.

PR #1178, opened 2024-07-11
[IT-3758] Update organziation account name

zaro0508:
Four points requires us to apply their AWS account naming convention to our organizations account.


zaro0508:
not sure this will work but worth a shot.

zaro0508:
The automation didn't update the account name.  It had to be done from the AWS console logged in with the root user.

PR #1182, opened 2024-07-22
Add spoke VPC for dpe-sandbox

BryanFauble:
**Problem:**

1. For https://sagebionetworks.jira.com/browse/IBCDPE-933 - I am working to set up a VPC that we can use and deploy AWS resources in to. 

**Solution:**

1. This Pull request is to add the VPC that was setup for sandbox testing as a spoke for the transit gateway.

### DIFF
--- a/org-formation/710-tgw/_tasks.yaml
+++ b/org-formation/710-tgw/_tasks.yaml
@@ -118,6 +118,9 @@ TransitGatewayRoutes:
       # org-sagebase-dpe-prod
       orca-vpc:
         CIDR: "10.255.24.0/24"
+      # org-sagebase-dpe-prod
+      dpe-sandbox-vpc:
+        CIDR: "10.51.0.0/16"
 
   Parameters:
     VpcSubnetIDs:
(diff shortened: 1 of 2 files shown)


zaro0508 (commented):
> This Pull request is to add the VPC that was setup for sandbox testing as a spoke for the transit gateway.

I've read the associated jira issue however I don't understand the purpose for this.  Can you please provide information on what you are trying to accomplish with this PR?  What are you trying to achieve by adding this VPC to the transit gateway?

zaro0508:
> Should I instead use this TGW approach INSTEAD of having an internet-facing NLB? I would be setting the security group to only allow inbound connections from the attachment, and I would be updating the load balancer to be internal instead of internet facing.

No, The main purpose of a TGW is to manage traffic between VPCs.  It also helps manage traffic between local networks and  cloud based networks.  That's why we use it to manage traffic between our VPN and VPCs.  

It sounds like you have applications in one VPC and you just want to manage internet bound traffic to your VPC and eventually your private compute instances.  The correct approach for this use case is to use a load balancer as the public interface to your application.  Depending on the type of traffic your apps need you can use either an application load balancer or network load balancer.

PR #1185, opened 2024-07-23
Limit role session name

brucehoff:
The role session name overruns the 64 character limit as the github run id gets larger.  This change ensures the role session name never exceeds 64 characters.   It uses the _suffix_ to make sure the entire run id is included for uniqueness sake.


xschildw (approved):
LGTM

PR #1186, opened 2024-07-23
Cascade role-session-name to sub-workflows

brucehoff:
The error that occurred in the root workflow, `deploy.yaml` also occurs in the sub-workflows.  This change propagates the fix (managing the role session name length) to the sub-workflows.

### DIFF
--- a/.github/workflows/aws-deploy.yaml
+++ b/.github/workflows/aws-deploy.yaml
@@ -10,6 +10,9 @@ on:
       role-to-assume:
         required: true
         type: string
+      role-session-name:
+        required: true
+        type: string
       role-duration-seconds:
         type: number
         default: 1200
@@ -48,7 +51,7 @@ jobs:
         with:
           aws-region: ${{ inputs.aws-region }}
           role-to-assume: ${{ inputs.role-to-assume }}
-          role-session-name: GitHubActions-${{ github.repository_owner }}-${{ github.event.repository.name }}-${{ github.run_id }}
+          role-session-name: ${{ inputs.role-session-name }}
           role-duration-seconds: ${{ inputs.role-duration-seconds }}
       - name: Deploy with sceptre
         run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/aws-deploy.yaml
```diff
@@ -10,6 +10,9 @@ on:
       role-to-assume:
         required: true
         type: string
+      role-session-name:
```
zaro0508, line 13: this PR is not complete.  you need to actually pass the role session name in from the deploy.yaml template.

zaro0508 (changes requested):

.github/workflows/aws-deploy.yaml
brucehoff: @zaro0508 I don't understand this comment.  Isn't the change to `deploy.yaml`, below, the required change?

zaro0508: yes, you are correct.  i somehow missed it.  sorry.

.github/workflows/deploy.yaml
```diff
@@ -70,117 +70,136 @@ jobs:
     uses: "./.github/workflows/aws-deploy.yaml"
     with:
       role-to-assume: "arn:aws:iam::531805629419:role/sagebase-github-oidc-sage-bionetworks-it"
+      role-session-name: ${{ steps.role-session-name.outputs.substring }}
```
zaro0508, line 73: can you reference a `step` from a different job?  i think `step` here is not in the same scope as the one that's in `org-formation` job.  you probably need to add that `Get role session name` to every one of these jobs?

zaro0508 (changes requested):

.github/workflows/deploy.yaml
brucehoff: Yes, you are right.  I have to find a way to pass the value between jobs.

brucehoff: Done!

ConsoleCatzirl (approved):

PR #1191, opened 2024-08-15
Allow developer to assume eks developer viewer role

BryanFauble:
**Background:**

- This work is related to this ticket: https://sagebionetworks.jira.com/browse/IBCDPE-1005
- In this pull request https://github.com/Sage-Bionetworks-Workflows/eks-stack/pull/17 I have updated terraform scripts to create a role for Developer level users to assume. This role gives Developer level users cluster wide read access.

**Problem:**

1. The current configuration in this `organizations-infra` repo prevents Developer level users from assuming any role.

**Solution:**

1. Updating configuration to allow Developer level users to assume a role in the format of `role/eks-developer-viewer-role-*`. Wild card is being used here as each individual EKS cluster that might be created within an AWS account will have a role in the format of: `role/eks-developer-viewer-role-${var.cluster_name}`

**Testing:**

1. I tested assuming this new role from the `Administrator` account and verified it got the appropriate EKS cluster access.
2. I was not able to directly test assuming the role from the `Developer` account as it is not currently allowed.

### DIFF
--- a/org-formation/700-aws-sso/_tasks.yaml
+++ b/org-formation/700-aws-sso/_tasks.yaml
@@ -390,6 +390,11 @@ SsoDeveloper:
           "Version": "2012-10-17",
           "Statement": [
               {
+                  "Effect": "Allow",
+                  "Resource": "arn:aws:iam::*:role/eks-developer-viewer-role-*",
+                  "Action": "sts:AssumeRole"
+              },
+              {
                   "Effect": "Deny",
                   "Action": "sts:AssumeRole",
                   "Resource": "*"


org-formation/700-aws-sso/_tasks.yaml
```diff
@@ -389,6 +389,11 @@ SsoDeveloper:
       {
           "Version": "2012-10-17",
           "Statement": [
+              {
+                  "Effect": "Allow",
+                  "Resource": "arn:aws:iam::*:role/eks-developer-viewer-role-*",
```
BryanFauble, line 394: Currently, I added in a wild card for the AWS account ID here:
`arn:aws:iam::*:role`

Is there a way I can add in a variable here instead? This `eks-developer-viewer-role` is going to exist within the same account that this `SsoDeveloper` role is going to exist, However, I was not sure how I could add that to this `_tasks.yaml` file to add in a template variable here.

zaro0508 (changes requested):
> The current configuration in this organizations-infra repo prevents Developer level users from assuming any role.

This is on purpose.  Allowing users to assume role outside of the ones defined in Jumpcloud (via organizations-infra) would provide a back door to allow users to potentially escalate themselves to more privileged roles.

If you want to setup a role to allow AWS users access to EKS then you should create a new role in here with the policies for that.  Then we can assign Jumpcloud users to that role which would give them access to EKS.  If it makes sense we can probably also expand the generic developer role privileges to include access to EKS.

BryanFauble:
Closing in favor of https://github.com/Sage-Bionetworks-Workflows/eks-stack/pull/17#issuecomment-2292098817

PR #1197, opened 2024-08-26
Fix 4point access policy

zaro0508:
This is a fix for commit 96e340e93d8

* cannot set AWS pseudo parameters in org-formation yaml so we change it to '*'
* the policy was mixing up IAM and resource policies. iam policies do not support a principal parameter. principal is only allowed in resource policies. We remnoved the policies and moved them to the bucket policy.

### DIFF
--- a/org-formation/600-access/_tasks.yaml
+++ b/org-formation/600-access/_tasks.yaml
@@ -186,8 +186,8 @@ FPTMinimumAccess:
             "Effect": "Allow",
             "Action": ["bcm-data-exports:*"],
             "Resource": [
-                  "Fn::Sub": "arn:aws:bcm-data-exports:${AWS::Region}:${AWS::AccountId}:export/*",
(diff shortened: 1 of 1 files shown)


org-formation/600-access/_tasks.yaml
```diff
@@ -186,53 +186,15 @@ FPTMinimumAccess:
...
+                  "arn:aws:bcm-data-exports:*:*:export/*",
+                  "arn:aws:bcm-data-exports:*:*:table/COST_AND_USAGE_REPORT"
                   ]
           },
           {
             "Sid": "CurDataAccess",
             "Effect": "Allow",
             "Action": ["cur:PutReportDefinition"],
             "Resource": "*"
-          },
```
zaro0508, line 198: these are resource policies.  they have been moved to the bucket.  it was manually added to the bucket policy because this bucket was not created with cloudformation

brucehoff (approved):

PR #1203, opened 2024-08-29
Should be stack name, not vpc name to copy output

xschildw:
Previous PR (#1202) failed at accessing an output value from the infra VPC stack (' !CopyValue [!Join ["-", [!Ref primaryRegion, !FindInMap [TgwSpokesB, VpcName, !Ref CurrentAccount], 'PrivateSubnet']]] ') so we need the stack name in the map, not the vpc name.

### DIFF
--- a/org-formation/710-tgw/_tasks.yaml
+++ b/org-formation/710-tgw/_tasks.yaml
@@ -149,7 +149,7 @@ Mappings:
       '055273631518': 'snowflakevpc'   # org-sagebase-scicomp
       '420786776710': 'bridge-aux'   # org-sagebase-bridgedev
       '325565585839': 'synapse-ops-vpc-v2' # org-sagebase-synapseprod
-      '449435941126': 'sagevpc' # org-sagebase-synapsedev
+      '449435941126': 'vpc' # org-sagebase-synapsedev
   TgwSpokesSynapseDev:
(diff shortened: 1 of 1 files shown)


org-formation/710-tgw/_tasks.yaml
```diff
@@ -149,7 +149,7 @@ Mappings:
       '055273631518': 'snowflakevpc'   # org-sagebase-scicomp
       '420786776710': 'bridge-aux'   # org-sagebase-bridgedev
       '325565585839': 'synapse-ops-vpc-v2' # org-sagebase-synapseprod
-      '449435941126': 'sagevpc' # org-sagebase-synapsedev
+      '449435941126': 'vpc' # org-sagebase-synapsedev
```
zaro0508, line 152: could it have been renamed manually from the aws console?  can you figure out what the name should be from the template that was used to create the VPC?

zaro0508 (approved):
i don't see anything wrong with this but you should check the template that created the VPC and verify the name it gives to this VPC matches what you are setting it to.

org-formation/710-tgw/_tasks.yaml
xschildw: It was passed as param.

PR #1204, opened 2024-08-30
[IT-3758] Tweak permissions for four points

zaro0508:
Four points tried to setup CUR reports and noticed that the service requires a `condition` setting.


xschildw (approved):
OK

PR #1206, opened 2024-09-04
PLFM-8554: Install Stack Armor Nessus scanner on all EC2 instances in Synapse account

brucehoff:
This PR adds a template to create an SSM Maintenance Window and uses it to run the  Stack Armor scanner installer each day, ensuring that the scanner is installed (or updated to a new version) on any machines deployed to the target account.

The installation script is a parameter, and we 'point to' the Nessus script, stored in GitHub.

The Nessus script requires a secret key.  We store the key in SSM Parameter store and, at execution time, retrieve it and pass it to the script as an environment variable.  The CF template takes a list of key/value pairs as input where the keys are SSM Parameter Store keys and the values are the names of environment variables in the scrip to which the secret values should be assigned.  In the immediate case the Parameter Store key is `/stack-armor/install-nessus-key` and its value is assigned to `STACK_ARMOR_KEY`.

### DIFF
--- a/org-formation/090-systems-manager/ScheduledMaintenance.yaml
+++ b/org-formation/090-systems-manager/ScheduledMaintenance.yaml
@@ -71,14 +71,14 @@ Resources:
           name: "runShellScript"
           inputs:
             runCommand:
-            - SCRIPT_PATH=/tmp/script.sh
             - SSM_TO_ENV_MAP=( {{ SsmParameterNameToEnvvarMap }} )
             - for KV in ${SSM_TO_ENV_MAP[@]} ; do
-            -   ${KV##*:}=$(aws --region {{ Region }} ssm get-parameters --names ${KV%%:*} --with-decryption --query Parameters[0].Value --output text)
+            -   export ${KV##*:}=$(aws --region {{ Region }} ssm get-parameters --names ${KV%%:*} --with-decryption --query Parameters[0].Value --output text)
             - done
+            - SCRIPT_PATH=/tmp/script.sh
             - wget -O $SCRIPT_PATH {{ ScriptUrl }}
             - chmod +x $SCRIPT_PATH
-            - $SCRIPT_PATH
+            - sudo $SCRIPT_PATH
 
   # Create Maintenance Window
   MaintenanceWindow:
@@ -97,13 +97,9 @@ Resources:
       WindowId: !Ref MaintenanceWindow
       ResourceType: "INSTANCE"
       Targets:
-        - Key: "tag:ssm-group"
+        - Key: "InstanceIds"
           Values:
-            - "foo"
-# TODO run on all instances
-#        - Key: "InstanceIds"
-#          Values:
-#            - "*"
+            - "*"
       Name: "SSMGroupTarget"
 
   # Create Maintenance Window Task
--- a/org-formation/090-systems-manager/ScheduledMaintenance.yaml
+++ b/org-formation/090-systems-manager/ScheduledMaintenance.yaml
@@ -16,41 +16,15 @@ Parameters:
     Description: >-
       space delimited list of key:value, where key is an SSM Parameter Store
       key and value is the shell envvar name to assign, e.g., /foo/bar:ENVVAR1 /bar/baz:ENVVAR2
+  TargetAccountId:
+    Type: 'String'
+    Description: The target account in which to run the Maintenance Task    
 Resources:
-  # Create IAM Role for SSM
-  SSMExecutionRole:
-    Type: "AWS::IAM::Role"
-    Properties:
-      AssumeRolePolicyDocument:
-        Version: '2012-10-17'
-        Statement:
-          - Effect: Allow
-            Principal:
-              Service: "ssm.amazonaws.com"
-            Action: "sts:AssumeRole"
-      Policies:
-        - PolicyName: "SSMExecutionPolicy"
-          PolicyDocument:
-            Version: '2012-10-17'
-            Statement:
-              - Effect: Allow
-                Action:
-                  - "ec2:DescribeInstances"
-                  - "ssm:SendCommand"
-                Resource: "*"
-        - PolicyName: "SSMParameterPolicy"
-          PolicyDocument:
-            Version: '2012-10-17'
-            Statement:
-              - Effect: Allow
-                Action:
-                  - "ssm:GetParameters"
-                Resource:
-                - !Join [ '', ['arn:', !Ref 'AWS::Partition', ':ssm:', !Ref 'AWS::Region', ':', !Ref 'AWS::AccountId', '*'] ]
   SSMDocument:
     Type: "AWS::SSM::Document"
     Properties:
       DocumentType: "Command"
+      Name: "run-parameterized-script"
       Content:
         schemaVersion: "2.2"
         description: "Run a script on EC2 instances"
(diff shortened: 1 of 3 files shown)


org-formation/090-systems-manager/_tasks.yaml
```diff
@@ -60,3 +60,17 @@ PatchMembers:
     IncludeMasterAccount: true
   Parameters:
     ManagementAccountNumber: !Ref accountId
+
+StackArmorNessusScanner:
+  Type: update-stacks
+  Template: ScheduledMaintenance.yaml
+  StackName: !Sub '${resourcePrefix}-${appName}-stackarmor-nessus-scsnner'
```
xschildw, line 67: typo

org-formation/090-systems-manager/ScheduledMaintenance.yaml
```diff
@@ -0,0 +1,127 @@
...
+    Type: 'String'
+    Description: 'Schedule to run script'
+    Default: "rate(24 hours)"
+  ScriptURL:
+    Type: 'String'
+    Description: 'URL for the script to run'
+  ScriptParameters:
+    Type: 'String'
+    Description: >-
+      space delimited list of key:value, where key is an SSM Parameter Store
```
zaro0508, line 17: can you please provide an example?

org-formation/090-systems-manager/ScheduledMaintenance.yaml
```diff
@@ -0,0 +1,127 @@
...
+            type: String
+            description: >-
+              space delimited list of key:value, where key is an SSM Parameter Store
+              key and value is the shell envvar name to assign.
+        mainSteps:
+        - action: "aws:runShellScript"
+          name: "runShellScript"
+          inputs:
+            runCommand:
+            - SSM_TO_ENV_MAP=( {{ SsmParameterNameToEnvvarMap }} )
```
zaro0508, line 74: what does `SsmParameterNameToEnvvarMap` look like?

org-formation/090-systems-manager/ScheduledMaintenance.yaml
```diff
@@ -0,0 +1,127 @@
...
+      AllowUnassociatedTargets: False
+
+  # Create Maintenance Window Target
+  MaintenanceWindowTarget:
+    Type: "AWS::SSM::MaintenanceWindowTarget"
+    Properties:
+      WindowId: !Ref MaintenanceWindow
+      ResourceType: "INSTANCE"
+      Targets:
+        - Key: "InstanceIds"
```
zaro0508, line 100: i think targeting by tags would be better because it's much more flexible.

org-formation/090-systems-manager/_tasks.yaml
```diff
@@ -60,3 +60,17 @@ PatchMembers:
...
+  StackDescription: Install Nessus scanner on instances in account
+  DefaultOrganizationBindingRegion: !Ref primaryRegion
+  DefaultOrganizationBinding:
+    Accounts:
+      - !Ref SynapseProdAccount
```
zaro0508, line 72: i think running from the central `SysmanCentralAccount` would be better for organizing runbooks.  it also allows running scripts from a central location in multiple accounts in the organization.

zaro0508 (changes requested):

xschildw (approved):
Nice! Makes sense to me.

org-formation/090-systems-manager/ScheduledMaintenance.yaml
```diff
@@ -0,0 +1,127 @@
...
+              Service: "ssm.amazonaws.com"
+            Action: "sts:AssumeRole"
+      Policies:
+        - PolicyName: "SSMExecutionPolicy"
+          PolicyDocument:
+            Version: '2012-10-17'
+            Statement:
+              - Effect: Allow
+                Action:
+                  - "ec2:DescribeInstances"
```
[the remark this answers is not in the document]
brucehoff: Thanks, but I already ran this in a dev' account and it worked great.  No additional permissions needed.

org-formation/090-systems-manager/ScheduledMaintenance.yaml
brucehoff: Yes!

org-formation/090-systems-manager/ScheduledMaintenance.yaml
brucehoff: I added an example to the parameter 'description'.

org-formation/090-systems-manager/ScheduledMaintenance.yaml
brucehoff: While targeting by tags works, I want to make sure that instances are scanned by default, not only when they are correctly tagged.  I want to catch the case in which someone provisions an instance, doesn't set it up the way we'd like, and then doesn't patch vulnerabilities.  If we require a specific tag to scan the instance then we could have a 'blind spot.'

org-formation/090-systems-manager/_tasks.yaml
brucehoff: I'm not sure what it means to "run" an SSM Maintenance window from a central account. Can you please explain?

zaro0508: `SysmanCentralAccount` is a delegated administrator account for systems manager feature which means that it has access to all instances in all member AWS accounts in our organization.  The way to run automation across multiple accounts is documented in https://docs.aws.amazon.com/systems-manager/latest/userguide/running-automations-multiple-accounts-regions.html

This is how we are able to setup one central aws account to do OS patching to instances in all accounts.   You may want to take another look at the [templates we used to setup patch manager and associated roles](https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/org-formation/090-systems-manager/_tasks.yaml).

zaro0508 (changes requested):

org-formation/090-systems-manager/_tasks.yaml
zaro0508: SysmanCentralAccount is a delegated administrator account for systems manager feature which means that it has access to all instances in all member AWS accounts in our organization. The way to run automation across multiple accounts is documented in https://docs.aws.amazon.com/systems-manager/latest/userguide/running-automations-multiple-accounts-regions.html

This is how we are able to setup one central aws account to do OS patching to instances in all accounts. You may want to take another look at the [templates we used to setup patch manager and associated roles](https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/org-formation/090-systems-manager/_tasks.yaml).

brucehoff: The AWS doc' you reference says:

> When you run an automation across multiple Regions and accounts, you target resources by using tags or the name of an AWS resource group. The resource group must exist in each target account and Region. The resource group name must be the same in each target account and Region. The automation fails to run on those resources that don't have the specified tag or that aren't included in the specified resource group.

So this approach can't run on all instances in a selected account by default.  The instances would have to be explicitly tagged or else be explicitly added to a resource group. Is this what we do to make patch manager work on all instances, tag or group the instances to patch?

xschildw: Can we setup a resource group based only on the type of resource (AWS::Ec2::Instances) with no tag? From what I see the patchManager depends on 2 tags (present in scipoolprod, but not in synapseprod).

zaro0508: > Is this what we do to make patch manager work on all instances, tag or group the instances to patch?

yes, we setup a resource group to target where to run the patches.  The definition of the patch resource group is here  https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/org-formation/090-systems-manager/patch-resource-group.yaml

brucehoff:
I am closing this PR. It is replaced by https://github.com/Sage-Bionetworks-IT/organizations-infra/pull/1222

PR #1208, opened 2024-09-10
IT-3773: Add perms for portal buckets/distributions

xschildw:
This PR adds permissions to push to S3 bucket to the oidc role for the Synapse monorepo repository.


org-formation/650-identity-providers/_tasks.yaml
```diff
@@ -504,7 +504,31 @@ GithubOidcSageBionetworksWebMonorepoInfra:
               "Resource": [ "arn:aws:s3:::prod.accounts.sagebionetworks.org",
                             "arn:aws:s3:::staging.accounts.sagebionetworks.org",
                             "arn:aws:s3:::prod.accounts.synapse.org",
-                            "arn:aws:s3:::staging.accounts.synapse.org"
+                            "arn:aws:s3:::staging.accounts.synapse.org",
+                            "arn:aws:s3:::prod-adknowledgeportalsynapse-org-websitebucket-1wcys549ufmd",
```
zaro0508, line 508: would it be possible to wildcard some of these?  for example `arn:aws:s3:::*-adknowledgeportalsynapse-org-websitebucket-*`

zaro0508 (approved):
seems fine, but might be better to use wild cards in some places

PR #1218, opened 2024-09-17
PLFM-8102: Add oidc perms to push image to repo and deploy lambda from image

xschildw:
This PR sets up OIDC role with permissions to push a docker image to a repo from the nbconvert GH repo, and create a lambda from that image.

### DIFF
--- a/org-formation/650-identity-providers/_tasks.yaml
+++ b/org-formation/650-identity-providers/_tasks.yaml
@@ -810,8 +810,11 @@ GithubOidcNbConvertDeploy:
     ProviderRoleName: !Sub ${resourcePrefix}-${appName}-nbconvert-deploy
     MaxSessionDuration: 7200
     ManagedPolicyArns:
-      - "arn:aws:iam::aws:policy/AdministratorAccess"
-      - "arn:aws:iam::aws:policy/AWSKeyManagementServicePowerUser"
+      - "arn:aws:iam::aws:policy/AmazonEC2ContainerRegistryFullAccess"
+      - "arn:aws:iam::aws:policy/AmazonLambdaFullAccess"
+      - "arn:aws:iam::aws:policy/CloudWatchLogsFullAccess"
+      - "arn:aws:iam::aws:policy/IAMFullAccess"
+      - "arn:aws:iam::aws:policy/CloudFormationFullAccess"
   TemplatingContext:
     GitHubOrg: "Sage-Bionetworks"
     Repositories:
--- a/org-formation/650-identity-providers/_tasks.yaml
+++ b/org-formation/650-identity-providers/_tasks.yaml
@@ -800,6 +800,51 @@ GithubOidcOpenChallengesDeploy:
       - !Ref ITSandboxAccount
     Region: us-east-1
 
+GithubOidcNbConvertDeploy:
+  Type: update-stacks
+  DependsOn: GithubOidcSageBionetworks
+  Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.7.6/templates/IAM/github-oidc-provider.j2
+  StackName: !Sub ${resourcePrefix}-${appName}-nbconvert-deploy
+  Parameters:
+    ProviderArn: !CopyValue [ !Sub '${resourcePrefix}-${appName}-ProviderArn' ]
+    ProviderRoleName: !Sub ${resourcePrefix}-${appName}-nbconvert-deploy
+    MaxSessionDuration: 7200
+    PolicyDocument: >-
+      {
+          "Version": "2012-10-17",
+          "Statement": [
+              {
+                  "Effect": "Allow",
+                  "Action": [
+                      "ecr:CompleteLayerUpload",
+                      "ecr:UploadLayerPart",
+                      "ecr:InitiateLayerUpload",
+                      "ecr:BatchCheckLayerAvailability",
+                      "ecr:PutImage"
+                  ],
+                  "Resource": [
+                      "arn:aws:ecr:region:449435941126:repository/dev-nbconvert",
+                      "arn:aws:ecr:region:325565585839:repository/prod-nbconvert"
+                  ]
+              },
+              {
+                  "Effect": "Allow",
+                  "Action": "ecr:GetAuthorizationToken",
+                  "Resource": "*"
+              }
+          ]
+      }
+  TemplatingContext:
+    GitHubOrg: "Sage-Bionetworks"
+    Repositories:
+      - name: "nbconvert-webapp"
+        branches: ["master"]
+  DefaultOrganizationBinding:
+    Account:
+      - !Ref SynapseDevAccount
+      - !Ref SynapseProdAccount
+    Region: us-east-1
+
 ############################### Managed Policies ###############################
 # Managed policies used in github OIDC providers
 # Note: Managed policies can be used as work around for the AWS cloudformation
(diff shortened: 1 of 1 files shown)


org-formation/650-identity-providers/_tasks.yaml
```diff
@@ -800,6 +800,54 @@ GithubOidcOpenChallengesDeploy:
...
+                      "ecr:CompleteLayerUpload",
+                      "ecr:UploadLayerPart",
+                      "ecr:InitiateLayerUpload",
+                      "ecr:BatchCheckLayerAvailability",
+                      "ecr:PutImage"
+                  ],
+                  "Resource": [
+                      "arn:aws:ecr:region:449435941126:repository/nbconvert",
+                      "arn:aws:ecr:region:449435941126:repository/dev-nbconvert",
+                      "arn:aws:ecr:region:325565585839:repository/prod-nbconvert"
```
zaro0508, line 830: does this work?  when this is deployed to `SynapseDevAccount` then i think `arn:aws:ecr:region:325565585839:repository/prod-nbconvert` is not available because it's in `SynapseProdAccount`?

org-formation/650-identity-providers/_tasks.yaml
```diff
@@ -800,6 +800,54 @@ GithubOidcOpenChallengesDeploy:
...
+GithubOidcNbConvertDeploy:
+  Type: update-stacks
+  DependsOn: GithubOidcSageBionetworks
+  Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.7.6/templates/IAM/github-oidc-provider.j2
+  StackName: !Sub ${resourcePrefix}-${appName}-nbconvert-deploy
+  Parameters:
+    ProviderArn: !CopyValue [ !Sub '${resourcePrefix}-${appName}-ProviderArn' ]
+    ProviderRoleName: !Sub ${resourcePrefix}-${appName}-nbconvert-deploy
+    MaxSessionDuration: 7200
+    PolicyDocument: >-
```
zaro0508, line 814: why not just give the CI the `AmazonEC2ContainerRegistryFullAccess` managed policy?

zaro0508 (changes requested):

org-formation/650-identity-providers/_tasks.yaml
xschildw: I'm actually not sure, do you think there's a check either at deploy time or at run time for all the values? My thinking was there's not and even if I tried to push cross account (which technically is allowed by this policy) I would be blocked by not having the cross account permission set up.
Ideally, I would just have used wildcards but they're not permitted as I found out in a previous PR.
If this does not work, then I would have to split the policy to match the account.

org-formation/650-identity-providers/_tasks.yaml
xschildw: I originally thought about giving more permissions in case I want to deploy the lambda itself from GH CI then decided to clamp it down since I only want to push the image from GH CI (and deploy the lambda from the regular Synapse CI).
It would simplify management...

org-formation/650-identity-providers/_tasks.yaml
```diff
@@ -800,6 +800,32 @@ GithubOidcOpenChallengesDeploy:
...
+  Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.7.6/templates/IAM/github-oidc-provider.j2
+  StackName: !Sub ${resourcePrefix}-${appName}-nbconvert-deploy
+  Parameters:
+    ProviderArn: !CopyValue [ !Sub '${resourcePrefix}-${appName}-ProviderArn' ]
+    ProviderRoleName: !Sub ${resourcePrefix}-${appName}-nbconvert-deploy
+    MaxSessionDuration: 7200
+    ManagedPolicyArns:
+      - "arn:aws:iam::aws:policy/AmazonEC2ContainerRegistryFullAccess"
+      - "arn:aws:iam::aws:policy/AmazonLambdaFullAccess"
+      - "arn:aws:iam::aws:policy/CloudWatchLogsFullAccess"
```
xschildw, line 817: I think I also need some IAM permissions here, but not keen on giving full access...

zaro0508 (approved):

org-formation/650-identity-providers/_tasks.yaml
```diff
@@ -812,9 +810,8 @@ GithubOidcNbConvertDeploy:
     ProviderRoleName: !Sub ${resourcePrefix}-${appName}-nbconvert-deploy
     MaxSessionDuration: 7200
     ManagedPolicyArns:
-      - "arn:aws:iam::aws:policy/AmazonEC2ContainerRegistryFullAccess"
-      - "arn:aws:iam::aws:policy/AmazonLambdaFullAccess"
-      - "arn:aws:iam::aws:policy/CloudWatchLogsFullAccess"
+      - "arn:aws:iam::aws:policy/AdministratorAccess"
```
zaro0508, line 813: hmm, In this case where you know this OIDC only needs to push to AWS ECR then you should probably scope it to just allow that.  I like the previously scoped permissions set you had better.

zaro0508 (changes requested):

org-formation/650-identity-providers/_tasks.yaml
xschildw: OK but this may gonna take a few iterations. I know I need at least IAM to modify some policies/statements so I'll add that.

zaro0508: > OK but this may gonna take a few iterations. I know I need at least IAM to modify some policies/statements so I'll add that.

The title of this PR is `Add oidc perms to push image to repo` so I thought that's all this was needed for?  if not, what else do you plan to deploy?

zaro0508: ahh so you are actually deploying a lambda not just pushing to ECR?  Depending on the lambda you'll probably want admin because the lambda can setup a bunch of AWS resources.  

I'm just wondering why you chose to run lambda in docker instead of deploying just a bare lambda?  it seems like an over complicated solution unless there's a good reason.  i think this is a good comparison of the two approaches. https://www.knowbe4.com/careers/blogs/engineering/containerized-lambdas-vs-binary-pros-and-conscon

xschildw: I tried using the python runtime but the python library we're using is missing a binary dependency so I moved to docker so I was able to add the dependency... 
I looked at the synthesized template and see a role, a function and a function permission (so IAM and Lambda) but as you say the deployer role could/should need more permissions (hence best guess about additional ones).
Best practice is nice but they sure don't make it easy!
So back to admin or we try to at least restrict the services?

brucehoff: @zaro0508 Thank you for your suggestion.  I am happy to make changes in support of the principle of least access. 
 But, why do you say that `GithubOidcSageBionetworksSynapseDockerRegistry` "only needs to push to AWS ECR"?  E.g., it needs to access EC2 (incl. ALB), Route53, Certificate Manager, Secrets Manager, tagging.  We would need some methodology to determine what permissions a stack deployment actually needs.

zaro0508: sorry, my mistake @brucehoff.  The name of the resource threw me off, it made it sound like this OIDC was just for deploying stuff to ECR.  I see it's for Sage-Bionetworks/synapse-docker-registry which is doing a lot more.

zaro0508 (approved):

PR #1222, opened 2024-09-24
IT-3860: Add automation to run Stack Armor agent installation script on instances tagged appropriately

brucehoff:
Add automation to run Stack Armor agent installation script on instances tagged appropriately.

This PR adds a template to use SSM Automation to run the Stack Armor scanner installer each day, which ensures that the scanner is installed (or updated to a new version) on appropriately tagged EC2 instances.  It runs from the Systems Manager central account and is applied to all member accounts.

The new template is based on the existing template `Scheduled-Patch-Automation.yaml` which runs a patching job, but this template runs the shell script whose URL is passed as a parameter. We 'point to' the Nessus script, stored in GitHub. Instead of defining a new 'resource group' we simply use tags to identify the instances to patch.

The Nessus script requires a secret key. We store the key in SSM Parameter store and, at execution time, retrieve it and pass it to the script as an environment variable.

### DIFF
--- a/org-formation/090-systems-manager/Scheduled-Script-Automation.yaml
+++ b/org-formation/090-systems-manager/Scheduled-Script-Automation.yaml
@@ -58,6 +58,10 @@ Parameters:
   TargetTagValue:
     Type: String
     Description: Value of the tag for instances to target.
+  ExecutionTimeoutSeconds:
+    Type: Number
+    Description:  Time limit for running the automation, in seconds
+    Default: 120
   TargetLocationMaxConcurrency:
     Type: String
     Default: '1'
@@ -168,6 +172,7 @@ Resources:
               TargetRegionIdsArray = b.split(",")
               TargetTagName=os.environ['TargetTagName']
               TargetTagValue=os.environ['TargetTagValue']
+              ExecutionTimeoutSeconds=os.environ['ExecutionTimeoutSeconds']
               TargetLocationMaxConcurrency=os.environ['TargetLocationMaxConcurrency']
               TargetLocationMaxErrors=os.environ['TargetLocationMaxErrors']
               ExecutionRoleName=os.environ['ExecutionRoleName']
@@ -184,6 +189,7 @@ Resources:
                       'TargetTagName' : [TargetTagName],
                       'TargetTagValue' : [TargetTagValue],
                       'ScriptUrl' : [ScriptUrl],
+                      'ExecutionTimeoutSeconds': [ExecutionTimeoutSeconds],
                       'ScriptParameters' : [ScriptParameters]
                   },
                   TargetLocations=[
@@ -214,6 +220,7 @@ Resources:
             ScriptParameters: !Ref ScriptParameters
             TargetTagName: !Ref TargetTagName
             TargetTagValue: !Ref TargetTagValue
+            ExecutionTimeoutSeconds: !Ref ExecutionTimeoutSeconds
       FunctionName: MultiAccountScriptExecution
       Handler: index.handler
       Role: !GetAtt AWSLambdaSSMMultiAccountScriptExecutionRole.Arn
@@ -270,10 +277,14 @@ Resources:
           TargetTagValue:
             type: String
             description: value of the tag marking instances on which to run the script
+          ExecutionTimeoutSeconds:
+            type: Number
+            description:  Time limit for running the automation, in seconds
+            default: 120
         mainSteps:
         - name: "runShellScript"
           action: "aws:runShellScript"
-          timeoutSeconds: 7200
+          timeoutSeconds: {{ ExecutionTimeoutSeconds }}
           onFailure: Abort
           inputs:
             Targets:
--- a/org-formation/090-systems-manager/Scheduled-Script-Automation.yaml
+++ b/org-formation/090-systems-manager/Scheduled-Script-Automation.yaml
@@ -264,6 +264,9 @@ Resources:
         description: "Run a parameterized script on EC2 instances"
         assumeRole: '{{ AutomationAssumeRole }}'
         parameters:
+          AutomationAssumeRole:
+            type: String
+            description: The ARN of the Automation service role to assume.
           ScriptUrl:
             type: String
             description: "The URL of the script to run"
--- a/org-formation/090-systems-manager/Scheduled-Script-Automation.yaml
+++ b/org-formation/090-systems-manager/Scheduled-Script-Automation.yaml
@@ -190,7 +190,7 @@ Resources:
                       'TargetTagValue' : [TargetTagValue],
                       'ScriptUrl' : [ScriptUrl],
                       'ExecutionTimeoutSeconds': [ExecutionTimeoutSeconds],
-                      'ScriptParameters' : [ScriptParameters]
+                      'SsmParameterNameToEnvvarMap' : [ScriptParameters]
                   },
                   TargetLocations=[
                       {
--- a/org-formation/090-systems-manager/Scheduled-Script-Automation.yaml
+++ b/org-formation/090-systems-manager/Scheduled-Script-Automation.yaml
@@ -257,11 +257,12 @@ Resources:
   AutomationDocumentScriptExecution:
     Type: "AWS::SSM::Document"
     Properties:
-      DocumentType: "Command"
-      Name: "Automation-Run-Script"
+      DocumentType: "Command" # Automation
+      Name: "Automation-Run-Script-5" # TODO
       Content:
-        schemaVersion: "2.2"
+        schemaVersion: "2.2" # 0.3
         description: "Run a parameterized script on EC2 instances"
+#        assumeRole: '{{ AutomationAssumeRole }}' TODO
         parameters:
           ScriptUrl:
             type: String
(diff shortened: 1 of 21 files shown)


org-formation/090-systems-manager/Scheduled-Script-Automation.yaml
```diff
@@ -0,0 +1,297 @@
...
+          TargetTagName:
+            type: String
+            description: name of the tag marking instances on which to run the script
+          TargetTagValue:
+            type: String
+            description: value of the tag marking instances on which to run the script
+        mainSteps:
+        - name: "runShellScript"
+          action: "aws:runShellScript"
+          timeoutSeconds: 7200
```
zaro0508, line 276: this 2hr timeout makes sense when running OS patches however it seems excessive when installing the nessus agent.  consider parameterizing this timeout?

org-formation/090-systems-manager/Scheduled-Script-Automation.yaml
```diff
@@ -0,0 +1,297 @@
...
+              Fn::Sub: arn:${AWS::Partition}:iam::*:role/AWS-SystemsManager-AutomationExecutionRole
+          - Effect: Allow
+            Action:
+            - organizations:ListAccountsForParent
+            Resource:
+            - "*"
+  AutomationDocumentScriptExecution:
+    Type: "AWS::SSM::Document"
+    Properties:
+      DocumentType: "Command"
```
zaro0508, line 253: It looks like you copy and pasted most of this from  the `Automation` DocumentType in `Scheduled-Patch-Automation.yaml` file?  I don't think the schema for that is the same as the `Command` DocumentType.  I think you will need to update to match the schema for the `Command` DocumentType https://docs.aws.amazon.com/systems-manager/latest/userguide/documents-command-ssm-plugin-reference.html#shared-inputs

org-formation/090-systems-manager/Scheduled-Script-Automation.yaml
```diff
@@ -0,0 +1,297 @@
...
+      Name: "Automation-Run-Script"
+      Content:
+        schemaVersion: "2.2"
+        description: "Run a parameterized script on EC2 instances"
+        parameters:
+          ScriptUrl:
+            type: String
+            description: "The URL of the script to run"
+          SsmParameterNameToEnvvarMap:
```
zaro0508, line 262: hmm, i don't see the connection from `ScriptParameters` to `SsmParameterNameToEnvvarMap`.   I mean `runCommand` processes vars in SsmParameterNameToEnvvarMap but how does the passed in vars in ScriptParameters get passed to SsmParameterNameToEnvvarMap?

zaro0508 (changes requested):
I like this approach of trying to generalize script execution with a SSM document however I think are are also a bunch of assumptions that are not obvious from just looking at the code.  I think it would be beneficial to document this in the README file.

org-formation/090-systems-manager/Scheduled-Script-Automation.yaml
```diff
@@ -0,0 +1,297 @@
...
+      The number of errors that are allowed before the system stops initiating the automation on additional targets. You can specify either an absolute number of errors, for example 10, or a percentage of the target set, for example 10%. The default value is 10%.
+  ScriptUrl:
+    Type: String
+    Description: The URL of the script to run on targeted instances. Must be publicly readable.
+  ScriptParameters:
+    Type: String
+    Description: >-
+      space delimited list of environment name-value pairs in the format name:value,
+      e.g., ENVNAME1:envval1 ENVNAME2:envval2
+  TargetTagName:
```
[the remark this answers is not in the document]
brucehoff: I'm not sure why you are asking this question, since your own solution for patching relies on [tags](https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/org-formation/090-systems-manager/patch-resource-group.yaml#L23-L29) as well.  Moreover, all options I've seen for targeting resources with SSM automation either depend on tags or on all the resources being in the same CF stack.  But to answer your question:  My thought is to have the Synapse team add the required tag to the instances deployed by the Synapse stack builder and if we want Stack Armor to scan any machines besides the Synapse EC2s then we would include the tag in the deployment infra' for those stacks as well.

brucehoff:
> I like this approach of trying to generalize script execution with a SSM document however I think are are also a bunch of assumptions that are not obvious from just looking at the code. I think it would be beneficial to document this in the README file.

Done!  If you feel there are still unaddressed 'assumptions' kindly let me know.

org-formation/090-systems-manager/Scheduled-Script-Automation.yaml
brucehoff: According to [the doc's](https://docs.aws.amazon.com/systems-manager/latest/userguide/documents-schemas-features.html) the schemaVersion 2.2 for Documents has the fields, `description`, `parameters`, and `mainSteps`, which is what we are using here.  (Your code uses schemaVersion 0.3, which is [required](https://docs.aws.amazon.com/systems-manager/latest/userguide/documents-schemas-features.html) for Documents of type Automation.)

If I am missing something (if you see something wrong in this code) please say what the mistake is.

org-formation/090-systems-manager/Scheduled-Script-Automation.yaml
brucehoff: Good catch:  On line 193
                      'ScriptParameters' : [ScriptParameters]
should be
                      'SsmParameterNameToEnvvarMap' : [ScriptParameters]
It's now updated in the PR.

org-formation/090-systems-manager/Scheduled-Script-Automation.yaml
```diff
@@ -0,0 +1,308 @@
...
+        mainSteps:
+        - name: "runShellScript"
+          action: "aws:runShellScript"
+          timeoutSeconds: {{ ExecutionTimeoutSeconds }}
+          onFailure: Abort
+          inputs:
+            Targets:
+            - Key: 'tag:{{ TargetTagName }}'
+              Values:
+                - '{{ TargetTagValue }}'
```
zaro0508, line 293: for example, you are using the `Command` document `aws:runShellScript` here and from the [documentation](https://docs.aws.amazon.com/systems-manager/latest/userguide/documents-command-ssm-plugin-reference.html#aws-runShellScript) for it there is no `Targets:` parameter for `inputs`.    In fact the only place where I see `Targets` being available is for the [aws:runCommand action](https://docs.aws.amazon.com/systems-manager/latest/userguide/automation-action-runcommand.html)

zaro0508 (changes requested):

org-formation/090-systems-manager/Scheduled-Script-Automation.yaml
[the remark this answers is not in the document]
brucehoff: > depending on human users to apply the right tag every time they deploy a resource is brittle at best

The expectation is not that human users apply the tag manually but that deployment (CF) scripts do it automatically.  I didn't realize that was not clear, so I have now tried to clarify that in the comments.

org-formation/090-systems-manager/Scheduled-Script-Automation.yaml
brucehoff: I think this document uses the correct plug-ins, i.e., this document is a `Command` document and uses the `aws:runShellScript` plugin, which is listed as a valid plug-in for the `Command` document type in the document you linked.  So I don't see any problem there.   However I see another potential problem:  [This](https://docs.aws.amazon.com/systems-manager/latest/userguide/systems-manager-automation.html) doc' says that Automation uses Automation runbooks and, "Runbooks are Systems Manager documents of type Automation, as opposed to Command, Policy, Session documents. "  So I think I have to switch from a `Command` document  type to an `Automation` document type, which would involve changing the document parameters.

org-formation/090-systems-manager/Scheduled-Script-Automation.yaml
brucehoff: As stated above, I think for this to work the Document has to be an `Automation` document.  I will update accordingly.

org-formation/090-systems-manager/Scheduled-Script-Automation.yaml
brucehoff: @zaro0508 The PR is now updated to use document type `Automation`, the compatible plug-in type, `aws:runCommand`, and the Document, `AWS-RunShellScript`.  I tested the CF template in a dev account to make sure the parameters are right.

org-formation/090-systems-manager/Scheduled-Script-Automation.yaml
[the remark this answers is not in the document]
brucehoff: > if someone deploys with a template that does not set the correct tags?

Then the automation will not run on the instance.  

I feel like you are trying to hint at an alternative.  If so, will you say what it is?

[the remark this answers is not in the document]
zaro0508: my suggestion was in my previous comment, 
```In this use case I think a better solution is to make AWS automatically tag every EC2 instance with a special tag (i.e. nessus-monitoring) then have this solution install the nessus agent on those instances. This would mitigate the brittle part of this intended solution.```

to `make AWS automatically tag` resources you can use the [AWS tag editor](https://docs.aws.amazon.com/tag-editor/latest/userguide/tagging.html), a script using the AWS CLI, or a script using the SDK.  Those automation options can be run with a lambda or some type of cron job.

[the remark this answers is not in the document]
brucehoff: >  make AWS automatically tag every EC2 instance with a special tag (i.e. nessus-monitoring) then have this solution install the nessus agent on those instances.

Yes, that is exactly my intention:  The [Synapse Stack Builder](https://github.com/Sage-Bionetworks/Synapse-Stack-Builder) will be changed to add the tag to all instances that we wish to monitor.  Since the change will be in that repo' you will not see it in this PR.

zaro0508 (approved):

PR #1227, opened 2024-09-27
IT-3924: Create dev account for Synapse LLM

xschildw:
This PR creates an account to host the Synapse LLM development environment.

### DIFF
--- a/org-formation/organization.yaml
+++ b/org-formation/organization.yaml
@@ -619,6 +619,19 @@ Organization:
         budget-alarm-threshold: 2000
         budget-alarm-threshold-email-recipient: aws-openchallenges-prod@sagebase.org
 
+  SynapseLlmDevAccount:
+    Type: OC::ORG::Account
+    Properties:
+      AccountName: org-sagebase-synapsellm-dev
+      RootEmail: aws-synapsellm-dev@sagebase.org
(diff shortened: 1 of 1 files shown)


org-formation/organization.yaml
```diff
@@ -619,6 +619,19 @@ Organization:
         budget-alarm-threshold: 2000
         budget-alarm-threshold-email-recipient: aws-openchallenges-prod@sagebase.org
 
+  SynapseLlmDevAccount:
```
zaro0508, line 622: what does `Llm` stand for?  maybe a comment to clarify?

zaro0508, line 622: how about adding it to one of the OUs?  SynapseProdOU or ItProdOU perhaphs?

zaro0508 (changes requested):

org-formation/organization.yaml
xschildw: Done

xschildw: Added to SynapseOU. This is for SCPs, correct?

zaro0508: yes that is correct.

zaro0508 (approved):

PR #1231, opened 2024-09-30
IT-3924: Add prod account

xschildw:
This PR creates the AWS account to host the Synapse LLM infrastructure.

Note: dev account to be deleted per discussion.

### DIFF
--- a/org-formation/organization.yaml
+++ b/org-formation/organization.yaml
@@ -634,6 +634,20 @@ Organization:
         budget-alarm-threshold: 2000
         budget-alarm-threshold-email-recipient: aws-synapsellm-dev@sagebase.org
 
+# Synapse Large Language Model prod account: hosts development Bedrock-related resources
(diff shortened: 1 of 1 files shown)


org-formation/organization.yaml
```diff
@@ -634,6 +634,20 @@ Organization:
         budget-alarm-threshold: 2000
         budget-alarm-threshold-email-recipient: aws-synapsellm-dev@sagebase.org
 
+# Synapse Large Language Model prod account: hosts development Bedrock-related resources
+  SynapseLlmProdAccount:
```
zaro0508, line 638: how about add to one of the OUs above?

zaro0508 (changes requested):

org-formation/organization.yaml
xschildw: Oops...

zaro0508 (approved):

PR #1233, opened 2024-09-30
IT-3294: Add role

xschildw:
Requires #1231 

This PR creates a role to give full access to BedRock and associates it to the developer group in the SynapseLLM prod account.

### DIFF
--- a/org-formation/700-aws-sso/_tasks.yaml
+++ b/org-formation/700-aws-sso/_tasks.yaml
@@ -626,21 +626,10 @@ SSoLLmDeveloper:
   Parameters:
     instanceArn: !Ref instanceArn
     principalId: !Ref SynapseLLmProdDeveloperGroup
-    permissionSetName: 'LlmDeveloper'
+    permissionSetName: 'llm-developer'
     managedPolicies:
       - 'arn:aws:iam::aws:policy/AmazonBedrockFullAccess'
     sessionDuration: 'PT12H'
-    inlinePolicy: >-
-      {
-          "Version": "2012-10-17",
-          "Statement": [
-              {
-                  "Effect": "Deny",
-                  "Action": "sts:AssumeRole",
-                  "Resource": "*"
-              }
-          ]
-      }
 
 # Role for a user that can only access AWS Athena in the Synapse Dev account
 SsoSynapseDWDevAthenaUser:
--- a/org-formation/700-aws-sso/_tasks.yaml
+++ b/org-formation/700-aws-sso/_tasks.yaml
@@ -1009,23 +1009,6 @@ SsoSynapseLlmProdAdmin:
     principalId: !Ref SynapseLlmProdAdminGroup
     permissionSetArn: !CopyValue [ !Sub '${resourcePrefix}-${appName}-admin-permission-set-arn' ]
 
-SsoSynapseLlmProdDeveloper:
-  Type: update-stacks
-  DependsOn: ### change SsoAdministrator
-  Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.7.7/templates/SSO/aws-sso.njk
-  StackName: !Sub '${resourcePrefix}-${appName}-synapsellm-prod-developer'
-  StackDescription: 'SSO: developer role used by SynapseLlm prod developer group'
-  DefaultOrganizationBindingRegion: !Ref primaryRegion
-  DefaultOrganizationBinding:
-    IncludeMasterAccount: true
-  OrganizationBindings:
-    TargetBinding:
-      Account: !Ref SynapseLlmProdAccount
-  Parameters:
-    instanceArn: !Ref instanceArn
-    principalId: !Ref SynapseLlmProdDeveloperGroup
-    permissionSetArn: !CopyValue [ !Sub '${resourcePrefix}-${appName}-admin-permission-set-arn' ]
-
 SsoImageCentralAmiLibrarian:
   Type: update-stacks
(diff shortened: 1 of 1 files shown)


org-formation/700-aws-sso/_tasks.yaml
```diff
@@ -601,6 +610,28 @@ SsoApplicationManager:
...
+  OrganizationBindings:
+    TargetBinding:
+      Account: !Ref SynapseLlmProdAccount
+  Parameters:
+    instanceArn: !Ref instanceArn
+    principalId: !Ref SynapseLLmProdDeveloperGroup
+    permissionSetName: 'llm-developer'
+    managedPolicies:
+      - 'arn:aws:iam::aws:policy/AmazonBedrockFullAccess'
+      - 'arn:aws:iam::aws:policy/AWSCloudFormationFullAccess'
```
zaro0508, line 632: why do developers need cloudformation access?

org-formation/700-aws-sso/_tasks.yaml
```diff
@@ -601,6 +610,28 @@ SsoApplicationManager:
         ]
       }
 
+# Role for a user that can access BedRock in the Synapse LLM prod account
+SSoLLmDeveloper:
```
zaro0508, line 614: is this specific to synapse?  if so maybe it should be `SynapseSsoLlmDeveloper`?

org-formation/700-aws-sso/_tasks.yaml
```diff
@@ -601,6 +610,28 @@ SsoApplicationManager:
...
+  DefaultOrganizationBinding:
+    IncludeMasterAccount: false
+  OrganizationBindings:
+    TargetBinding:
+      Account: !Ref SynapseLlmProdAccount
+  Parameters:
+    instanceArn: !Ref instanceArn
+    principalId: !Ref SynapseLLmProdDeveloperGroup
+    permissionSetName: 'llm-developer'
+    managedPolicies:
```
zaro0508, line 630: At a minimum you will need to give developers some type of general READ access to the account otherwise they won't be able to view anything.  Basically you'll need to duplicate some of the permissions that are given to [genernal SsoDeveloper](https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/org-formation/700-aws-sso/_tasks.yaml#L394-410).  You'll definately want to add the `Deny` policy.

zaro0508 (changes requested):

org-formation/700-aws-sso/_tasks.yaml
xschildw: They setup/debug using the UI then eventually move to Cloudformation for production resources.

org-formation/700-aws-sso/_tasks.yaml
xschildw: Will check.

org-formation/700-aws-sso/_tasks.yaml
xschildw: Will add deny from SSoDeveloper... Is it not the case that we don't want llm devs to create any resources beyond Bedrock? I think they are only supposed to tinker with the prompts and manage Bedrock agents.

org-formation/700-aws-sso/_tasks.yaml
```diff
@@ -601,6 +610,39 @@ SsoApplicationManager:
...
+  DefaultOrganizationBindingRegion: !Ref primaryRegion
+  DefaultOrganizationBinding:
+    IncludeMasterAccount: false
+  OrganizationBindings:
+    TargetBinding:
+      Account: !Ref SynapseLlmProdAccount
+  Parameters:
+    instanceArn: !Ref instanceArn
+    principalId: !Ref SynapseLlmProdDeveloperGroup
+    permissionSetName: 'llm-developer'
```
zaro0508, line 629: this does not match the below, `permissionSetArn: !CopyValue [ !Sub '${resourcePrefix}-${appName}-synapsellmdeveloper-permission-set-arn' ]`

org-formation/700-aws-sso/_tasks.yaml
zaro0508: ok, we can try with this however i don't think it's enough for an LLM developer.

zaro0508 (changes requested):

org-formation/700-aws-sso/_tasks.yaml
xschildw: My understanding is that at https://github.com/Sage-Bionetworks-IT/organizations-infra/pull/1233/files#diff-c89df69681d7714e12cc8907e3d7e5b10a9bee477bbff86cd269b81ad0c7a1b5R1038, we reference the export name of aws-sso.njk, which is using the stack name. Did I misunderstand?
`Outputs:
  PermissionSetArn:
    Value: !If [ includePermissionSet, !GetAtt PermissionSet.PermissionSetArn, !Ref permissionSetArn ]
    Export:
      Name: !Sub ${AWS::StackName}-permission-set-arn
`

zaro0508: sorry, you are correct.

zaro0508 (approved):

org-formation/700-aws-sso/_tasks.yaml
zaro0508: ok, we can try with this scoped policy but i don't think it will be enough.

org-formation/700-aws-sso/_tasks.yaml
xschildw: Not at all, your answer to my question on Slack prompted me to look ;-)

PR #1237, opened 2024-10-02
PORTALS-3261: Updated distribution IDs

xschildw:
This PR updates the permitted distribution IDS (generated from the list of distributions rather than the IDs in the Jenkins jobs).


xschildw:
We were missing one in the earlier list, hopefully this is not worse. If we still have problem may just put '*' and be done with it...

zaro0508 (approved):

PR #1254, opened 2024-10-11
PLFM-8102: Use Admin Access

xschildw:
This PR switches from specifying the permissions to using a role created by cdk bootstrap that allows assuming a set of policies covering these permissions (also created by cdk bootstrap).

### DIFF
--- a/org-formation/650-identity-providers/_tasks.yaml
+++ b/org-formation/650-identity-providers/_tasks.yaml
@@ -836,8 +836,18 @@ GithubOidcNbConvertDeploy:
     ProviderArn: !CopyValue [ !Sub '${resourcePrefix}-${appName}-ProviderArn' ]
     ProviderRoleName: !Sub ${resourcePrefix}-${appName}-nbconvert-deploy
     MaxSessionDuration: 7200
-    ManagedPolicyArns:
-      - "arn:aws:iam::${AWS::AccountId}:policy/cdk-assume-role-policy"
+    PolicyDocument: !Sub |
+      {
+          "Version": "2012-10-17",
+          "Statement": [
+              {
+                  "Sid": "AssumeRoleStatement",
+                  "Effect": "Allow",
+                  "Action": "sts:AssumeRole",
+                  "Resource": "arn:aws:iam::${AWS::AccountId}:role/cdk-*-role-*-us-east-1"
+              }
+          ]
+      }
   TemplatingContext:
     GitHubOrg: "Sage-Bionetworks"
     Repositories:
--- a/org-formation/650-identity-providers/_tasks.yaml
+++ b/org-formation/650-identity-providers/_tasks.yaml
@@ -166,6 +166,30 @@ GithubOidcSageBionetworksSynapseDockerRegistry:
       - !Ref SynapseProdAccount
     Region: us-east-1
 
+GithubOidcSageBionetworksSynapse:
+  Type: update-stacks
+  DependsOn: GithubOidcSageBionetworks
+  Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.7.6/templates/IAM/github-oidc-provider.j2
+  StackName: !Sub ${resourcePrefix}-${appName}-sage-bionetworks-synapse
+  Parameters:
+    ProviderArn: !CopyValue [ !Sub '${resourcePrefix}-${appName}-ProviderArn' ]
+    ProviderRoleName: !Sub ${resourcePrefix}-${appName}-sage-bionetworks-synapse
+    ManagedPolicyArns:
+      - "arn:aws:iam::aws:policy/AdministratorAccess"
+      - "arn:aws:iam::aws:policy/AWSKeyManagementServicePowerUser"
+  TemplatingContext:
+    GitHubOrg: "Sage-Bionetworks"
+    Repositories:
+      - name: "synapse-docker-registry"
+        branches: ["*"]
+      - name: "nbconvert-webapp"
+        branches: ["master", "develop"]
+  DefaultOrganizationBinding:
+    Account:
+      - !Ref SynapseDevAccount
+      - !Ref SynapseProdAccount
+    Region: us-east-1
+
 GithubOidcSageBionetworksGenieBPCInfra:
   Type: update-stacks
   DependsOn: GithubOidcSageBionetworks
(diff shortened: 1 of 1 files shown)


org-formation/650-identity-providers/_tasks.yaml
```diff
@@ -809,11 +809,7 @@ GithubOidcNbConvertDeploy:
     ProviderRoleName: !Sub ${resourcePrefix}-${appName}-nbconvert-deploy
     MaxSessionDuration: 7200
     ManagedPolicyArns:
-      - "arn:aws:iam::aws:policy/AmazonEC2ContainerRegistryFullAccess"
-      - "arn:aws:iam::aws:policy/AWSLambda_FullAccess"
-      - "arn:aws:iam::aws:policy/CloudWatchLogsFullAccess"
-      - "arn:aws:iam::aws:policy/IAMFullAccess"
-      - "arn:aws:iam::aws:policy/AWSCloudFormationFullAccess"
+      - "arn:aws:iam::449435941126:policy/cdk-assume-role-policy"
```
xschildw, line 812: Not sure if this is going to cause an error when used in SynapseProd account (where this particular policy does not exist).

ConsoleCatzirl: This is a cross-account policy reference, the policy only needs to exist in 449435941126. as long as all of our other accounts have been granted access to this policy, it should be fine

ConsoleCatzirl (approved):

org-formation/650-identity-providers/_tasks.yaml
zaro0508: First, I'm doubtful that this will work.  Second, where is this policy created?  Third, It's probably not a good practice to depend on a policy from a different account.  I recommend not taking this approach, instead you should probably create this policy in every account you plan to deploy this  stack to then you can just set ManagedPolicyArns to `arn:aws:iam::aws:policy/cdk-assume-role-policy`

zaro0508 (changes requested):

org-formation/650-identity-providers/_tasks.yaml
xschildw: This was a temporary test (i.e. not intended for prod use) but @zaro0508 's right, the cdk-assume-role-policy should be created by the 'cdk bootstrap' command in every account, I think it should be referred to as ' arn:aws:iam::${AWS::AccountId}:policy/cdk-assume-role-policy ' since it's user-managed.
Note re: if it works, I created an oidc role manually in the dev account using the policy and was able to use it to update a stack from a GH workflow.

xschildw:
Turns out that 'cdk-assume-role-policy' was not created by the CDK so I just use an equivalent inline policy.

org-formation/650-identity-providers/_tasks.yaml
```diff
@@ -836,12 +836,18 @@ GithubOidcNbConvertDeploy:
...
-      - "arn:aws:iam::aws:policy/AWSCloudFormationFullAccess"
+    PolicyDocument: !Sub |
+      {
+          "Version": "2012-10-17",
+          "Statement": [
+              {
+                  "Sid": "AssumeRoleStatement",
+                  "Effect": "Allow",
+                  "Action": "sts:AssumeRole",
+                  "Resource": "arn:aws:iam::${AWS::AccountId}:role/cdk-*-role-*-us-east-1"
```
zaro0508, line 847: to be more precise, how about `arn:aws:iam::${AWS::AccountId}:role/cdk-*-cfn-exec-role-${AWS::AccountId}-us-east-1`?

zaro0508: The idea here doesn't really make sense to me because if you look at the policy for the cdk cloudformation execution role you'll see that it is given `AdministratorAccess` therefore it would be equivalent to giving this OIDC admin access directly.  Why give this role the runaround of assuming another role with admin access when you can give this role admin access directly?

zaro0508 (changes requested):

org-formation/650-identity-providers/_tasks.yaml
xschildw: Yeah, I was reading about it afterwards... As I understand the way to solve this is to create a custom policy then redo the bootstrap such that the default Admin perms are changed to more constrained permissions (I'm a bit lost as that set of permissions would have to cover all the cases in a given account...). Maybe you're right and we may as well be explicit. I'll change it.

xschildw: Changed to explicitly use AdministratorAccess.

org-formation/650-identity-providers/_tasks.yaml
```diff
@@ -882,11 +882,8 @@ GithubOidcNbConvertDeploy:
     ProviderRoleName: !Sub ${resourcePrefix}-${appName}-nbconvert-deploy
     MaxSessionDuration: 7200
     ManagedPolicyArns:
-      - "arn:aws:iam::aws:policy/AmazonEC2ContainerRegistryFullAccess"
-      - "arn:aws:iam::aws:policy/AWSLambda_FullAccess"
-      - "arn:aws:iam::aws:policy/CloudWatchLogsFullAccess"
-      - "arn:aws:iam::aws:policy/IAMFullAccess"
-      - "arn:aws:iam::aws:policy/AWSCloudFormationFullAccess"
+      - "arn:aws:iam::aws:policy/AdministratorAccess"
```
zaro0508, line 885: This resource is now basically the same as `GithubOidcSageBionetworksSynapseDockerRegistry`.  Would it make sense to consolidate these to a single generic `GithubOidcSageBionetworksSynapse` resource that gives access to multiple repos?  The benefit is that when new repos pop up to deploy lambda or CDK they can just use the generic resource.

xschildw: Good idea! Renamed, added registry repo.

org-formation/650-identity-providers/_tasks.yaml
```diff
@@ -166,6 +166,30 @@ GithubOidcSageBionetworksSynapseDockerRegistry:
       - !Ref SynapseProdAccount
     Region: us-east-1
 
+GithubOidcSageBionetworksSynapse:
```
zaro0508, line 169: do you plan to also replace the role used by GH actions in `synapse-docker-registry` repo with the one generated from this new `GithubOidcSageBionetworksSynapse` resource?  And then of course also remove `GithubOidcSageBionetworksSynapseDockerRegistry`?

zaro0508 (approved):
Generally looks good.  Just remember to update the synapse-docker-registry with the new role.

org-formation/650-identity-providers/_tasks.yaml
xschildw: Correct.

PR #1268, opened 2024-10-25
[IT-3931] Create a hosted zone for dpe.org

zaro0508:
We have aquired the `sagedpe.org` domain in the org-sagebase-sageit account. Now we need to create a hosted zone for that domain.

### DIFF
--- a/org-formation/100-shared-dns/_tasks.yaml
+++ b/org-formation/100-shared-dns/_tasks.yaml
@@ -114,3 +114,15 @@ BridgeDigitalHealhHostedZone:
     Account: !Ref SageITAccount
   Parameters:
     DnsDomainName: "bridgedigital.health"
+
+# Hosted zone for DPE (data processing engineering) applications, e.g. 'myapp.sagedpe.org'
+SageDpeZone:
+  Type: update-stacks
+  Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.6.8/templates/R53-hostedzone.yaml
+  StackName: !Sub '${resourcePrefix}-sage-dpe-zone'
+  StackDescription: Create a shared hosted zone for application in dpe.org
+  DefaultOrganizationBindingRegion: !Ref primaryRegion
+  DefaultOrganizationBinding:
+    Account: !Ref SageITAccount
+  Parameters:
+    DnsDomainName: "dpe.sagedpe.org"


BryanFauble (approved):
LGTM!

org-formation/100-shared-dns/_tasks.yaml
```diff
@@ -114,3 +114,15 @@ BridgeDigitalHealhHostedZone:
...
+
+# Hosted zone for DPE (data processing engineering) applications, e.g. 'myapp.sagedpe.org'
+SageDpeZone:
+  Type: update-stacks
+  Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.6.8/templates/R53-hostedzone.yaml
+  StackName: !Sub '${resourcePrefix}-sage-dpe-zone'
+  StackDescription: Create a shared hosted zone for application in dpe.org
+  DefaultOrganizationBindingRegion: !Ref primaryRegion
+  DefaultOrganizationBinding:
+    Account: !Ref SageITAccount
```
xschildw, line 126: Should the subdomain not be created in the account where the assets are going to be deployed?

zaro0508: Since we requested the `sagedpe.org` in the org-sagebase-sageit account then i think we need a hosted zone for it in the same account as well.  The hosted zone will allow us to verify TLS certificate ownership (via DNS verification) and setup redirect to apps running in other accounts.  That seems like the pattern that has been establish to me or am I missing something @xschildw ?

xschildw: Sorry for the delay...
The domain 'sagedpe.org' has an associated hosted zone. I think it would work as-is (including certificate validation for subdomains), with the caveat that you'll need cross-account access if you want to create from another account (where your resources live), or use a separate script running in sageit to create the records.
This PR creates a new zone 'dpe.sagedpe.org', that would work to create records for that domain and its subdomains. NS records would need to be created in 'sagedpe.org' to delegate DNS to this zone for the 'dpe' subdomain. The zone can be created in any account (typically we put it where the resources that need DNS records are created, so your DNS is colocated with your resources). I think same account is required if you use alias records. Another advantage is that it's cleaner if you have a lot of subdomains/records. 
If 'dpe.sagedpe.org' is just an endpoint maybe we don't need an extra zone at all, if there are several resources/records in 'dpe.sagedpe.org' I'd create a zone in the account where the resources are.

zaro0508: OMG, i think AWS automatically creates a hosted zone upon a successful request of a new domain (sagedpe.org).  Darn, sorry I didn't notice that the hosted zone was created.   Thanks @xschildw.

zaro0508:
hosted zone already setup, no need to create another one.

PR #1269, opened 2024-10-29
[IT-3950] Updated Defender for Cloud template

frsfx:
PR Checklist:
- [ ] Updated the Defender for Cloud template with the latest generated template for a single AWS account.  
- [ ] Setup pre-commit and run the validators (info in README.md)
- [ ]


zaro0508 (changes requested):
can you please explain what you changed in MS defender to make it generate such a  different version of the template from the previous one?

frsfx:
> can you please explain why you are updating this template? Also what did you changed in MS defender to make it generate such a different version of the template from the previous one?

I'm generating a new template that only supports a single AWS account.  The original template was to support multiple AWS accounts.

zaro0508 (approved):

PR #1273, opened 2024-11-05
IT-3988: Add permission to create object in S3

xschildw:
This PR adds a in-line policy to LLMDeveloper role that gives the permission to create objects in S3 (upload templates).


brucehoff (approved):
I'm approving this but I wonder if you also need permission to read and delete objects.

xschildw:
Good point, the error was specifically to upload the file so hoping it'll be enough... Let's try!

PR #1277, opened 2024-11-07
IT-3988: Add role to invoke model and move inline sso policy to managed policy

xschildw:
This PR creates a role that gives Bedrock permission to execute model


org-formation/700-aws-sso/_tasks.yaml
```diff
@@ -641,18 +641,6 @@ SsoLlmDeveloper:
     managedPolicies:
       - 'arn:aws:iam::aws:policy/AmazonBedrockFullAccess'
       - 'arn:aws:iam::aws:policy/AWSCloudFormationFullAccess'
-#   https://stackoverflow.com/questions/58125181/cloud-formation-cant-upload-template-file
```
xschildw, line 644: Moving to a managed policy with the hope we can tweak it manually until we get what we need. Inline, we get an error because it's under AWS control.

org-formation/600-access/_tasks.yaml
```diff
@@ -388,3 +388,31 @@ SynapseAthenaUserAccessPolicy:
         ]
       }
     PolicyName: SynapseAthenaUserAccessPolicy
+
+SynapseLlmBedrockAgentPolicy:
```
xschildw, line 392: @zaro0508 , wondering if I should have the policy here or if I should move it as a resource alongside the bedrock agent role below (org-formation/745-bedrock-agent-role/bedrock-agent-role.yaml).

org-formation/745-bedrock-agent-role/bedrock-agent-role.yaml
```diff
@@ -0,0 +1,35 @@
+AWSTemplateFormatVersion: '2010-09-09'
+Description: Enables executing a Bedrock model
+
+Resources:
+  bedrockAgentRole:
+    Type: AWS::IAM::Role
```
xschildw, line 6: This is the role that the Bedrock service assumes, it gives it the permission to execute a Bedrock model.

brucehoff (approved):

org-formation/600-access/_tasks.yaml
zaro0508: > @zaro0508 , wondering if I should have the policy here or if I should move it as a resource alongside the bedrock agent role below.

You'll notice that existing policy definitions created in here are meant to be applied to all accounts (`Account: '*'`) because some AWS features require us to apply the same policy.  If you are intending to apply this to one account then I would say that it's more appropriate to  setup this policy with the resource that needs it.

org-formation/745-bedrock-agent-role/bedrock-agent-role.yaml
zaro0508: instead of creating a separate folder, i think you can just put this in org-formation/600-access/_tasks.yaml

zaro0508 (changes requested):
t

zaro0508 (changes requested):
From reading the issue it looks like the goal is to update the sso `llm developer` role with a few missing permissions however I don't understand your implementation to accomplish that goal.  I might be missing something but I don't see an explanation as to why you need to make all these changes  to accomplish your goal.  Why not just keep the format as-is and update the inline policy in org-formation/700-aws-sso/_tasks.yaml?

xschildw:
We first have to determine which permissions we need, ideally out of the PR cycle (i.e. only have one PR with everything we need). With an inline policy, we are not able to add permissions manually in the console (the policy in under AWS control). So we're trying to setup a managed policy, and in next PR we'll reference it in the role.

zaro0508:
> With an inline policy, we are not able to add permissions manually in the console (the policy in under AWS control). So we're trying to setup a managed policy, and in next PR we'll reference it in the role.

ahh, i didn't realize that there is no way to update an inline policy in AWS console.  For testing purposes, I typically use the AWS console to manually create a new inline policy with additional access and apply that to a role.  That lets me craft a least privilege policy before committing it to cloudformation.

Anyways, now that I understand I don't have a problem with your approach either it just makes it more difficult to follow in this org-formation/cloudformation setup.

org-formation/600-access/_tasks.yaml
```diff
@@ -388,3 +388,39 @@ SynapseAthenaUserAccessPolicy:
...
+          {
+            "Effect": "Allow",
+            "Action": "iam:PassRole",
+            "Resource": "*"
+          }
+        ]
+      }
+    PolicyName: SynapseLlmDeveloperPolicy
+
+BedrockAgentRole:
```
zaro0508, line 420: why is this role necessary when the goal is to only fix the developer role?

org-formation/700-aws-sso/_tasks.yaml
```diff
@@ -641,18 +641,6 @@ SsoLlmDeveloper:
...
-    inlinePolicy: >-
```
zaro0508, line 645: Instead of removing this inline policy in this PR i would suggest you do this in a follow on PR that uses `SynapseLlmDeveloperPolicy`.  Make the follow on PR `depends on` this PR.  It will help provide reviewers with a complete picture of this change.

zaro0508 (changes requested):

org-formation/700-aws-sso/_tasks.yaml
xschildw: OK

org-formation/600-access/_tasks.yaml
xschildw: This is the policy for the user (sso-llmdeveloper) to be able to work in the console/cloudformation to create an agent.

org-formation/600-access/bedrock-agent-role.yaml
```diff
@@ -0,0 +1,35 @@
+AWSTemplateFormatVersion: '2010-09-09'
```
xschildw, line 1: This is the role used by the agent when it's running. It has the right to execute a model.

org-formation/300-account-defaults/bedrock-agent-role.yaml
```diff
@@ -0,0 +1,36 @@
...
+                Action: "bedrock:InvokeModel"
+                Resource:
+                  - !Sub "arn:aws:bedrock:${AWS::Region}::foundation-model/*"
+
+Outputs:
+  BedrockAgentRoleArn:
+    Description: The ARN of the Bedrock Agent Role
+    Value: !GetAtt bedrockAgentRole.Arn
+    Export:
+      Name: BedrockAgentRoleArn
```
zaro0508, line 36: outputs need to be unique therefore it would be better if you add stack name here, i.e. `!Sub '${AWS::StackName}-BedrockAgentRoleArn'`

zaro0508 (changes requested):
just a small fix up and should be good to go.

org-formation/300-account-defaults/bedrock-agent-role.yaml
xschildw: Fixed

zaro0508 (approved):

PR #1279, opened 2024-11-12
[IT-3984] Schematic on new infra deployment

zaro0508:
There's a new deployment for Schematic app at Sage-Bionetwork-IT/schematic-infra-v2. setup github action CI to deploy from there.

depends on https://github.com/Sage-Bionetworks-IT/schematic-infra-v2/pull/4


org-formation/650-identity-providers/_tasks.yaml
```diff
@@ -112,10 +112,10 @@ GithubOidcSageBionetworksSchematicInfra:
       - "arn:aws:iam::aws:policy/AdministratorAccess"
       - "arn:aws:iam::aws:policy/AWSKeyManagementServicePowerUser"
   TemplatingContext:
-    GitHubOrg: "Sage-Bionetworks"
+    GitHubOrg: "Sage-Bionetworks-IT"
```
ConsoleCatzirl, line 115: I think it would be better to add support for schematic-infra-v2 in one PR, and remove schematic-infra in another PR after the migration just in case you need to run another deployment for schematic-infra during the migration

ConsoleCatzirl (changes requested):

org-formation/650-identity-providers/_tasks.yaml
zaro0508: done

ConsoleCatzirl (approved):

PR #1282, opened 2024-11-13
IT-3988: Replace inline policy by managed policy

xschildw:
This PR uses the managed policy created in #1277 instead of the inline policy.


org-formation/700-aws-sso/_tasks.yaml
```diff
@@ -642,17 +642,7 @@ SsoLlmDeveloper:
...
-          "Version": "2012-10-17",
-          "Statement": [
-              {
-                  "Effect": "Allow",
-                  "Action": "s3:*",
-                  "Resource": "arn:aws:s3:::cf-template*"
-              }
-          ]
-      }
+      - !Sub 'arn:aws:iam::${AWS::AccountId}:policy/SynapseLlmDeveloperPolicy'
```
zaro0508, line 645: What's the deal with the comment just above?  does it pertain to this managed policy?

zaro0508 (approved):

org-formation/700-aws-sso/_tasks.yaml
xschildw: Moved comment to definition of the managed policy

PR #1292, opened 2024-11-19
IT-4016/IT-4017: Give PowerUser permissions and scale down later

xschildw:
As LLM developers are using features in the console, we'll keep getting permission errors. After discussion with John, we decided to up to PowerUserAccess, then track and scale down. We may still have to adjust IAM permissions if/as needed.

### DIFF
--- a/org-formation/700-aws-sso/_tasks.yaml
+++ b/org-formation/700-aws-sso/_tasks.yaml
@@ -644,6 +644,17 @@ SsoLlmDeveloper:
     permissionSetName: 'LlmDeveloper'
     managedPolicies:
       - 'arn:aws:iam::aws:policy/PowerUserAccess'
+    inlinePolicy: >-
+      {
+          "Version": "2012-10-17",
+          "Statement": [
+              {
+                  "Effect": "Deny",
+                  "Action": "sts:AssumeRole",
+                  "Resource": "*"
+              }
+          ]
+      }
     sessionDuration: 'PT12H'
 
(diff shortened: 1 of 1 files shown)


org-formation/700-aws-sso/_tasks.yaml
```diff
@@ -643,8 +643,7 @@ SsoLlmDeveloper:
     principalId: !Ref llmDeveloperGroup
     permissionSetName: 'LlmDeveloper'
     managedPolicies:
-      - 'arn:aws:iam::aws:policy/AmazonBedrockFullAccess'
-      - 'arn:aws:iam::aws:policy/AWSCloudFormationFullAccess'
+      - 'arn:aws:iam::aws:policy/PowerUserAccess'
```
zaro0508, line 646: Note, this is essentially the same as [developer role](https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/org-formation/700-aws-sso/_tasks.yaml#L393-L432) access, so why not just use the developer role?

Also PowerUser should be paired with a deny assume role policy to prevent unauthorized role escalation.

zaro0508 (changes requested):

org-formation/700-aws-sso/_tasks.yaml
xschildw: The goal is to track the permissions actually used and move to more a more restricted set of permissions. Added policy to prevent assume-role (same as ssoDeveloper).

zaro0508 (approved):

PR #1296, opened 2024-11-21
Tear down DNS redirect for schematic-dev refactor

BryanFauble:
Since moving to the schematic multi-container deployment this stack is no longer needed and is being torn down


zaro0508 (commented):
Don't we want to remove `SchematicStagingAppDnsForward` and `SchematicProdAppDnsForward` as well?

BryanFauble:
> Don't we want to remove `SchematicStagingAppDnsForward` and `SchematicProdAppDnsForward` as well?

Not at this point.

`SchematicStagingAppDnsForward` is pointing to the correct location and should not be changed.

`SchematicProdAppDnsForward` will be updated once we get the green light to release changes to production using the new stack. Validation is still taking place on the staging stack.

zaro0508 (approved):

PR #1299, opened 2024-11-22
[IT-3995] Setup DNS redirect for ECS Agora deployment

zaro0508:
Setup redirects for Agora deployment in ECS.  We setup temporary names for now. It can be switch over to the vanity plates when we are to use in production.


xschildw (approved):
LGTM

tschaffter (approved):

PR #1310, opened 2025-01-02
[IT-2431] Require secure access for `ConfigAuditBucket`

ConsoleCatzirl:
Secure the `ConfigAuditBucket` by denying any S3 API calls for the bucket that do not use TLS/HTTPS.

This addresses a Security Hub finding from
`cis-aws-foundations-benchmark/v/1.4.0/2.1.2`.

Ref: https://repost.aws/knowledge-center/s3-bucket-policy-for-config-rule

### DIFF
--- a/org-formation/080-aws-config-inventory/config.yaml
+++ b/org-formation/080-aws-config-inventory/config.yaml
@@ -67,6 +67,16 @@ Resources:
             Condition:
               StringEquals:
                 's3:x-amz-acl': 'bucket-owner-full-control'
+          - Sid: AWSConfigBucketDenyInsecure
+            Effect: Deny
+            Principal: '*'
+            Action: 's3:*'
+            Resource:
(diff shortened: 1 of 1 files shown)


brucehoff (approved):

brucehoff:
@ConsoleCatzirl What do you think about adding this restriction at the "organization" level, as a [security control policy](https://docs.aws.amazon.com/organizations/latest/userguide/orgs_manage_policies_scps.html)?  I ask because we have 24 Jira issues for the `cis-aws-foundations-benchmark/v/1.4.0/2.1.2` control, across the accounts in our AWS organization, and I'm wondering if we can address all with a single policy.

ConsoleCatzirl:
@brucehoff I think that's a great idea. I'm closing this PR in favor of Sage-Bionetworks-IT/organizations-infra#1312

ConsoleCatzirl:
Unfortunately we can't do this with an SCP, reopening

PR #1312, opened 2025-01-07
[IT-2431] Deny insecure access to S3 with an SCP

ConsoleCatzirl:
Create a service control policy to deny any S3 api calls that don't use a secure transport layer, such as HTTP, effectively forcing HTTPS.

This should address several security findings including IT-2431

### DIFF
--- a/org-formation/_scp.yaml
+++ b/org-formation/_scp.yaml
@@ -175,3 +175,20 @@
             Action:
               - organizations:LeaveOrganization
             Resource: "*"
+
+  RequireSecureS3:
+    Type: OC::ORG::ServiceControlPolicy
+    Properties:
+      PolicyName: RequireSecureS3
+      Description: Restrict users from accessing S3 over HTTP
+      PolicyDocument:
+        Version: '2012-10-17'
+        Statement:
+          - Sid: DenyInsecureS3
+            Effect: Deny
+            Action:
+              - 's3:*'
+            Resource: '*'
+            Condition:
+              Bool:
+                'aws:SecureTransport': false


org-formation/_scp.yaml
```diff
@@ -175,3 +175,20 @@
...
+        Version: '2012-10-17'
+        Statement:
+          - Sid: DenyInsecureS3
+            Effect: Deny
+            Action:
+              - 's3:*'
+            Resource: '*'
+            Condition:
+              Bool:
+                'aws:SecureTransport': false
```
zaro0508, line 194: The following AWS KB seems to indicate that we would need `"Principal": "*"` in this PolicyDocument?
https://repost.aws/knowledge-center/s3-bucket-policy-for-config-rule

You should test to see what type of message you get when attempting to get files over http

zaro0508 (changes requested):

org-formation/_scp.yaml
ConsoleCatzirl: Thanks, I saw that KB article but it's describing a bucket policy and this PR is creating a service control policy, so I don't think the `Principal` is needed, I don't see it in any of the examples [here](https://docs.aws.amazon.com/organizations/latest/userguide/orgs_manage_policies_scps_examples_general.html)

I'm working on setting up a test in the sandbox organization, I'll comment my results in the ticket.

zaro0508: ohh, hmm.  i'm not sure `'aws:SecureTransport': false` is supported for SCP but i guess the easiest thing to do is try a deploy and see what happens.

brucehoff (approved):
Looks good!

org-formation/_scp.yaml
ConsoleCatzirl: Oh, hrm. I tried this out in the test org, and it created the SCP just fine but it had no effect. We'll have to go through each bucket individually and either add the policy or suppress the finding.

ConsoleCatzirl:
This SCP had no effect in the test org, closing this and reopening Sage-Bionetworks-IT/organizations-infra#1310

PR #1319, opened 2025-01-09
IT-4081: suppress cis-aws-foundations-benchmark/v/1.4.0/3.9 since we use GuardDuty to monitor flow logs

brucehoff:
suppress cis-aws-foundations-benchmark/v/1.4.0/3.9 since we use GuardDuty to monitor flow logs

### DIFF
--- a/org-formation/075-security-hub/security-hub-suppress-infra.yaml
+++ b/org-formation/075-security-hub/security-hub-suppress-infra.yaml
@@ -377,6 +377,8 @@ Resources:
               - 'arn:aws:securityhub:::ruleset/cis-aws-foundations-benchmark/v/1.2.0/rule/4.1'
               - 'arn:aws:securityhub:::ruleset/cis-aws-foundations-benchmark/v/1.2.0/rule/4.2'
               - 'cis-aws-foundations-benchmark/v/1.4.0/3.5'  # (IT-3619) "3.5 Ensure AWS Config is enabled in all regions"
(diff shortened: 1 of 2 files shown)


org-formation/075-security-hub/security-hub-suppress-infra.yaml
```diff
@@ -377,6 +377,8 @@ Resources:
               - 'arn:aws:securityhub:::ruleset/cis-aws-foundations-benchmark/v/1.2.0/rule/4.1'
               - 'arn:aws:securityhub:::ruleset/cis-aws-foundations-benchmark/v/1.2.0/rule/4.2'
               - 'cis-aws-foundations-benchmark/v/1.4.0/3.5'  # (IT-3619) "3.5 Ensure AWS Config is enabled in all regions"
+              # suppress cis-aws-foundations-benchmark/v/1.4.0/3.9 since we use GuardDuty to monitor flow logs:
+              - 'cis-aws-foundations-benchmark/v/1.4.0/3.9'
```
ConsoleCatzirl, line 381: My understanding is that guard duty is good for detection but not for investigation so I think we should enforce this finding at least for production accounts, but I'm ok with suppressing it in non-prod accounts

ConsoleCatzirl (changes requested):

org-formation/075-security-hub/security-hub-suppress-infra.yaml
zaro0508: it would probably be beneficial if you could add a reference to issue IT-2314 and provide justification for this suppression in the issue.

zaro0508 (approved):

PR #1321, opened 2025-01-10
IT-4232: Add permission to access S3

xschildw:
This PR adds permissions for a bedrock agent to access S3

### DIFF
--- a/org-formation/300-account-defaults/bedrock-agent-role.yaml
+++ b/org-formation/300-account-defaults/bedrock-agent-role.yaml
@@ -31,7 +31,9 @@ Resources:
                 Action: "lambda:InvokeFunction"
                 Resource:
                   - !Sub "arn:aws:lambda:${AWS::Region}:${AWS::AccountId}:function:*"
-
+              - Effect: Allow
+                Action: "s3:*"
+                Resource: "*"
 Outputs:
   BedrockAgentRoleArn:
     Description: The ARN of the Bedrock Agent Role


org-formation/300-account-defaults/bedrock-agent-role.yaml
```diff
@@ -31,7 +31,9 @@ Resources:
                 Action: "lambda:InvokeFunction"
                 Resource:
                   - !Sub "arn:aws:lambda:${AWS::Region}:${AWS::AccountId}:function:*"
-
+              - Effect: Allow
+                Action: "s3:*"
+                Resource: "*"
```
zaro0508, line 36: why not use just apply the [AmazonS3FullAccess policy](https://docs.aws.amazon.com/aws-managed-policy/latest/reference/AmazonS3FullAccess.html) managed policy?

zaro0508 (changes requested):

org-formation/300-account-defaults/bedrock-agent-role.yaml
xschildw: Good idea! Updated.

org-formation/300-account-defaults/bedrock-agent-role.yaml
```diff
@@ -18,6 +18,8 @@ Resources:
                 aws:SourceAccount: !Ref AWS::AccountId
               ArnLike:
                 aws:SourceArn: !Sub "arn:aws:bedrock:${AWS::Region}:${AWS::AccountId}:agent/*"
+      ManagedPolicyArns:
+        - arn:aws:iam::aws:policy/AmazonS3FullAccess
```
zaro0508, line 22: you can also replace your `bedrockAgentPolicy` with the AWS managed `AWSLambdaRole` policy..
image

zaro0508 (changes requested):

PR #1323, opened 2025-01-16
[IT-3831] GH OIDC image builder deployments

zaro0508:
Instead of piggy backing on a packer role it's probably better to create a separate role for image builder deployments.


brucehoff (approved):
lgtm

ConsoleCatzirl (approved):

PR #1334, opened 2025-02-07
[IT-4287] Automate SSM patching for all instances in all accounts

ConsoleCatzirl:
Use SSM Quick Setup to create a Patch Policy that applies to all managed nodes in the Platform organizational unit (Synapse accounts and Bridge accounts).

User Guide:
https://docs.aws.amazon.com/systems-manager/latest/userguide/quick-setup-patch-manager.html 

AWS Blog:
https://aws.amazon.com/blogs/mt/deploy-aws-systems-manager-quick-setup-programmatically-across-your-aws-organization/

Examples:
https://github.com/aws-samples/aws-management-and-governance-samples/blob/master/AWSSystemsManager/Quick-Setup-API/patch-policy-examples/patch-policy-cfn-template.yaml
https://github.com/aws-samples/aws-management-and-governance-samples/blob/master/AWSSystemsManager/Quick-Setup-API/patch-policy-examples/default-patch-baselines.txt

### DIFF
--- a/org-formation/090-systems-manager/QuickSetup-PatchPolicy-Prod.yaml
+++ b/org-formation/090-systems-manager/QuickSetup-PatchPolicy-Prod.yaml
@@ -1,0 +1,56 @@
+AWSTemplateFormatVersion: '2010-09-09'
+Description: An SSM patch policy to patch all instances in the given OUs
+Parameters:
+  TargetOrgUnits:
+    Type: String
+    Description: >-
+      Comma separated list of AWS Organizational Units to target.
+  TargetRegions:
+    Type: String
+    Description: >-
+      Comma separated list of AWS Organizational Units to target.
+Resources:
+  PatchPolicy:
+    Type: "AWS::SSMQuickSetup::ConfigurationManager"
+    Properties:
+      Name: "PatchResourceGroup"
+      Description: "Patch all SSM managed nodes in production accounts"
+      ConfigurationDefinitions:
+        Type: AWSQuickSetupType-PatchPolicy
+        Parameters:
+          PatchPolicyName: PatchPolicy
+          # Patch baselines can be found with the command:
+          #   aws ssm describe-patch-baselines |jq '.BaselineIdentities | map({ (.OperatingSystem): (.value = .BaselineId | .label = .BaselineName | .description = .BaselineDescription | .disabled=false | del(.BaselineId, .BaselineName, .OperatingSystem, .BaselineDescription, .DefaultBaseline)) }) | add'
+          #
+          # More information at:
+          #   https://aws.amazon.com/blogs/mt/deploy-aws-systems-manager-quick-setup-programmatically-across-your-aws-organization/
+          #   https://github.com/aws-samples/aws-management-and-governance-samples/blob/master/AWSSystemsManager/Quick-Setup-API/patch-policy-examples/patch-policy-cfn-template.yaml
+          #   https://github.com/aws-samples/aws-management-and-governance-samples/blob/master/AWSSystemsManager/Quick-Setup-API/patch-policy-examples/default-patch-baselines.txt
+          SelectedPatchBaselines: >-
+            {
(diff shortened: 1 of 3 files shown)


org-formation/090-systems-manager/QuickSetup-PatchPolicy-Prod.yaml
```diff
@@ -0,0 +1,56 @@
...
+Description: An SSM patch policy to patch all instances in the given OUs
+Parameters:
+  TargetOrgUnits:
+    Type: String
+    Description: >-
+      Comma separated list of AWS Organizational Units to target.
+  TargetRegions:
+    Type: String
+    Description: >-
+      Comma separated list of AWS Organizational Units to target.
```
zaro0508, line 11: it's probably better to change these to [CommaDelimitedList](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/parameters-section-structure.html#parameters-section-structure-properties)  types

org-formation/090-systems-manager/_tasks.yaml
```diff
@@ -61,6 +61,18 @@ PatchMembers:
   Parameters:
     ManagementAccountNumber: !Ref accountId
 
+PatchPolicy:
+  Type: update-stacks
+  Template: QuickSetup-PatchPolicy-Prod.yaml
+  StackName: !Sub '${resourcePrefix}-${appName}-patchpolicy'
+  StackDescription: Setup Systems manager patch policy for prod OUs
```
zaro0508, line 68: don't we want to patch all instances in the entire organization (all accounts)?

org-formation/090-systems-manager/QuickSetup-PatchPolicy-Prod.yaml
```diff
@@ -0,0 +1,56 @@
+AWSTemplateFormatVersion: '2010-09-09'
```
zaro0508, line 1: what is the significance of `Prod` in the filename?  do you ever plan to have a `dev` version?

org-formation/090-systems-manager/_tasks.yaml
```diff
@@ -61,6 +61,18 @@ PatchMembers:
...
+  DefaultOrganizationBindingRegion: !Ref primaryRegion
+  DefaultOrganizationBinding:
+    Account: !Ref accountId
+  Parameters:
+    TargetOrgUnits: !Ref PlatformOU
```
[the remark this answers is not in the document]
zaro0508: To answer your question, I'm not sure [this example](https://github.com/org-formation/org-formation-cli/blob/master/docs/task-files.md#update-stacks) seems to indicate that it's possible however i'm not sure if it's only possible in context of an `OrganizationBindings`.

if it doesn't work you might want to think about converting the cloudformation template to a nunjucks template so that you can list each OU or account here.

org-formation/090-systems-manager/QuickSetup-PatchPolicy-Prod.yaml
ConsoleCatzirl: Oops, that's a remnant of when I was only going to apply this to prod accounts

org-formation/090-systems-manager/QuickSetup-PatchPolicy-Prod.yaml
ConsoleCatzirl: I don't think that's appropriate, that would unpack the input into a list of strings, but the parameters these are being passed to expect a comma-delimited string, not a list of strings, so I'm pretty sure I would have to repack the values into a single string with `!Join` if this were a `CommaDelimitedList`

org-formation/090-systems-manager/_tasks.yaml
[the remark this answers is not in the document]
ConsoleCatzirl: Unfortunately `*` isn't valid, but we can specify the organization root to target all accounts in the organization. I'll try doing it with a `!Ref`, but I might have to hard-code the value

zaro0508 (approved):
looks good.  Do you plan to remove [the existing patching system](https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/0dafbba661b10a5c7d6fb2f8969640b015f5a87b/org-formation/090-systems-manager/_tasks.yaml#L22-L62) in a follow on PR?

PR #1336, opened 2025-02-12
IT-4204: Add IAM permission boundary for synapse llm lambda exec role

xschildw:
This PR creates a managed policy that's going to be used as an IAM permission boundary. The boundary will be associated with roles created by LLM developers, the roles are assumed by lambda functions.

See https://docs.aws.amazon.com/IAM/latest/UserGuide/access_policies_boundaries.html
and https://aws.amazon.com/blogs/security/when-and-where-to-use-iam-permissions-boundaries/

Note: only put minimal boundary here

### DIFF
--- a/org-formation/600-access/_tasks.yaml
+++ b/org-formation/600-access/_tasks.yaml
@@ -417,3 +417,26 @@ SynapseLlmDeveloperPolicy:
         ]
       }
     PolicyName: SynapseLlmDeveloperPolicy
+
+SynapseLlmLambdaExecRoleBoundary:
+  Type: update-stacks
+  Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.5.1/templates/IAM/managed-policy.yaml
+  StackName: synllm-lambdaexecrole-boundary
+  DefaultOrganizationBinding:
+    IncludeMasterAccount: true
+    Account:
+      - !Ref SynapseLlmProdAccount
+    Region: !Ref primaryRegion
+  Parameters:
+    PolicyDocument: >-
+      {
+        "Version": "2012-10-17",
+        "Statement": [
+          {
+            "Effect": "Allow",
+            "Action": ["logs:*", "bedrock:*" ],
+            "Resource": "*"
+          }
+        ]
+      }
+    PolicyName: SynLlmLambdaExecRoleBoundary


xschildw:
Forgot to add that this policy will be added as dependency to new IAM-related permissions given to the SSO LlmDeveloper role (which currently does not have any IAM permissions beyond iam:PassRole).

zaro0508:
I don't fully understand your idea on how to fix issue IT-4204.  Are you planning to change it so that the SSO LlmDeveloper role is allowed to create other roles but only if they attach this permission boundary?

zaro0508:
>  Overkill?

No, I guess you are using permission boundary as it was intended. I just wasn't sure how you intend for other people to use it. Will they know how to use it and will they know what to do if the boundary is not adequate for them?  If developers just need a common role for a lambda it might make sense for IT to just create it for the developers.

org-formation/600-access/_tasks.yaml
```diff
@@ -417,3 +417,35 @@ SynapseLlmDeveloperPolicy:
...
+  Type: update-stacks
+  Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.5.1/templates/IAM/managed-policy.yaml
+  StackName: synllm-lambdaexecrole-boundary
+  DefaultOrganizationBinding:
+    IncludeMasterAccount: true
+    Account:
+      - !Ref SynapseLlmProdAccount
+    Region: !Ref primaryRegion
+  Parameters:
+    PolicyDocument: >-
```
zaro0508, line 431: please [read this article](https://easytocloud.com/index.php/2020/11/27/iam-permission-boundaries/), it implements a permission boundary for the exact same use case.  Take note of the boiler plate set of permissions you should add to the PB.  I think you need to add them as well..

The next step is to ensure inheritance of the policy. What we’re trying to achieve is the following:
* Any user or role that is created, needs to have this boundary policy attached. In other words, we deny API calls that create users or roles without this PB.
* You cannot attach another boundary policy to an existing user or role. In other words, we deny API calls that try to put a new PB on a user or role unless the PB is this PB.
* You cannot delete this boundary policy from an existing user or role

zaro0508 (changes requested):

xschildw:
@zaro0508 , could you have another look at https://github.com/Sage-Bionetworks-IT/organizations-infra/pull/1336 and https://github.com/Sage-Bionetworks-IT/organizations-infra/pull/1337 when you get a chance? #1336 just creates the policy that's going to be attached as a permission boundary, once it's in I can reference it in #1337 (there's placeholder now) where the developer permissions require roles to have a permission boundary and have a prefix name. Thx!

org-formation/600-access/_tasks.yaml
zaro0508: yes, because you will still want the deny to be there in case  the `PowerUserAccess` policy is removed.

org-formation/600-access/_tasks.yaml
```diff
@@ -430,3 +430,35 @@ SynapseLlmDeveloperPolicy:
         ]
       }
     PolicyName: SynapseLlmDeveloperPolicy
+
+SynapseLlmLambdaExecRoleBoundary:
+  Type: update-stacks
+  Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.5.1/templates/IAM/managed-policy.yaml
+  StackName: synllm-lambdaexecrole-boundary
+  DefaultOrganizationBinding:
+    IncludeMasterAccount: true
```
zaro0508, line 439: if you intention is to deploy only to synapse prod account then this should be `false`

org-formation/600-access/_tasks.yaml
```diff
@@ -430,3 +430,35 @@ SynapseLlmDeveloperPolicy:
         ]
       }
     PolicyName: SynapseLlmDeveloperPolicy
+
+SynapseLlmLambdaExecRoleBoundary:
```
zaro0508, line 434: This is a policy boundry, so how about `SynapseLlmLambdaExecRolePolicyBoundary`?

org-formation/600-access/_tasks.yaml
```diff
@@ -430,3 +430,35 @@ SynapseLlmDeveloperPolicy:
...
+  Type: update-stacks
+  Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.5.1/templates/IAM/managed-policy.yaml
+  StackName: synllm-lambdaexecrole-boundary
```
zaro0508, line 437: how about `synllm-lambdaexecrole-policy-boundary`?

org-formation/600-access/_tasks.yaml
```diff
@@ -430,3 +430,35 @@ SynapseLlmDeveloperPolicy:
...
+            "Action": [
+              "s3:GetObject",
+              "s3:PutObject",
+              "s3:ListBucket"
+            ],
+            "Resource": "*"
+          }
+        ]
+      }
+    PolicyName: SynLlmLambdaExecRoleBoundary
```
zaro0508, line 464: `SynLlmLambdaExecRolePolicyBoundary`?

zaro0508 (changes requested):

PR #1337, opened 2025-02-13
IT-4204: Add IAM permissions to llm dev role

xschildw:
This PR adds IAM permissions to create roles and policies to the LLM developer role.
Depends on #1336 to be deployed.

### DIFF
--- a/org-formation/600-access/_tasks.yaml
+++ b/org-formation/600-access/_tasks.yaml
@@ -413,6 +413,40 @@ SynapseLlmDeveloperPolicy:
             "Effect": "Allow",
             "Action": "iam:PassRole",
             "Resource": "*"
+          },
+          {
+            "Effect": "Allow",
+            "Action": [
+              "iam:CreateRole",
+              "iam:PutRolePermissionsBoundary"
+            ],
+            "Resource": "arn:aws:iam::${AWS::AccountId}:role/synllmrole-*",
+            "Condition": {
+              "StringEquals": {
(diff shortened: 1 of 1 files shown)


org-formation/600-access/_tasks.yaml
```diff
@@ -413,6 +413,40 @@ SynapseLlmDeveloperPolicy:
...
+          {
+            "Effect": "Allow",
+            "Action": [
+              "iam:CreateRole",
+              "iam:PutRolePermissionsBoundary"
+            ],
+            "Resource": "arn:aws:iam::${AWS::AccountId}:role/synllmrole-*",
+            "Condition": {
+              "StringEquals": {
+                "iam:PermissionsBoundary": "${synLlmLambdaExecPermissionBoundary}"
```
xschildw, line 426: ARN of boundary policy needs to be set here (had both resources in same template in original code).

xschildw:
At https://github.com/Sage-Bionetworks-IT/organizations-infra/pull/1337/files#diff-53a0ea2d94923ae47a98d7eec8aaec07f78ae1926779844b5ec54f35a877bef6L398 , should not this be false?

zaro0508 (changes requested):
So if i understand correctly, this will create a policy to allow SSO users to create roles which must contain a  ${synLlmLambdaExecPermissionBoundary}?

xschildw:
Per discussion with John, we (IT/Synapse) should create the necessary roles in the account to be used by LlmDevs.

PR #1339, opened 2025-02-19
[IT-4171] Block SSH/RPD on public Bridge networks

ConsoleCatzirl:
Block all SSH and RPD traffic on public networks used by Bridge.


xschildw (approved):
Do you ever need to access the instances (or is SSM access enabled)?

PR #1343, opened 2025-02-28
[IT-4190] Suppress security hub rule

zaro0508:
I recommend we suppress this finding:
  5.1 Ensure no Network ACLs allow ingress from 0.0.0.0/0 to remote server administration ports

It’s typically normal to allow NACL ingress from all ports. This is in fact how the default AWS VPC is setup.
I think it’s already cumbersome to configure access in AWS, closing off NACL ports without an explicit reason would just makes it more cumbersome and confusing to configure access to resources.  There is an additional mechanism, security groups, that would need to be setup to allow access to resources.


xschildw (approved):
I agree. I think we'd have to make exceptions where it matters anyway.

ConsoleCatzirl (approved):

PR #1350, opened 2025-03-04
[IT-4236] Suppress security hub finding

zaro0508:
suppress finding "S3 Block Public Access setting should be enabled at the bucket-level" for all accounts.

We can remediate this check by enabling all those settings however it would mean that all buckets in an AWS account would need to be private.  We have legittimate use case for public buckets in our AWS accounts therefore remediating this check does not make sense.

### DIFF
--- a/org-formation/075-security-hub/security-hub-suppress-infra.yaml
+++ b/org-formation/075-security-hub/security-hub-suppress-infra.yaml
@@ -379,6 +379,7 @@ Resources:
               - 'arn:aws:securityhub:::ruleset/cis-aws-foundations-benchmark/v/1.2.0/rule/4.1'
               - 'arn:aws:securityhub:::ruleset/cis-aws-foundations-benchmark/v/1.2.0/rule/4.2'
               - 'cis-aws-foundations-benchmark/v/1.4.0/3.5'  # (IT-3619) "3.5 Ensure AWS Config is enabled in all regions"
+              - 'cis-aws-foundations-benchmark/v/1.4.0/2.1.5.1'  # (IT-4193) "2.1.5.1 S3 Block Public Access setting should be enabled"
             Workflow:
               Status:
               - NEW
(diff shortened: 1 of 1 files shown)


org-formation/075-security-hub/security-hub-suppress-infra.yaml
```diff
@@ -379,6 +379,7 @@ Resources:
               - 'arn:aws:securityhub:::ruleset/cis-aws-foundations-benchmark/v/1.2.0/rule/4.1'
               - 'arn:aws:securityhub:::ruleset/cis-aws-foundations-benchmark/v/1.2.0/rule/4.2'
               - 'cis-aws-foundations-benchmark/v/1.4.0/3.5'  # (IT-3619) "3.5 Ensure AWS Config is enabled in all regions"
+              - 'cis-aws-foundations-benchmark/v/1.4.0/2.1.5.1'  # (IT-4193) "2.1.5.1 S3 Block Public Access setting should be enabled"
```
ConsoleCatzirl, line 382: I don't like the idea of suppressing this for all buckets in all accounts, how about adding more bucket prefixes to `SuppressFindingsForPublicBucketsRule`?

ConsoleCatzirl (changes requested):

brucehoff:
Whenever there is something suppressed, the rationale for the decision should be readily available. One way to do this is to add a comment in the `.yaml` file.  If you want to put a Jira ID as a comment in the .yaml file, that's OK but then the rationale must be entered as a comment in the Jira issue.

zaro0508:
> If you want to put a Jira ID as a comment in the .yaml file, that's OK but then the rationale must be entered as a comment in the Jira issue.

@brucehoff @ConsoleCatzirl The rational is in the jira issue, I just had the wrong issue number in the file.  I have updated to the correct issue number.

ConsoleCatzirl (approved):

org-formation/075-security-hub/security-hub-suppress-infra.yaml
ConsoleCatzirl: Oh, I see, 2.1.5.1 is the account-level setting, and 2.1.5.2 is the bucket-level setting

brucehoff (approved):

PR #1359, opened 2025-03-11
[IT-4072] Suppress findings for public Bridge buckets

ConsoleCatzirl:
Add several Bridge buckets to the list for suppressing findings related to public buckets.


zaro0508 (changes requested):
please provide reasoning for suppression of these specific buckets in the issue.

zaro0508 (approved):

PR #1372, opened 2025-04-03
[IT-4287] Use SSM QuickSetup to create a Patch Policy

ConsoleCatzirl:
A second attempt at creating a Patch Policy using SSM QuickSetup
to update all managed nodes in the organization. Add poorly-
documented parameters. Also create an S3 bucket for patch logs
(every member account must have write access to the bucket).

The local deployment roles were created automatically by SSM QuickSetup
by creating an empty patch policy in AWS Portal.

Documentation provided by AWS in a support case:

General process:
https://repost.aws/questions/QUscQKRgKDQ0O7gjhTuioxzw/do-i-need-to-be-in-the-managerment-account-to-use-system-manager-patch-manager-to-patch-instances-across-an-organization
https://docs.aws.amazon.com/systems-manager/latest/userguide/quick-setup-config-types.html
https://docs.aws.amazon.com/systems-manager/latest/userguide/quick-setup-patch-manager.html
https://docs.aws.amazon.com/systems-manager/latest/userguide/quick-setup-getting-started.html
https://docs.aws.amazon.com/systems-manager/latest/userguide/quick-setup-api.html

Patch Baselines:
https://docs.aws.amazon.com/cli/latest/reference/ssm/describe-patch-baselines.html
https://docs.aws.amazon.com/systems-manager/latest/userguide/patch-manager-manage-patch-baselines.html

Permissions:
https://docs.aws.amazon.com/systems-manager/latest/userguide/quick-setup-patch-manager.html#patch-policy-instance-profile-service-role

CloudFormation API:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ssmquicksetup-configurationmanager-configurationdefinition.html
https://docs.aws.amazon.com/quick-setup/latest/APIReference/API_ConfigurationDefinitionInput.html

GitHub examples:
https://github.com/aws-samples/aws-management-and-governance-samples/blob/master/AWSSystemsManager/Quick-Setup-API/patch-policy-examples/patch-policy-cfn-template.yaml
https://github.com/aws-samples/aws-management-and-governance-samples/tree/master/AWSSystemsManager/Quick-Setup-API/patch-policy-examples

### DIFF
--- a/org-formation/090-systems-manager/QuickSetup-PatchPolicy.yaml
+++ b/org-formation/090-systems-manager/QuickSetup-PatchPolicy.yaml
@@ -1,15 +1,46 @@
 AWSTemplateFormatVersion: '2010-09-09'
 Description: An SSM patch policy to patch all instances in the given OUs
 Parameters:
-  TargetOrgUnits:
+  TargetOrgRoot:
     Type: String
     Description: >-
-      Comma separated list of AWS Organizational Units to target.
+      AWS Organization root ID to target.
   TargetRegions:
     Type: String
     Description: >-
-      Comma separated list of AWS Organizational Units to target.
+      Comma separated list of AWS regions to target.
 Resources:
+  PatchLogBucket:
+    Type: "AWS::S3::Bucket"
+    Properties:
+      PublicAccessBlockConfiguration:
+        BlockPublicAcls: true
+        BlockPublicPolicy: true
+        IgnorePublicAcls: true
+        RestrictPublicBuckets: true
+  PatchLogBucketPolicy:
+    Type: 'AWS::S3::BucketPolicy'
+    Properties:
+      Bucket: !Ref PatchLogBucket
+      PolicyDocument:
+        Version: '2012-10-17'
+        Statement:
+          # All member accounts need write access to this bucket
+          - Effect: Allow
+            Principal: '*'
+            Action:
+              - 's3:Get*'
+              - 's3:List*'
+              - 's3:Put*'
+            Resource:
+              - !Sub '${PatchLogBucket.Arn}'
+              - !Sub '${PatchLogBucket.Arn}/*'
+            Condition:
+              # require an account principal associated with our Org
+              StringEquals:
+                "aws:PrincipalOrgID": !Ref TargetOrgRoot
+              ArnLike:
+                "aws:PrincipalArn": "arn:aws:iam::*:root"
   PatchPolicy:
     Type: "AWS::SSMQuickSetup::ConfigurationManager"
     Properties:
@@ -17,8 +48,10 @@ Resources:
       Description: "Patch all SSM managed nodes"
       ConfigurationDefinitions:
       - Type: AWSQuickSetupType-PatchPolicy
+        LocalDeploymentExecutionRoleName: AWS-QuickSetup-StackSet-Local-ExecutionRole
+        LocalDeploymentAdministrationRoleArn: "arn:aws:iam::531805629419:role/AWS-QuickSetup-PatchPolicy-LocalAdministrationRole"
         Parameters:
-          PatchPolicyName: PatchPolicy
+          PatchPolicyName: PatchPolicyAll
           # Patch baselines can be found with the command:
           #   aws ssm describe-patch-baselines |jq '.BaselineIdentities | map({ (.OperatingSystem): (.value = .BaselineId | .label = .BaselineName | .description = .BaselineDescription | .disabled=false | del(.BaselineId, .BaselineName, .OperatingSystem, .BaselineDescription, .DefaultBaseline)) }) | add'
           #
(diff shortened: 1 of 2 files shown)


org-formation/090-systems-manager/QuickSetup-PatchPolicy.yaml
```diff
@@ -0,0 +1,92 @@
...
+            Patch Baseline for Rocky Linux Provided by AWS.","disabled":false},"SUSE":{"value":"arn:aws:ssm:us-east-1:075727635805:patchbaseline/pb-07d8884178197b66b","label":"AWS-SuseDefaultPatchBaseline","description":"Default
+            Patch Baseline for Suse Provided by AWS.","disabled":false},"UBUNTU":{"value":"pb-06e3563bd35503f2b","label":"custom-UbuntuServer-Blog-Baseline","description":"Default
+            Patch Baseline for Ubuntu Provided by AWS.","disabled":false},"WINDOWS":{"value":"pb-016889927b2bb8542","label":"custom-WindowsServer-Blog-Baseline","disabled":false}}'
+          TargetType: '*'
+          TargetOrganizationalRoot: !Ref TargetOrgRoot
+          TargetRegions: !Ref TargetRegions
+Outputs:
+  PatchPolicyArn:
+    Description: 'Patch policy ARN'
+    Value: !Ref PatchPolicyAll
```
zaro0508, line 90: PatchPolicyAll is not a resource so i don't think you can make a reference to it

org-formation/090-systems-manager/QuickSetup-PatchPolicy.yaml
```diff
@@ -0,0 +1,92 @@
+AWSTemplateFormatVersion: '2010-09-09'
+Description: An SSM patch policy to patch all instances in the given OUs
```
zaro0508, line 2: there's no ability to target an OU is there?  `TargetOrgRoot` means the entire organization, correct?

org-formation/090-systems-manager/QuickSetup-PatchPolicy.yaml
```diff
@@ -0,0 +1,92 @@
...
+        Parameters:
+          PatchPolicyName: PatchPolicyAll
+          # Patch baselines can be found with the command:
+          #   aws ssm describe-patch-baselines |jq '.BaselineIdentities | map({ (.OperatingSystem): (.value = .BaselineId | .label = .BaselineName | .description = .BaselineDescription | .disabled=false | del(.BaselineId, .BaselineName, .OperatingSystem, .BaselineDescription, .DefaultBaseline)) }) | add'
+          #
+          # More information at:
+          #   https://aws.amazon.com/blogs/mt/deploy-aws-systems-manager-quick-setup-programmatically-across-your-aws-organization/
+          #   https://github.com/aws-samples/aws-management-and-governance-samples/blob/master/AWSSystemsManager/Quick-Setup-API/patch-policy-examples/patch-policy-cfn-template.yaml
+          #   https://github.com/aws-samples/aws-management-and-governance-samples/blob/master/AWSSystemsManager/Quick-Setup-API/patch-policy-examples/default-patch-baselines.txt
+          ConfigurationOptionsPatchOperation: "ScanAndInstall"
```
zaro0508, line 62: is this going to installing security patches only? or will it do more than that? Also do these patches require re-booting resources?

org-formation/090-systems-manager/_tasks.yaml
```diff
@@ -61,6 +61,18 @@ PatchMembers:
   Parameters:
     ManagementAccountNumber: !Ref accountId
 
+PatchPolicy:
```
zaro0508, line 64: how about naming it `QuickSetupPatchPolicy`?   

Does this conflict with the existing patch setup?  Is the idea to install this then remove the old setup or do we need to remove the old setup first?

zaro0508 (changes requested):

org-formation/090-systems-manager/_tasks.yaml
ConsoleCatzirl: No, they don't conflict, we can run both at the same time. The idea is to add this new process, then remove the old setup afterwards.

org-formation/090-systems-manager/QuickSetup-PatchPolicy.yaml
ConsoleCatzirl: Great question. The short answer is yes, it only installs security patches, but how they're selected depends on the flavor of Linux[1]. And there's a `RebootOption` parameter to enable rebooting when required[2]. I left it out to use the default of no reboot, but I can make that explicit.

[1] https://docs.aws.amazon.com/systems-manager/latest/userguide/patch-manager-selecting-patches.html
[2] https://docs.aws.amazon.com/quick-setup/latest/APIReference/API_ConfigurationDefinitionInput.html

org-formation/090-systems-manager/QuickSetup-PatchPolicy.yaml
ConsoleCatzirl: whoops

org-formation/090-systems-manager/QuickSetup-PatchPolicy.yaml
```diff
@@ -0,0 +1,95 @@
...
+    Properties:
+      PublicAccessBlockConfiguration:
+        BlockPublicAcls: true
+        BlockPublicPolicy: true
+        IgnorePublicAcls: true
+        RestrictPublicBuckets: true
+  QuickSetupPatchLogBucketPolicy:
+    Type: 'AWS::S3::BucketPolicy'
+    Properties:
+      Bucket: !Ref QuickSetupPatchLogBucket
```
zaro0508, line 24: i assume this bucket will need to be excluded in security hub findings for `2.1.5.2 S3 Block Public Access setting`?

zaro0508 (approved):

org-formation/090-systems-manager/QuickSetup-PatchPolicy.yaml
ConsoleCatzirl: no it shouldn't need an exception, blocking public access is set to true

PR #1377, opened 2025-04-05
PLFM-8908: Add role to allow organization to list imagebuilder images

brucehoff:
In PLFM-8908 we create a Image Builder pipeline to build an AMI for Synapse weekly.  We need to be able to look-up the AMI ID of the image created by the pipeline from the Synapse (dev and prod) accounts.  This PR adds a role that anyone in the Organization can assume and which will allow them to call `imagebuilder get-image`.


zaro0508 (changes requested):
I don't understand why this role is needed or how it's used.  Would you be able to better describe how you intend to use this role or provide a PR that shows its usage?  Note that work in [IT-3831](https://sagebionetworks.jira.com/browse/IT-3831?focusedCommentId=232834) was to setup a shared AMI repository so that when images get created they are accessible from any repo.  Docs are also in the [confluence wiki(https://sagebionetworks.jira.com/wiki/spaces/IT/pages/914882624/AWS+Image+Repository)

brucehoff:
> I don't understand why this role is needed or how it's used. Would you be able to better describe how you intend to use this role or provide a PR that shows its usage? 

Yes, the role is to be used to find/lookup the IDs of the AMIs generated by an Image Builder Pipeline.  It is used here: 
https://github.com/Sage-Bionetworks/Synapse-Stack-Builder/pull/703/files#diff-b380fae366976125ec652d9f464768c08a42a2cbeaf5e867359c9ba7e09d65feR208

The AWS API being used is https://docs.aws.amazon.com/imagebuilder/latest/APIReference/API_ListImagePipelineImages.html

Note that the Synapse Stack Builder is run in the Synapse (dev/prod) account, not in Image Central, which is why we need this role.


> Note that work in [IT-3831](https://sagebionetworks.jira.com/browse/IT-3831?focusedCommentId=232834) was to setup a shared AMI repository so that when images get created they are accessible from any aws account in our organization. Docs are also in the [confluence wiki(https://sagebionetworks.jira.com/wiki/spaces/IT/pages/914882624/AWS+Image+Repository)

Note that sharing AMIs does not also allow others to list the AMIs for an image pipeline.  Again, a role like this one is needed to extend that permission.

zaro0508:
> Yes, the role is to be used to find/lookup the IDs of the AMIs generated by an Image Builder Pipeline
Doesn't the stack builder already use a role to deploy to AWS?  I'm assuming, and hope, that it's using a Github OIDC associated role?  if so why not just add `AWSImageBuilderReadOnlyAccess` policy to that role?

Also I wouldn't mind adding `AWSImageBuilderReadOnlyAccess` policy to the `SsoViewer` and `SsoViewerPlus` roles in [our organization SSO roles](https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/org-formation/700-aws-sso/_tasks.yaml).  So that SSO viewers can search for shared AMIs in the AWS console.

brucehoff:
> > Yes, the role is to be used to find/lookup the IDs of the AMIs generated by an Image Builder Pipeline
> 
> Doesn't the stack builder already use a role to deploy to AWS? I'm assuming, and hope, that it's using a Github OIDC associated role? if so why not just add `AWSImageBuilderReadOnlyAccess` policy to that role instead of creating a new role?
> 
The stack builder runs in a different account, a Synapse dev' or prod' account, not in image-central, so image-central needs to grant cross-account access. That is the purpose of this role.  Doing what you suggest wouldn't work, it would only grant `AWSImageBuilderReadOnlyAccess` in the Synapse dev' or prod' account.


> Also I wouldn't mind adding `AWSImageBuilderReadOnlyAccess` policy to the `SsoViewer` and `SsoViewerPlus` roles in [our organization SSO roles](https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/org-formation/700-aws-sso/_tasks.yaml). So that SSO viewers can search for shared AMIs in the AWS console.

You are welcome to do so for the benefit of `SsoViewers` in the the image-central account, but we still need to grant cross account access to the Synapse accounts, hence the need for this PR.

zaro0508:
> The AWS API being used is https://docs.aws.amazon.com/imagebuilder/latest/APIReference/API_ListImagePipelineImages.html

Ahh ok, I must have missed this comment.  I am confusing access to the AMI vs access to API.  Sorry about that, and thanks for the info. 

> You are welcome to do so for the benefit of `SsoViewers` in the the image-central account, but we still need to grant cross account access to the Synapse accounts, hence the need for this PR.

Done in PR #1380

zaro0508 (changes requested):
this looks generally fine however I would like to request to move this to `org-formation/600-access/_tasks.yaml`.  The `sceptre` folder is setup for legacy deployments, before we started using org-formation.  We prefer all infra stuff deployed from org-formation going forward.

Also can you please provide a few comments yaml file(s) on what is making use of this role?

brucehoff:
> this looks generally fine however I would like to request to move this to `org-formation/600-access/_tasks.yaml`. The `sceptre` folder is setup for legacy deployments, before we started using org-formation. We prefer all infra stuff deployed from org-formation going forward.

Could you please add some guidance (in a ReadMe, maybe) to keep people from making this mistake in the future?

> 
> Also can you please provide a few comments in the yaml file(s) on what is making use of this role?

Certainly.

zaro0508:
> Could you please add some guidance (in a ReadMe, maybe) to keep people from making this mistake in the future?

done in PR #1381

brucehoff:
> this looks generally fine however I would like to request to move this to `org-formation/600-access/_tasks.yaml`. The `sceptre` folder is setup for legacy deployments, before we started using org-formation. We prefer all infra stuff deployed from org-formation going forward.
> 
> Also can you please provide a few comments in the yaml file(s) on what is making use of this role?

Done!

org-formation/600-access/_tasks.yaml
```diff
@@ -387,3 +387,18 @@ SynapseLlmDeveloperPolicy:
         ]
       }
     PolicyName: SynapseLlmDeveloperPolicy
+
+# Setup cross-account access to Image Builder Pipeline in Image Central account from Synapse accounts
+SynapseImageCentralImageBuilderPipelineAccess:
```
zaro0508, line 392: This name seems overly specific to me.  I would remove `Synapse` because I think it would be useful to allow all member accounts access to this role.

zaro0508 (approved):

org-formation/600-access/_tasks.yaml
```diff
@@ -387,3 +387,18 @@ SynapseLlmDeveloperPolicy:
...
+  Type: update-stacks
+  Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.2.10/templates/IAM/cross-account-access.yaml
+  StackName: synapse-image-builder-pipline-read-access
```
zaro0508, line 395: same here

PR #1379, opened 2025-04-09
[IT-4269] Add data-transfer budget alerts

ConsoleCatzirl:
Send email notifications when total data-transfer costs across all accounts approach 10k (approximately 15% of our average cloud spend).

depends on https://github.com/Sage-Bionetworks/aws-infra/pull/426

### DIFF
--- a/org-formation/040-budgets/_tasks.yaml
+++ b/org-formation/040-budgets/_tasks.yaml
@@ -12,3 +12,13 @@ BudgetAlarms:
     budgetName: !Sub '${resourcePrefix}-budget-${CurrentAccount.AccountId}'
     budgetAmount: !GetAtt CurrentAccount.Tags.budget-alarm-threshold
     emailAddress: !GetAtt CurrentAccount.Tags.budget-alarm-threshold-email-recipient
+
+BudgetAlarmTransferCosts:
+  Type: update-stacks
+  Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.9.7/templates/Billing/budgets-transfer-costs.yaml
+  StackName: !Sub '${resourcePrefix}-data-transfer-alarm'
+  DefaultOrganizationBinding:
+    Account: !Ref MasterAccount
+    Region: !Ref primaryRegion
+  Parameters:
+    budgetAmount: 10000


org-formation/040-budgets/_tasks.yaml
```diff
@@ -12,3 +12,14 @@ BudgetAlarms:
     budgetName: !Sub '${resourcePrefix}-budget-${CurrentAccount.AccountId}'
     budgetAmount: !GetAtt CurrentAccount.Tags.budget-alarm-threshold
     emailAddress: !GetAtt CurrentAccount.Tags.budget-alarm-threshold-email-recipient
+
+BudgetAlarmTransferCosts:
+  Type: update-stacks
+  Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.9.5/templates/Billing/budgets-transfer-costs.yaml
+  StackName: !Sub '${resourcePrefix}-billing-alarm'
+  DefaultOrganizationBinding:
+    Account: !Ref MasterAccount
```
zaro0508, line 21: we are setting up a budget for the organization (i.e. all accounts)?  Would the notification tell us which specific account is busting the budget?

org-formation/040-budgets/_tasks.yaml
```diff
@@ -12,3 +12,14 @@ BudgetAlarms:
     budgetName: !Sub '${resourcePrefix}-budget-${CurrentAccount.AccountId}'
     budgetAmount: !GetAtt CurrentAccount.Tags.budget-alarm-threshold
     emailAddress: !GetAtt CurrentAccount.Tags.budget-alarm-threshold-email-recipient
+
+BudgetAlarmTransferCosts:
+  Type: update-stacks
+  Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.9.5/templates/Billing/budgets-transfer-costs.yaml
+  StackName: !Sub '${resourcePrefix}-billing-alarm'
```
zaro0508, line 19: this name seems kinda generic.  can we make it more specific?

zaro0508 (changes requested):

org-formation/040-budgets/_tasks.yaml
ConsoleCatzirl: no, the notification would just say that total data transfer costs have reached a certain number. someone would have to go to cost explorer to figure out what service and region the costs were coming from.

zaro0508: We could just decrease the budget and deploying to every account.  Is there a reason for not doing that?

ConsoleCatzirl: Do you mean decreasing the budget threshold tagged on the accounts and used by the stack `BudgetAlarms` created in this same file? Those budgets track the total charges in the tagged account, including both direct and indirect costs. They're useful for alerting a study when they go over their monthly grant budget.
This new budget will filter costs to only track charges related to data transfer, and sum those charges across all accounts. This is useful for the organization as a whole to know when total data transfer costs (an indirect charge) reach 15% of our average monthly cloud-spend.

zaro0508: > Do you mean decreasing the budget threshold tagged on the accounts and used by the stack BudgetAlarms created in this same file?

I'm referring to the `budgetAmount` parameter a few lines below this.  That's the budget for transfer cost, correct?

ConsoleCatzirl: Yeah, that's setting a 10k/month budget for data transfer

ConsoleCatzirl:
@zaro0508 can you take another look at this?

zaro0508 (approved):

PR #1390, opened 2025-04-21
[IT-4089] Suppress access-key rotation for service users

ConsoleCatzirl:
Don't require the Bridge service users to rotate access keys every 90 days. A lack of certainty around knowing all locations to update creates a risk of a prolonged outage with data loss.

Also refactor suppression for credentials unused for 45 days so that it is also restricted to the Bridge service users.


ConsoleCatzirl:
Yeah, I would love to move to OIDC, but Bridge is still using Travis so it's not worth the effort.

xschildw (approved):

PR #1394, opened 2025-04-24
IT-4382: Change directory in which Nessus agent installation script is installed

brucehoff:
CIS Level 1 hardened AMI removes execution permission for /tmp, breaking the installation of the StackArmor Nessus agent installation script.  So this PR updates the download location for the script to be executed to the user's home directory.


xschildw (approved):
The scripts are executed as root, correct?

brucehoff:
> The scripts are executed as root, correct?

Yes, note the first line of the run book (line 323), "sudo su root"

PR #1395, opened 2025-04-24
[IT-4380] Update redirects

zaro0508:
Update application endpoints for Agora

This will require one additional manual step of updating the route53
record `agora.adknowledgeportal.org` to forward to
`agora-prod.adknowledgeportal.org`


JessterB (approved):
LGTM

PR #1396, opened 2025-04-24
It 4382: Change name of Automation-Run-Script document in Scheduled-Script-Automation

brucehoff:
Build fails with:
> ERROR: Resource AutomationDocumentScriptExecution failed because CloudFormation cannot update a stack when a custom-named resource requires replacing. Rename Automation-Run-Script and update the stack again.

I tried deleting the SSM Document, `Automation-Run-Script` and rerunning the broken build, but it didn't work.  So the next thing to try is renaming the Document.


org-formation/090-systems-manager/Scheduled-Script-Automation.yaml
```diff
@@ -277,7 +277,7 @@ Resources:
     Type: "AWS::SSM::Document"
     Properties:
       DocumentType: Automation
-      Name: "Automation-Run-Script"
+      Name: "Automation-Run-Script-v2"
```
xschildw, line 280: It's not required, could we just omit it?

brucehoff: Maybe, but the document does show up in the AWS Console (under the SSM Documents section) and it would be nice for it to have a meaningful, human-readable name.

xschildw: I think that's what is does when not specified (name of the resource (possibly truncated) and some random string at the end). My only concern is if next time we modify it we forget to rename it.

brucehoff: OK, I have removed the name.

xschildw (approved):

PR #1398, opened 2025-04-30
IT-4369: Update svc-catalog-lib-version for dev

xschildw:
This PR updates the SC products to 1.4.10 which adds the policy for SSM patching to the instance role


zaro0508 (approved):
LGTM, just wondering why you've restricted to dev environment.  if we've decided to use SSM quick setup you might as well deploy to prod as well.

xschildw:
I want to confirm that it works (all policies are attached to instance role, and I can still terminate the instance).

PR #1403, opened 2025-05-07
[IT-4166] suppress security hub finding 2.1.5.2

zaro0508:
Another attempt at PR #1399 which was reverted in PR #1402

My theory is that the `AWS::Events::Rule` has a character size limit
and PR #1401 bumped it over the limit.  This PR creates a separate
suppression rule.


xschildw (approved):
Can't we use wildcards in the prefixes?

zaro0508:
> Can't we use wildcards in the prefixes?

how would you propose to do that?  bucket names are unique.

PR #1409, opened 2025-05-19
[IT-4166] Test suppression rule

zaro0508:
None of the buckets in `SuppressFindingsForPublicBucketsRule2` are being suppressed.  We suspect that wildcards *might* be invalidating the rule

We are testing that theory by commenting out the finding IDs that contain wilcards


xschildw (approved):

org-formation/075-security-hub/security-hub-suppress-infra.yaml
```diff
@@ -484,14 +484,14 @@ Resources:
                 - prefix: 'arn:aws:s3:::mobiletoolbox.org'
                 - prefix: 'arn:aws:s3:::privacytoolkit.sagebase.org'
                 - prefix: 'arn:aws:s3:::privacytoolkit.sagebionetworks.org'
-                - prefix: 'arn:aws:s3:::prod*'
+              #  - prefix: 'arn:aws:s3:::prod*'
```
ConsoleCatzirl, line 487: This is matching on the prefix of the bucket name, so we should be able to remove just the `*` character

PR #1411, opened 2025-05-20
[IT-4418] Remove resources for old Agora

zaro0508:
Remove AWS resources specific to the old Agora elastic beanstalk infrastructure environment.

### DIFF
--- a/org-formation/710-tgw/_tasks.yaml
+++ b/org-formation/710-tgw/_tasks.yaml
@@ -36,7 +36,6 @@ TransitGateway:
       - !Ref BridgeProdAccount
       - !Ref SynapseDevAccount
       - !Ref SynapseProdAccount
-      - !Ref AgoraDevAccount
       - !Ref iAtlasProdAccount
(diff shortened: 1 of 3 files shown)


org-formation/710-tgw/_tasks.yaml
```diff
@@ -36,7 +36,6 @@ TransitGateway:
       - !Ref BridgeProdAccount
       - !Ref SynapseDevAccount
       - !Ref SynapseProdAccount
-      - !Ref AgoraDevAccount
```
ConsoleCatzirl, line 39: Why is Agora being removed from transit gateway? Does this mean that Agora v3 doesn't need a vpc for vpn-only access?

zaro0508: The VPC for this agora deployment is being deleted.  Also I believe we initially set this up to allow access to resources in the VPC thru the VPN but now we use SSM sessions therefore it's not really needed anymore.

ConsoleCatzirl (approved):

PR #1414, opened 2025-06-04
Plfm 8754 synapse-ops repo's to deploy Synapse infrastructure

brucehoff:
Allow the synapse-ops repo's to deploy to the Synapse AWS Accounts.

### DIFF
--- a/org-formation/650-identity-providers/_tasks.yaml
+++ b/org-formation/650-identity-providers/_tasks.yaml
@@ -158,17 +158,53 @@ GithubOidcSageBionetworksSynapse:
   TemplatingContext:
     GitHubOrg: "Sage-Bionetworks"
     Repositories:
-      - name: "synapse-docker-registry"
-        branches: ["*"]
       - name: "nbconvert-webapp"
         branches: ["master", "develop"]
+  DefaultOrganizationBinding:
+    Account:
+      - !Ref SynapseDevAccount
+      - !Ref SynapseProdAccount
+    Region: us-east-1
+
+GithubOidcSageBionetworksSynapseOpsDev:
+  Type: update-stacks
+  DependsOn: GithubOidcSageBionetworks
+  Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.7.6/templates/IAM/github-oidc-provider.j2
+  StackName: !Sub ${resourcePrefix}-${appName}-sage-bionetworks-synapse-ops-dev
+  Parameters:
+    ProviderArn: !CopyValue [ !Sub '${resourcePrefix}-${appName}-ProviderArn' ]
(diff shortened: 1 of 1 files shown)


org-formation/650-identity-providers/_tasks.yaml
```diff
@@ -162,6 +162,10 @@ GithubOidcSageBionetworksSynapse:
         branches: ["*"]
       - name: "nbconvert-webapp"
         branches: ["master", "develop"]
+      - name: "synapse-ops-dev"
+        branches: ["main"]
+      - name: "synapse-ops-prod"
+        branches: ["main"]
```
zaro0508, line 168: why do we have two repos for dev and prod?  can we not deploy with parametrized values? 

also you are setting up a `synapse-ops-dev` repo to deploy to `SynapseProdAccount` account.  Are you sure you want to allow that?

zaro0508 (changes requested):

org-formation/650-identity-providers/_tasks.yaml
brucehoff: > you are setting up a synapse-ops-dev repo to deploy to SynapseProdAccount account. Are you sure you want to allow that?

That is a good point.  I updated the PR to split the role into two, one role that can only access the Synapse Dev account and can only be used by the synapse-ops-dev repo and another that can only access the Synapse Prod account and can only be used by the synaspe-ops-prod repo.

> why do we have two repos for dev and prod? can we not deploy with parametrized values?

We use two repo's to isolate the self-hosted runners, deployed into the Synapse Dev and Prod accounts.  Each runner is scoped to its repo' and this way the dev' runner can only be accessed by the dev' repo while the prod runner can only be accessed by the prod repo.

xschildw (approved):

org-formation/650-identity-providers/_tasks.yaml
zaro0508: What's up? you didn't even give me a chance to re-review before merging?  My dis approval still applies to this PR.

anyways, I was gonna say you probably don't need separate repos for isolation if you use aws codedeploy as suggested in PLFM-8754.  You could have github trigger code deploy to provision the runners which means you wouldn't even need to setup runners in github at all.

brucehoff: > you didn't even give me a chance to re-review before merging? 

The PR was approved by @xschildw .  However I am happy to receive your feedback too.

> You could have github trigger code deploy

Thank you for the suggestion.  I would like to use GitHub as the "UI" for launching the various workflows and it works pretty seamlessly with hosted runners to run the jobs without introducing any additional components.  I like the idea of separate repo's because (1) it's pretty simple to set up and maintain (the second repo' will  just be a fork of the first) and (2) it reduces the chance of catastrophic user error (doing something in prod' when dev' was intended) if the user has to switch to an entirely different repo'.

PR #1434, opened 2025-07-07
IT-4497 fix in-line policy

brucehoff:
CF does not seem to allow ImportValue within an in-line JSON policy.  The previous PR [failed](https://github.com/Sage-Bionetworks-IT/organizations-infra/actions/runs/16125791680/job/45502264068#step:12:1675) with:
"Invalid PermissionsPolicy Document".   This PR tries instead to use the Fn::Sub approach.


org-formation/700-aws-sso/_tasks.yaml
```diff
@@ -435,8 +435,9 @@ SsoDeveloper:
       - 'arn:aws:iam::aws:policy/AWSBillingReadOnlyAccess'
       - 'arn:aws:iam::aws:policy/AmazonBedrockFullAccess'
     sessionDuration: 'PT12H'
-    inlinePolicy: >-
-      {
+    inlinePolicy:
```
xschildw, line 438: This is a parameter to a template, can we actually use a function here? From the previous failure it looks like it was passed straight through.

xschildw: Let's try...

xschildw (approved):

PR #1435, opened 2025-07-07
IT-4497

brucehoff:
Previous PR [failed](https://github.com/Sage-Bionetworks-IT/organizations-infra/actions/runs/16128159759/job/45509922042#step:10:15) with

> ERROR: !Sub first element must be string


zaro0508 (changes requested):
I don't think we have ever tried configuring a solution for IT-4497 before.  Are you sure that this policy will even work?  if not, I would suggest you just hardcode the allow role ARN into the inline policy to make the CI deploy it then validate that it works.  After that you can go back to improving this by updating it to a cloudformation reference.

PR #1446, opened 2025-07-30
IT-4204: add bedrock execution role

xschildw:
This PR creates a role to be passed to the Bedrock service and giving it permission to invoke models and interact with OpenSearch. Once created, we need to give the SynapseLlmDeveloper role the right to pass that role to Bedrock.


xschildw:
DPE manages the SynapseLlmProd account, devs have admin access and can create IAM roles as needed.

PR #1467, opened 2025-09-16
Plfm 9092

brucehoff:
OIDC integration allowing developer forks of Synapse repo to access Synapse dev' account

depends on sage-bionetworks/aws-infra#437 and assumes a tag on that repo' of v0.10.4


org-formation/650-identity-providers/_tasks.yaml
```diff
@@ -166,6 +166,44 @@ GithubOidcSageBionetworksSynapse:
...
+    ProviderArn: !CopyValue [ !Sub '${resourcePrefix}-${appName}-ProviderArn' ]
+    ProviderRoleName: !Sub ${resourcePrefix}-${appName}-sage-bionetworks-synapse-build
+    ManagedPolicyArns:
+      - "arn:aws:iam::aws:policy/AdministratorAccess"
+  TemplatingContext:
+    Repositories:
+      - owner: "Sage-Bionetworks"
+        name: "Synapse-Repository-Services"
+        branches: ["*"]
+      - owner: "brucehoff"
```
ConsoleCatzirl, line 184: Do developers need to deploy from their forks? Could each developer have a branch on the origin repo instead?

Also, I don't know much about the platform build process, but if these are all deploying to the same EC2 then I worry about developers breaking the build with parallel deploys, or one failed build blocking all other developers.

PR #1572, opened 2026-04-15
PLFM-4151: Use wildcards to reduce policy size

xschildw:
This PR fixes #1570 which exceeded the size limit on policy.
Use prefix 'dev', 'staging', 'prod' for the buckets (+ grandfather existing buckets that don't fit the pattern')
Use '*' for distribution (the action is 'invalidate' which is non-destructive).


zaro0508:
an alternative to using wildcard is to minify the json, take a look at this example https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/org-formation/600-access/_tasks.yaml#L190-L192

xschildw:
Is that not just postponing the inevitable? We already split because of size. The wildcard has the extra advantage that if the pattern is followed, there's no need for a PR to give permissions when creating a new portal.

zaro0508:
> Is that not just postponing the inevitable?

maybe?  there could be other things we can do 😄 

> The wildcard has the extra advantage that if the pattern is followed, there's no need for a PR to give permissions when creating a new portal.

I guess that can also be a detriment if you don't want the same setup for every bucket that's created.  Anyway this is just a suggestion, if you prefer wildcard for these buckets then it's fine by me.

zaro0508 (approved):

xschildw:
In this context (synapsemonorepo portals), I think it's fine to wildcard following the deployment pattern. Looking at the buckets, this covers about 2/3 of the buckets. In the remaining 1/3 I see some that look like portals (from name), need to figure out if still active and if so how they're used.

PR #1582, opened 2026-04-29
[IT-5017] AWS account for clients to use bedrock

zaro0508:
Create a centralized Bedrock account. This keeps all AI spend in one place,
makes credit application straightforward and gives us a single pane for quotas,
logging, guardrails, and cost reporting.

We use this account to allow clients such as claude code to use Bedrock from
this account.  This allows us to use bedrock credits that will be applied to this account.

#### PR Dependency Tree

* **PR #1582** 👈
  * **PR #1583**

### DIFF
--- a/org-formation/organization.yaml
+++ b/org-formation/organization.yaml
@@ -77,7 +77,7 @@ Organization:
         - !Ref RestrictDnsRegistrationSCP
       Accounts:
         - !Ref SceptreDevAccount
-        - !Ref BedrockExternalAccount
+        - !Ref BedrockCentralAccount
 
   ItProdOU:
     Type: OC::ORG::OrganizationalUnit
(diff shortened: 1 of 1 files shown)


ConsoleCatzirl (approved):

org-formation/organization.yaml
```diff
@@ -697,6 +698,21 @@ Organization:
         budget-alarm-threshold: 2000
         budget-alarm-threshold-email-recipient: aws-sagebrain-dev@sagebase.org
 
+  # This account is for external-facing Bedrock workloads
+  # External clients (claude, copilot, etc..) will use Bedrock in this account
+  BedrockExternalAccount:
```
brucehoff: To me the use of the word "external" is confusing.  It sounds like the account is external to our organization or that the users are external to Sage.  Is there a different word we can use here?

brucehoff (changes requested):

org-formation/organization.yaml
zaro0508: of course but I'm bad with names.  By all means, you are welcome to make suggestions.

Since this is an account that is a store of bedrock credits, how about `BedrockPool` or more financial related `BedrockBank`?

brucehoff: > of course but I'm bad with names. By all means, you are welcome to make suggestions.
> 
> Since this is an account that is a store of bedrock credits, how about `BedrockPool` or more financial related `BedrockBank`?

I'm bad with names too, but I like your idea of using an existing naming convention (BedrockPool) or even Bedrock-Central, if you think that's suitable.  Thank you.

brucehoff (changes requested):

org-formation/organization.yaml
zaro0508: ok, named it `org-sagebase-bedrockcentral`.

brucehoff (approved):

PR #1584, opened 2026-04-30
IT-4981: Add temporary redirector from modeladexplorer.org to prod.modeladexpl…

xschildw:
This PR sets up a temporary redirect to get around a certificate issue.


zaro0508 (approved):
LGTM but why is it temporary and what is the certificate issue?  if that's significant then a comment in the code would be nice.

xschildw:
You're right, added comment. Will open issue to fix cert and remove this.

PR #1587, opened 2026-05-04
[IT-5049] Reduce Vanta scanner scope

zaro0508:
Instead of scanning the entire AWS account, scope the vanta
scanner to only scan synapse production accounts

#### PR Dependency Tree

* **PR #1587** 👈

### DIFF
--- a/org-formation/600-access/vanta-readonly-role-cf.yml
+++ b/org-formation/600-access/vanta-readonly-role-cf.yml
@@ -1,4 +1,5 @@
 # This template was provided by Vanta
+# From Integrations->Add connection->Cloudformation->AWS->Policy and role
 AWSTemplateFormatVersion: 2010-09-09
 Parameters:
   ExternalId:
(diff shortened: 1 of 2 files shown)


org-formation/600-access/vanta-readonly-role-cf.yml
brucehoff, line 18: Should this name be changed from `VantaAdditionalPermissionsPolicy` to `VantaExclusionPermissionsPolicy`?  It seems to only remove permissions, not add any additional ones?

brucehoff (changes requested):

org-formation/600-access/vanta-readonly-role-cf.yml
zaro0508: Look at the comment on lines 1-2.  This template is provided by Vanta as-is.  I would prefer to not make any modifications to it.  The existing template was for scanning an organization account, this updated one is for scanning member accounts.

image

brucehoff: ok

brucehoff (approved):

PR #1602, opened 2026-05-21
[IT-5070] Change patch schedule to daily

zaro0508:
The patch manager is set to patch instances on a weekly schedule
which can leave instances vulnerable to security vulnerabilities.
    
We switch the schedule from weekly (Tuesdays at 10:15 PM UTC) to
daily (every day at 10:15 PM UTC).

#### PR Dependency Tree

* **PR #1602** 👈


brucehoff (approved):
LGTM.  I'm not an expert on CRON expressions, so I assume this is correct.

brucehoff (approved):

PR #1614, opened 2026-06-12
Add user-specific ARN for BenchmarkCollaboratorAccess

Bankso:
Replaces root ARN previously listed for collaborator account

### DIFF
--- a/org-formation/600-access/_tasks.yaml
+++ b/org-formation/600-access/_tasks.yaml
@@ -225,7 +225,7 @@ BenchmarkCollaboratorAccess:
     Region: us-east-1
   Parameters:
     PrincipalArns:
-      - arn:aws:iam::246534652589:root     # Nikolaos Kalavros
+      - arn:aws:iam::246534652589:user/OrionLLMBenchUser     # Nikolaos Kalavros
     PolicyDocument: !Sub >-
       {
         "Version": "2012-10-17",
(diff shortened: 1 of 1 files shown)


org-formation/600-access/_tasks.yaml
```diff
@@ -225,7 +225,7 @@ BenchmarkCollaboratorAccess:
     Region: us-east-1
   Parameters:
     PrincipalArns:
-      - arn:aws:iam::246534652589:root     # Nikolaos Kalavros
+      - arn:aws:iam::246534652589:user/OrionLLMBenchUser     # Nikolaos Kalavros
```
zaro0508: what's up with the `# Nikolaos Kalavros` comment?

Bankso: I was following the pattern for other configurations in the `Roles for cross-account access` section, where the user's name was listed as a comment (like lines 105 and 106). I can remove it if not needed!

zaro0508: no, i mean why is the user `OrionLLMBenchUser` and the comment referring to someone else?

Bankso: I see what you're saying - this is what was provided to me when I asked our collaborator if we could get a user-specific ARN for this config. `OrionLLMBenchUser` was the name they assigned when creating the user in their account, I'm assuming to help them easily identify if they've assumed the correct role. `Nikolaos Kalavros` is the name of the collaborator being given access.

zaro0508: that is a confusing.  would you be able to provide a comment that explains this situation better? you can add it on the line above this ARN.

Bankso: Just added a comment, please let me know if I should add any additional information.

zaro0508 (approved):

PR #1616, opened 2026-06-16
IT-5066: Remove unsupported AWS::GuardDuty::OrganizationConfiguration resource; add companion CLI script for org auto-enable

brucehoff:
Remove unsupported AWS::GuardDuty::OrganizationConfiguration resource; add companion CLI script for org auto-enable

`OrganizationConfiguration` is apparently unsupported in CloudFormation.  It was used to configure GuardDuty to enroll new accounts in all regions.  This PR:

1) Removes `OrganizationConfiguration` from `org-formation/070-guard-duty/guardduty.yaml`

2) Adds a one-time script to configure all the regons.

Note: We considered automating the running of the script as part of the deployment however doing so would add a lot of complexity, requiring either (1) creating a lambda to run the script and wiring it up to CF as a custom resource or (2) adding a new OIDC role for the GitHub workflow to use in the SecurityCentral account so it can run the script as a workflow job.  Since the script only needs to be run once, it's simplest to leave it to be run manually, once.


org-formation/070-guard-duty/configure-org-auto-enable.sh
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env bash
```
zaro0508, line 1: AWS guardduty already supports automatic enabling of new accounts.  https://docs.aws.amazon.com/guardduty/latest/ug/set-guardduty-auto-enable-preferences.html

Therefore I don't understand why we even need this at all?

zaro0508 (changes requested):

org-formation/070-guard-duty/configure-org-auto-enable.sh
brucehoff: Yes, this script invokes the very feature you linked to.  Without this script you would have to select it in the AWS console, once for every region.

zaro0508 (approved):

PR #1618, opened 2026-06-23
configure aws open data buckets for strides migrations

wpoehlm:
As outlined in https://sagebionetworks.jira.com/browse/PSI-95, we are migrating 7 studies in the AD Knowledge Portal into AWS Open Data. This PR configures an AWS Open Data bucket for each study, with access granted to myself and @thomasyu888. Data will be migrated using the SOP outlined here: https://sagebionetworks.jira.com/wiki/spaces/SYNPY/pages/3132915721/Synapse+Project+Data+Migration.


thomasyu888 (approved):
LGTM, but will defer to IT personnel for final review.

xschildw (approved):
LGTM

PR #1619, opened 2026-07-02
[PLFM-9782] Grant synapsedev Developer scoped IAM delete permissions

BryanFauble:
## Summary
Adds `iam:DeleteRolePolicy` / `iam:DeleteRole` / `iam:DetachRolePolicy` to the shared `Developer` SSO permission set's inline policy, scoped to synapsedev roles named `*shared-resources*`, so developers can clean up `DELETE_FAILED` personal sandbox CloudFormation stacks.

Developers get the cleanup ability in the `Developer` role they already assume — no separate role to switch into. (AWS Identity Center permission sets don't compose/inherit, so a single usable role means adding the statement to the Developer set itself.)

### Why this is safe despite the permission set being shared across ~20 accounts
`iam:DeleteRole*` / `iam:DetachRolePolicy` act on IAM roles in the caller's own account, and the statement's `Resource` is pinned to the synapsedev account ID (`arn:aws:iam::<synapsedev>:role/*shared-resources*`). In any other account this permission set is assigned to, a caller's delete action targets that account's roles, whose ARNs can't match the synapsedev ARN — so the statement grants nothing there.

Note: deploying this re-provisions the `Developer` permission set in every account it's assigned to (the inline policy is re-pushed), but the effective permissions change only in synapsedev.

Jira: https://sagebionetworks.jira.com/browse/PLFM-9782

## Test plan
- [ ] Deploy and confirm the `Developer` permission set inline policy contains the new `SynapseDevSharedResourcesCleanup` statement
- [ ] Re-attempt deleting a `DELETE_FAILED` stack in synapsedev with the affected IAM roles and confirm it succeeds

### DIFF
--- a/org-formation/700-aws-sso/_tasks.yaml
+++ b/org-formation/700-aws-sso/_tasks.yaml
@@ -1098,6 +1098,47 @@ SsoSynapseDevDeveloper:
     principalId: !Ref synapseDevDeveloperGroup
     permissionSetArn: !CopyValue [ !Sub '${resourcePrefix}-${appName}-developer-permission-set-arn' ]
 
+SsoSynapseDevDeveloperIamScoped:
+  Type: update-stacks
+  DependsOn: SsoDeveloper
+  Template: https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.3.8/templates/SSO/aws-sso.yaml
+  StackName: !Sub '${resourcePrefix}-${appName}-synapsedev-developer-iam-scoped'
+  StackDescription: 'SSO: additional scoped IAM role-management rights for synapsedev developer group'
+  TerminationProtection: false
+  DefaultOrganizationBindingRegion: !Ref primaryRegion
+  DefaultOrganizationBinding:
+    IncludeMasterAccount: true
+  OrganizationBindings:
+    TargetBinding:
+      Account: !Ref SynapseDevAccount
+  Parameters:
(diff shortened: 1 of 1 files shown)


org-formation/700-aws-sso/_tasks.yaml
```diff
@@ -455,10 +455,21 @@ SsoDeveloper:
...
+                },
+                {
+                  "Sid": "SynapseDevSharedResourcesCleanup",
+                  "Effect": "Allow",
+                  "Action": [
+                    "iam:DeleteRolePolicy",
+                    "iam:DeleteRole",
+                    "iam:DetachRolePolicy"
+                  ],
+                  "Resource": "arn:aws:iam::${SynapseDevAccountId}:role/*shared-resources*"
```
BryanFauble, line 467: As a warning - This would deploy out to all accounts that have a `Developer` role (~20 accounts). But it is scoped to only the `SynapseDevAccount` AWS account.

I chose this route so that we could continue to use the same `Developer` role in `SynapseDev` without needing to assume a new role.

If editing this policy that would be applied to ~20 accounts is not acceptable, then I would create a copy of the configuration for `SynapseDev` only. The risk is that any common changes would need to be made in 2 places instead of this shared configuration.

brucehoff (approved):
LGTM. In fact I wouldn't be against giving developers admin' access to the synapse-dev account.

PR #1631, opened 2026-08-12
IT-5133 Restore NYU access to headless-agents-benchmark-results bucket

brucehoff:
IT-5133 Restore NYU access to headless-agents-benchmark-results bucket

We had granted and then removed temporary access.  Solly requested we restore access for one more month.


org-formation/600-access/_tasks.yaml
```diff
@@ -215,6 +215,77 @@ BedrockCentralCrossAccountAccess:
...
+            "ecr:CompleteLayerUpload",
+            "ecr:BatchCheckLayerAvailability",
+            "ecr:PutImage",
+            "batch:RegisterJobDefinition",
+            "batch:SubmitJob",
+            "batch:ListJobs"
+          ],
+          "Resource": [
+            "*"
+          ]
```
zaro0508, line 251: AWS IAM supports a [DateLessThan condition](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_examples_aws-dates.html)  that allows you to restrict AWS permissions so that an action is only permitted if the request time is earlier than a specific date and time.  You can use that to automatically time box access to these resources.

zaro0508 (approved):
Generally LGTM.  To automate disabling of this access you might want to look at the `DateLessThan` condition.

